PR #2, opened 2015-08-13
Add list command

mtibben:
Adds an `ls` command to list profiles stored in the vault.

```
$ aws-vault ls
default
work
home
```

### DIFF
--- a/command/list.go
+++ b/command/list.go
@@ -1,0 +1,39 @@
+package command
+
+import (
+	"strings"
+
+	"github.com/99designs/aws-vault/Godeps/_workspace/src/github.com/mitchellh/cli"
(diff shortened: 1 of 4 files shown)


lox:
Will this list the session credentials too?

command/list.go
```diff
@@ -0,0 +1,39 @@
...
+
+	c.Ui.Output(strings.Join(profileNames, "\n"))
+
+	return 0
+}
+
+func (c *ListCommand) Help() string {
+	helpText := `
+Usage: aws-vault ls
+  Lists credential profile names in the vault
```
lox: "Lists profiles with credentials in the vault"

PR #3, opened 2015-08-13
Add an alias command with instructions for wrapping aws

mtibben:
Shows instructions for wrapping a command, appropriate for the current shell. With -s, outputs shell script suitable for eval.

```
$ aws-vault alias aws
# Wrap aws automatically by adding the following to ~/.bash_profile:
eval "$(aws-vault alias -s aws)"
```

```
$ aws-vault alias -s aws
alias aws="aws-vault exec aws"
```

Using eval means that if we change the command structure we can change the outputted command easily, but maybe it's not necessary. I followed what the github `hub` command does

### DIFF
--- a/aws-vault.go
+++ b/aws-vault.go
@@ -47,6 +47,11 @@ func main() {
 				Keyring: k,
 			}, nil
 		},
+		"alias": func() (cli.Command, error) {
+			return &command.AliasCommand{
+				Ui: ui,
+			}, nil
+		},
 	}
 
 	exitStatus, err := c.Run()
--- a/command/alias.go
+++ b/command/alias.go
@@ -1,0 +1,73 @@
+package command
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/99designs/aws-vault/Godeps/_workspace/src/github.com/mitchellh/cli"
+)
+
+type AliasCommand struct {
+	Ui cli.Ui
+}
+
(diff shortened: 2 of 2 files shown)


pda:
I'm not convinced of the value of this command; it's not ultra-elegant, and doesn't really solve anything a few lines in the README wouldn't. It specifically mentions `.bash_profile` but other shells like Zsh are also popular. And if I were to add the alias to my shell, I'd skip the wrapper command anyway to reduce shell startup overhead (even if it's negligible). But… ¯\\\_(ツ)\_/¯

mtibben:
It does already support zsh @pda.

Yeah, I don't disagree, but I thought it might be useful given it's
something that most devs would want to do, and putting it as a command
gives you some flexibility if you need to automate in a script etc.

pda:
Yeah I don't feel strongly either way.

Re. zsh, I meant that the `alias` command without `-s` basically just emits documentation, but that documentation is currently shell-specific, even though the alias itself is cross-shell. Either way, I think README / `--help` / `man aws-vault` is the better place for this stuff.

mtibben:
@pda not sure what you mean - if you look at the code it gives instructions for either bash or zsh, whatever you're running - I basically just did the same thing that `hub alias` does

pda:
Oh right, cool :) I was only looking at the example output :tired_face:

lox:
I'd almost prefer to add an `aws-vault aws` command which acts as a wrapper to aws. It's what I end up typing 90% of the time.

PR #6, opened 2015-08-13
Error for missing profile reports profile name and config file.

pda:
Before:

```
$ aws-vault exec env
Profile not found

$ aws-vault exec --profile=foo env >/dev/null
Profile not found
```

After:

```
$ aws-vault exec env
Profile 'default' not found in /Users/pda/.aws/config

$ aws-vault exec --profile=foo env >/dev/null
Profile 'foo' not found in /Users/pda/.aws/config
```

It's not immediately obvious that an entry needs to exist in `~/.aws/config` for each `aws-vault store` credential, so this error should help explain that.

This could also be helped by `aws-vault store` adding a stub entry to the AWS config if one did not already exist.

### DIFF
--- a/vault/profiles.go
+++ b/vault/profiles.go
@@ -1,7 +1,7 @@
 package vault
 
 import (
-	"errors"
+	"fmt"
 	"os"
 	"os/user"
 	"strings"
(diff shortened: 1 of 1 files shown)


pda:
Tangential handy one-liner to create config stubs for all vaulted profiles:

```sh
for p in `aws-vault ls`; do fgrep -q "[$p]" ~/.aws/config || echo "[$p]" >> ~/.aws/config; done
```

pda:
Rebased onto master, resolved conflicts since #4 merged.

PR #8, opened 2015-08-17
Refactor sts calls into vault session

lox:
My aim here was to apply [SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle) more to the exec command so that it's easier (or possible at all) to test. 

This moves the logic for getting temporary session credentials from AWS into `vault/session.go`, which should in turn be easier to test its self. The keychain session caching moves into a decorator.

Along the way, this also wraps access to the serialized credentials in a decorate on the Keychain that is accessible from the new Profile object.

### DIFF
--- a/aws-vault.go
+++ b/aws-vault.go
@@ -1,13 +1,11 @@
 package main
 
 import (
-	"flag"
 	"log"
 	"os"
 
 	"github.com/99designs/aws-vault/Godeps/_workspace/src/github.com/mitchellh/cli"
 	"github.com/99designs/aws-vault/command"
-	"github.com/99designs/aws-vault/keyring"
 )
 
 var (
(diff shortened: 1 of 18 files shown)


aws-vault.go
```diff
@@ -21,44 +19,27 @@ func main() {
 		ErrorWriter: os.Stderr,
 	}
 
-	// handle profile at the top level, I always do this.
```
lox: This might be annoying, but I think it's kind of horrible having it here as it introduces a leaky abstraction and strange behaviour like needing to support the profile flag in both spots.

lox:
Thoughts / feels @joho or @mtibben or @pda? Trying to turn this into something neatly testable has been a bit of a struggle.

PR #13, opened 2015-09-15
Use a cocoa dialog for prompting for mfa token

lox:
screenshot


lox:
IMO this should probably be an external app, or make use of https://gnupg.org/related_software/pinentry/index.html

lox:
Nope, turns out that pinentry is horrible.

lox:
Cocoa doesn't like running from background threads, so this is a no-go.

PR #16, opened 2015-09-16
Add a command to generate login links for AWS Console

lox:
This command generates a login link for assumed roles:

```
$ aws-vault login myaccount
https://signin.aws.amazon.com/federation?Action=login....
```

### DIFF
--- a/login.go
+++ b/login.go
@@ -1,0 +1,83 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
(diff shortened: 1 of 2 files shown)


joho:
Given this is hard coupled to osx should it just call `open` on the url directly?

lox:
@joho I thought about it, but I'd like to avoid hard coupling to OSX. Plus, command-click of a link in terminal opens it.

lox:
Meh, merging it.

PR #18, opened 2015-09-23
Replace env vars rather than appending

### LINKED ISSUE #17: aws-vault should replace env vars, not append
When using `aws-vault exec`, if `AWS_*` env vars have already defined, aws-vault does not replace them. Instead it appends environment variables which leads to undefined behaviour.

aws-vault should replace any environment variable that it needs to define, rather than appending

mtibben:
When using `exec`, if AWS_* env vars have already defined, aws-vault currently does not replace them. Instead it appends environment variables which leads to undefined behaviour.

This changes the behaviour to replace any environment variable that it needs to define, rather than appending

Fixes #17

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lox:
```go
	env := []string{}

	// skip keys that we are replacing
	for _, e := range os.Environ() {
		switch e {
		case "AWS_ACCESS_KEY_ID":
			continue
		case "AWS_SECRET_ACCESS_KEY":
			continue
		case "AWS_SECRET_ACCESS_KEY":
			continue
		default:
			env = append(env, e)
		}
	}

	env = append(env,
		"AWS_ACCESS_KEY_ID="+val.AccessKeyID,
		"AWS_SECRET_ACCESS_KEY="+val.SecretAccessKey,
		"AWS_DEFAULT_PROFILE="+input.Profile,
	)
```

lox:
Reckon the above might be terser. Thoughts?

mtibben:
yep sure

mtibben:
you'll also need to handle `AWS_SESSION_TOKEN`

lox:
Will you update this @mtibben?

lox:
Closing, this is fixed in #26

PR #19, opened 2015-09-23
Document MFA configuration

dgoodlad:
Adds some documentation about the `mfa_serial` property.

This is important, because if your IAM account has MFA enabled, you need to include your MFA token in its profile configuration in order to make any requests with STS-generated temporary credentials. Otherwise the `AWS_SESSION_TOKEN` will be treated as invalid.


lox:
Many thanks!

PR #20, opened 2015-09-23
Store sessions encrypted on disk locally

lox:
Previously sessions (from GetSessionToken and AssumeRole) were stored in the keyring. This had the annoying effect of prompting you about whether to allow access to the session, rather than simply the original profile. 

The secondary issue was that sessions were tied to profile. This change stores them locally (in `~/.awsvault/sessions`), encrypted with the AWS secret access key, with the filename a hash of the profile name, the access key id and the mfa_serial of the profile. This means if you change your profile config or the credentials you will get a new session.

Currently there is no garbage collection on old sessions, but the expiry time is in the token metadata, so doesn't need a private key to read.

This makes use of https://github.com/dvsekhvalnov/jose2go with PBES2_HS256_A128KW for key derivation and the A256GCM encryption algorithm.

### DIFF
--- a/add.go
+++ b/add.go
@@ -8,9 +8,10 @@ import (
 )
 
 type AddCommandInput struct {
-	Profile string
-	Keyring keyring.Keyring
-	FromEnv bool
+	Profile      string
+	Keyring      keyring.Keyring
+	SessionStore *sessionStore
+	FromEnv      bool
 }
 
 func AddCommand(ui Ui, input AddCommandInput) {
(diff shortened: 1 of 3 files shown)


lox:
I'll add the new vendors once this gets some review. @pda feels?

lox:
I guess the only downside to this is that the original credentials are in memory, but if an attacker is in a position where they can dump process memory, it's probably way too late.

mtibben:
I'm not sure about this approach. The keychain is designed to store credentials, it seems to me that this is just reproducing what keychain functionality

Can't we configure the keychain to dial down the confirmation noise?

lox:
The keychain isn't designed to store ephemeral credentials, IMO.

pda:
I quite like being prompted every time something accesses AWS as me. Hitting spacebar confirms the prompt, so it's not very disruptive (unlike if you have to switch from keyboard to mouse and back).

I haven't done enough real-world aws-vault use yet to know how much of an improvement this would be, but I'd be hesitant to add complexity, third-party libraries, encryption schemes and perhaps a slightly weakened security model unless it's really important.

lox:
Yeah, after some thought I think you are both correct.

PR #26, opened 2015-09-30
Use a local metadata server for credentials, fixes #24

### LINKED ISSUE #24: Long running sub-processes result in expired keys
The maximum expiry on assume-session credentials is unfortunately 60 minutes (http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html). What this means is that a long running process (for instance a complicated `packer` build) will end up with expired credentials.

Because `aws-vault` gets credentials to the child process as environmental variables, there aren't any mechanisms to refresh these. The various SDK's and `aws-cli` refresh these internally.  

I'm currently investigating different options, none so far which are a perfect fit.

Interested in thoughts. @joho @dgoodlad @pda @mtibben

lox:
All AWS SDK's check for a [metadata service](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) at http://169.254.169.254, which is used for instance roles. 

This PR runs a server for every exec call and passes it to the subprocess as an `HTTP_PROXY` var and then intercepts the call to the metadata service and returns credentials. This seems to work great and the SDK's know to fetch new credentials when the current ones expire. If one provides a longer underlying session expiry then your role credentials will be renewed without prompting for MFA.

There is a `--write-env` fallback that will also write the env variables like the earlier versions.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 	"os/user"
@@ -11,14 +14,22 @@ import (
 
 type profiles map[string]map[string]string
 
-func parseProfiles() (profiles, error) {
-	file := os.Getenv("AWS_CONFIG_FILE")
+func configFile() (file string, err error) {
+	file = os.Getenv("AWS_CONFIG_FILE")
 	if file == "" {
 		usr, err := user.Current()
 		if err != nil {
-			return nil, err
+			return file, err
 		}
 		file = usr.HomeDir + "/.aws/config"
+	}
+	return file, err
+}
+
+func parseProfiles() (profiles, error) {
+	file, err := configFile()
+	if err != nil {
+		return nil, err
 	}
 
 	log.Printf("Parsing config file %s", file)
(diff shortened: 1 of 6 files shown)


lox:
Ok, this is in a pretty good place now. Some testing would be helpful testing it is slightly painful as the minimum timeout is 15 minutes.

lox:
I've released this on the experimental release: https://github.com/99designs/aws-vault/releases/tag/experimental

exec.go
```diff
@@ -35,24 +38,85 @@ func ExecCommand(ui Ui, input ExecCommandInput) {
...
+
+// write out a config excluding role switching keys
+func profileConfig(profile string) (*os.File, error) {
+	conf, err := parseProfiles()
 	if err != nil {
-		ui.Error.Fatal(err)
+		return nil, err
+	}
+
+	tmpConfig, err := ioutil.TempFile(os.TempDir(), "aws-vault")
```
joho: should you be `defer`ing a close for this config file?

config.go
```diff
@@ -11,15 +12,23 @@ import (
...
-			return nil, err
+			return file, err
 		}
 		file = usr.HomeDir + "/.aws/config"
 	}
+	return file, err
+}
+
+func parseProfiles() (profiles, error) {
+	file, err := configFile()
```
joho: likewise, this getting cleaned up?

exec.go
lox: Good catch.

config.go
lox: this one isn't opening a file, just getting the path to a file.

config.go
```diff
@@ -11,15 +12,23 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
 
 	log.Printf("Parsing config file %s", file)
 	f, err := ini.LoadFile(file)
```
joho: Sorry, and this doesn't need to close cos the ini package takes care of it. I guess I'm out of nits.

Issue #23, opened 2015-09-29
Linux support


lox:
Support for Linux users, whether command-line or desktop environment users. 

Todo:
- [x] Implement KWallet4 keyring backend
- [ ] Implement KWallet5 keyring backend
- [ ] Implement gpg-agent backend
- [x] Implement encrypted file backend
- [x] Release process needs to build linux binaries

ketzacoatl:
With Go previously installed on Linux (Ubuntu Trusty, AMD64), I was able to build aws-vault with the following:

```
mkdir -p /home/user/.golang/src/github.com/99designs/
cd /home/user/.golang/src/github.com/99designs/
git clone https://github.com/99designs/aws-vault/
cd aws-vault
export GOPATH=/home/user/.golang
export GOROOT=/usr/local/go
go get
make
```

`aws-vault --help` works, but I haven't yet gotten further (need to add kwallet).

```
aws-vault
usage: aws-vault [<flags>] <command> [<args> ...]

A vault for securely storing and accessing AWS credentials in development environments.

Flags:
  --help             Show context-sensitive help (also try --help-long and --help-man).
  --debug            Show debugging output
  --prompt=terminal  Prompt driver to use [terminal]
  --version          Show application version.

Commands:
  help [<command>...]
    Show help.

  add [<flags>] <profile>
    Adds credentials, prompts if none provided

  ls
    List profiles

  exec [<flags>] <profile> [<cmd>] [<args>...]
    Executes a command with AWS credentials in the environment

  rm [<flags>] <profile>
    Removes credentials, including sessions

  login [<flags>] <profile>
    Generate a login link for the AWS Console

  server
    Run an ec2 instance role server locally
```

lox:
Thanks for trying it out @ketzacoatl, what secret store are you hoping to use it with? Definitely need to get better feedback into the tool on what secret stores are available and what it's using!

ketzacoatl:
@lox thanks for checking in, I would be happy to provide input. ATM I am evaluating `aws-vault` as a tool for use in a team environment, and I am interested in the simple/functional route, and a cross-platform solution to the team's need for AWS credential management across multiple AWS accounts, regions, and environments, and with MFA in the workflow.

[...]

PR #27, opened 2015-10-01
Add support for the kwallet keyring

Vektah:
Adds support for the KDE wallet. CLI/Gnome would require their own backends.

When first creating the wallet it will prompt for a new password:
new-wallet

If the wallet is not open it will prompt for the password:
open-wallet

Unlike the OSX keyring the KDE wallet 'entries' are purely key value stores. To fit the entire Item I added a second layer of json encoding.

I'm not sure of the best way to automatically test this. I guess I could stub the dbus connection or the wrapper?

see https://github.com/99designs/aws-vault/issues/23

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -11,6 +11,10 @@
 			"Rev": "6b4e7dc5e3143b85ea77909c72caf89416fc2915"
 		},
 		{
+			"ImportPath": "github.com/aulanov/go.dbus",
+			"Rev": "25c3068a42a0b50b877953fb249dbcffc6bd1bca"
+		},
+		{
 			"ImportPath": "github.com/aws/aws-sdk-go/aws",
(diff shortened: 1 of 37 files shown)


lox:
Nice one @Vektah, Godeps need updating.

lox:
Mind rebasing master and updating Godeps?

lox:
So just to clarify, Kwallet doesn't have libsecret support? That would be the vastly better option IMO:
https://developer.gnome.org/libsecret/unstable/

Vektah:
I've rebased and added godeps. The gitignore for vendor threw me off.

https://github.com/tmc/keyring/blob/master/keyring_linux.go uses the libsecret dbus api via `org.freedesktop.Secret.Service` but it didn't prompt for a wallet password at all and ended up being stored somewhere else.

PR #31, opened 2015-10-05
Show MFA token while typing

Vektah:
This makes the MFA token visible while entering. It is not vulnerable to being pulled out of the scrollback buffer as it only lasts for a short duration. 

Both AWS and Google show the MFA token as its being typed.


lox:
Yeah, think this is totally fine.

pda:
:+1:

PR #34, opened 2015-10-10
Add a --master arg to exec to use master credentials

lox:
I've run into a few tools that use older boto libraries that don't support STS tokens
or IAM instance credentials. 

I'm torn on whether this is a good idea, and if it is a good idea what it should be called.
It came down to either `--master` or `--no-session`. 

Thoughts? @joho @pda @dgoodlad

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -18,18 +18,23 @@ import (
 )
 
 type ExecCommandInput struct {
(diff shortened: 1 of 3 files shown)


lox:
Closing in favour of #35

PR #37, opened 2015-10-11
Add an `--mfa-token` param to `exec`, closes #25

### LINKED ISSUE #25: Provide MFA token as a command-line argument
There are some workflows which use command-line tools such as [oath-toolkit](http://www.nongnu.org/oath-toolkit/oathtool.1.html) for generating MFA tokens.

It would be handy to be able to pass the MFA token value into the `aws-vault` call as a flag, which would make it easy to integrate.  Something like this:

```
$ aws-vault exec --mfa-token=123456 admin -- aws ...
```

I'm not familiar with Golang but will try to create a PR for this!

lox:


lox:
https://github.com/99designs/aws-vault/pull/37

simpsora:
@lox works great, thank you!  I'd also find this useful in the `login` command -- would you accept a commit for that in this PR?

lox:
Yup, sure!

simpsora:
@lox: ah, I don't have permission to push to this branch.

I did have a build issue:

```$ GO15VENDOREXPERIMENT=1 go build
# github.com/99designs/aws-vault
./provider.go:194: cannot use p.VaultOptions.MfaToken (type string) as type *string in assignment```

Casting to `aws.String()` seems to fix it:
```
--- a/provider.go
+++ b/provider.go
@@ -191,7 +191,7 @@ func (p *VaultProvider) getSessionToken(creds *credentials.Value) (sts.Credentia
                        }
                        params.TokenCode = aws.String(token)
                } else {
-                       params.TokenCode = p.MfaToken
+                       params.TokenCode = aws.String(p.MfaToken)
                }
        }
```

For the change to `login`, since I can't push here, shall I just open a new PR?

simpsora:
@lox: please disregard my earlier comment (deleted) and see this PR instead:
https://github.com/99designs/aws-vault/pull/38

coen-hyde:
thanks for this @lox and @simpsora!

PR #42, opened 2015-10-14
aws-vault logout <profile>: delete session credential.

pda:
`aws-vault logout <profile>` removes the session credential for that profile.

I was trying to solve an aws-vault problem and suspected a bad session credential, but that turned out to not be the case. Figured this command might be useful in future anyway.

### DIFF
--- a/logout.go
+++ b/logout.go
@@ -1,0 +1,18 @@
+package main
+
+import (
+	"github.com/99designs/aws-vault/keyring"
+)
+
+type LogoutCommandInput struct {
+	Profile string
+	Keyring keyring.Keyring
+}
+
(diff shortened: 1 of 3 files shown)


lox:
Would this be better as `aws-vault rm --session` ?

pda:
I considered that (actually mostly implemented it, then rolled it back). I'm not sure… maybe?

I think `rm` maps more closely to the internal implementation, but `logout` maps more closely to what the user is trying to do. I was also considering `clear-session` etc.

I was also considering something like `cleanup` or `clear-sessions --all`.

lox:
I'd be good with `clear-sessions`.

lox:
Superceded by #52.

Issue #41, opened 2015-10-13
Some AWS SDK's don't support http_proxy for Instance Metadata


lox:
Some SDKs don't support `http_proxy` for Instance Metadata, which means that they can't get credentials from aws-vault, as it exposes the server instance via `http_proxy`. 

The impact of this issue is that some tools just aren't able to get credentials, like in #40.

(Side note: some SDKs have very short timeouts, probably a different issue)

So whilst there is the `--write-env` that provides a band-aid, it's hard to imagine how to provide a solution that would work universally. Some options I'm considering are:

 1. Change the default back to writing out environment vars and allow opt-in to the metadata approach
 2. Provide a small daemon that would actually bind to `169.254.169.254` and forward to the correct exec server via magic. :'(
 3. ???

mtibben:
Could you just do both? Write out env vars as well as use the metadata
approach?

pda:
I'm guessing the presence of the environment variables would stop the SDK from using the instance metadata server; presumably on EC2 you can override the latter with the former…

You can't bind to `169.254.169.254` without having a network interface associated with that address/network. Dynamically reconfiguring network interfaces is another jump in complexity. It might work though, and would seem more reliable than `http_proxy`.

lox:
After some thought, I think the best way forward is to do a couple of things:

 * Default to writing environment vars. 
 * Allow opting in to `--ec2-metadata`, which defaults to 127.0.0.1:0 via `http_proxy` env.
 * Allow `--ec2-metadata=169.254.169.254`, which will work if you've run the `contrib/lo-ec2-metadata.sh script that will be written. 
 * Add an `aws` command for invoking the aws cli with the correct params.

mtibben:
Perhaps you could introduce the idea of profiles - a set of environment var config to make a certain command or sdk work. The `default` profile should attempt to work in most use cases, but you could potentially have `boto` or `old-aws` or `ruby-sdk` profiles to give those environments what they're expecting

lox:
It's tempting @mtibben, but I'd rather something that simply works for everything.

lox:
Fixed in master.

PR #43, opened 2015-10-15
Bind instance role server directly to 169.254.169.254:80

lox:
This avoids the compatibility issues associated with #41 by binding directly to the 
address that the aws sdk's look for with `aws-vault server` (as root) and proxying to 
a less privileged http server run by exec on `127.0.0.1:9099`. 

The exec command tries to do this, including adding the alias on `lo` for `169.254.169.254` and falls back to using the env vars if not possible.

This approach is possible, but will require some change for Windows.

The major caveat is that this doesn't support parallel `aws-vault exec` invocations. The second one will fail to bind and will fall back to env.


lox:
My take is that this is actually less voodoo than the current approach. The only bit that makes me uncomfortable is the exec command launching `aws-vault server` as a background process if it's not running, but I think it's still reasonable.

mtibben:
I would vote for env vars being the default. Needing `sudo` and changing network configuration seems heavy handed and unexpected for a credential manager

lox:
@mtibben yeah, I think you are right.

server.go
```diff
@@ -0,0 +1,118 @@
...
+	log.Printf("Fetched credentials from %s", localServerUrl)
+
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(http.StatusOK)
+
+	io.Copy(w, resp.Body)
+}
+
+func installNetworkAlias() ([]byte, error) {
+	return exec.Command("ifconfig", "lo0", "alias", "169.254.169.254").CombinedOutput()
```
mtibben: does this need to be removed at some point e.g. when the server is stopped ?

lox: Not really, it doesn't do any harm and it's addition is idempotent.

Issue #44, opened 2015-10-15
Instance Metadata server complicates process model


pda:
Introducing the Instance Metadata server in https://github.com/99designs/aws-vault/pull/26 meant replacing `syscall.Exec` with `exec.Command().Run()`; the former replaces the current aws-vault process with the command being executed, the latter runs it as a child process. This introduces lots of complexity and downside;

- The pid of the command being executed is different to the PID that the shell saw launched.
- Signal handling needs to be managed.
- Exit status needs to be captured and passed through.
- The proctitle remains `aws-vault` rather than that of the command being executed.
    - Go doesn't support reading/writing proctitle without using barely maintained third-party libs.

This isn't a call-to-action issue, just highlighting some of the downsides alongside discussion in https://github.com/99designs/aws-vault/pull/43 etc.

lox:
Yeah, I'd prefer to go back to Exec if possible. One catch would be that Windows doesn't support Exec, but that could be special cased I guess when that problem becomes a thing. 

The only complexity I forsee is that with Exec there is no way to cleanup the generated AWS_CONFIG_FILE. Perhaps there is no point in generating it and we should point it at `/dev/null` or equiv.

lox:
Btw, the other major annoyance with Exec is that it swallows `go test` runs :)

lox:
It seems like `proctitle` would be reasonably easy to address in that list. Do you have any example of where the different pid causes issues?

pda:
Nope, it's mostly proctitle that bothers me, but also the awareness of additional complexity. But that awareness goes away when it all Just Works for a while ;)

PR #45, opened 2015-10-15
Use ForkExec rather than Subprocesses

lox:
Things are simpler with ForkExec/Exec vs Subprocesses. Refactoring the env handling also saved a fair amount of code. 

This also gets rid of the generation of a config. Everything that is needed from the config goes into the environment vars.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -22,6 +22,10 @@ type ExecCommandInput struct {
 }
 
 func ExecCommand(ui Ui, input ExecCommandInput) {
+	if os.Getenv("AWS_VAULT") != "" {
+		ui.Fatal("aws-vault sessions should be nested with care, unset $AWS_VAULT to force")
+	}
+
 	creds, err := NewVaultCredentials(input.Keyring, input.Profile, VaultOptions{
 		SessionDuration: input.Duration,
 		MfaToken:        input.MfaToken,
--- a/exec.go
+++ b/exec.go
@@ -19,7 +19,6 @@ type ExecCommandInput struct {
 	Duration    time.Duration
 	MfaToken    string
 	StartServer bool
-	Signals     chan os.Signal
 }
 
 func ExecCommand(ui Ui, input ExecCommandInput) {
--- a/exec.go
+++ b/exec.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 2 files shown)


lox:
@pda :)

exec.go
```diff
@@ -46,23 +44,16 @@ func ExecCommand(ui Ui, input ExecCommandInput) {
...
-	env = overwriteEnv(env, "AWS_CONFIG_FILE", cfg.Name())
-	env = overwriteEnv(env, "AWS_DEFAULT_PROFILE", input.Profile)
-	env = overwriteEnv(env, "AWS_PROFILE", input.Profile)
-
-	env = unsetEnv(env, "AWS_ACCESS_KEY_ID")
-	env = unsetEnv(env, "AWS_SECRET_ACCESS_KEY")
-	env = unsetEnv(env, "AWS_CREDENTIAL_FILE")
+	env := environ(os.Environ())
+	env.Set("AWS_CONFIG_FILE", "/dev/null")
+	env.Set("AWS_VAULT", input.Profile)
```
lox: This paired with the removal of the generated config file will make for a 3.x.x release.

exec.go
```diff
@@ -78,85 +73,49 @@ func ExecCommand(ui Ui, input ExecCommandInput) {
...
 	if err != nil {
-		return nil, err
+		ui.Error.Fatal(err)
 	}
 
-	newConfig := map[string]string{}
+	pid, err := syscall.ForkExec(path, append([]string{input.Command}, input.Args...), &syscall.ProcAttr{
+		Env:   env,
+		Files: []uintptr{0, 1, 2},
+	})
```
pda: I think `syscall.ForkExec()` has most of the same issues that https://github.com/99designs/aws-vault/issues/44 describes (i.e. it's still a subprocess), and that `syscall.Exec()` would be required to go back to single-process simplicity. That wont be viable in server mode, but perhaps in env mode?

lox: My understanding was that ForkExec is like Exec, but forks the current process to the background.

lox: It seems to work that way, could you try it? https://github.com/99designs/aws-vault/releases/tag/v3.0.0-dev

pda:
```
$ git checkout v3.0.0-dev
HEAD is now at 546d8ef... Add warning about nested exec sessions
$ go build && ./aws-vault exec pda -- pstree $$
-+= 24359 pda -zsh
 \-+= 26285 pda ./aws-vault exec pda -- pstree 24359
   \-+- 26287 pda pstree 24359
     \--- 26288 root ps -axwwo user,pid,ppid,pgid,command
```

Compared to:

```
$ ./aws-vault-2.2.1 exec pda -- pstree $$
-+= 24359 pda -zsh
 \-+= 26391 pda pstree 24359
   \--- 26392 root ps -axwwo user,pid,ppid,pgid,command
```

The process title can be visually tested with:

```sh
./aws-vault-2.2.1 exec pda -- sleep 2  # process title changes to `sleep`
./aws-vault exec pda -- sleep 2        # process title remains `aws-vault`
```

(as reported by tmux window name or iTerm/Terminal tab name)

> My understanding was that ForkExec is like Exec, but forks the current process to the background.

I believe `ForkExec` is how e.g. a shell works; parse the command, fork the current shell process, exec the program in the child process with env updated and new argv. So the executed command remains a child process.

What you're aiming for would be more like this:

* Fork the go process.
* Launch the metadata server in the child process.
* Disown the child server process; reparent to PID 1?
* Exec the user-specified command to replace the original process with a new one.

Thus the originally launched PID becomes the user-specified program after aws-vault has modified the environment and started another process. However that's not possible because Go can't fork itself; it can only fork to exec other commands.

So; in normal mode we could configure the environment then `syscall.Exec()` to replace the process with the user command, and in server mode we could use `syscall.ForkExec` or `exec.Command().Run()` to run the user command as a child process, and miss out on things like process title, simplified signal handling, exit status propagation etc.

lox:
I'd love to find a solution to this, but this isn't it. Pulling out the env and profile stuff elsewhere.

PR #53, opened 2015-11-11
Use open-golang to open login url in default browser

lox:
I got sick of copying and pasting or command-clicking the url for `aws-vault login`, so this uses http://github.com/skratchdot/open-golang/open for cross-platform "open in default browser".


pda:
:+1: works on my machine.

PR #57, opened 2015-12-02
Support different dialog types

lox:
Because `aws-vault` becomes the background process, re-prompting for mfa in long running modes isn't possible. This adds a dialog prompt via `osascript` for osx and a pluggable mechanism for prompting on other systems. The default is still terminal, but you can override this with either a command-line arg or and env:

```
aws-vault --prompt osascript exec sandbox
```

or

```
AWS_VAULT_PROMPT=osascript aws-vault exec sandbox 
```


lox:
Thoughts on this @pda?

pda:
Seems better than broken background prompts. But perhaps it should be `--osascript` not `-cocoa`?

PR #60, opened 2016-03-15
Command to rotate credentials

sj26:
It can be useful to rotate credentials quickly and securely. After one of these rotations the only place my secret access key has ever been stored is within aws-vault's keychain.

This is pretty rough, but it works. Seeking thoughts/comments.


lox:
Awesome! Thanks for the contribution, I've been wanting a rotate command for a while. 

Should the default be to leave the old key inactive or actually deleted? (https://blogs.aws.amazon.com/security/post/Tx15CIT22V4J8RP/How-to-rotate-access-keys-for-IAM-users)

sj26:
Yeah I'm not sure. You can only have two access keys at a time, even if they're inactive, so it will only work if you've deleted your previous access key. When testing I was rotating over and over without using the console because it was deleted.

But you're probably right, the officially recommended workflow is creating a new one, replacing existing usage (which I guess could be on other machines, too?), mark the old one inactive, then eventually delete the old one. We can't encapsulate that whole process in a single command. Should we do the create & mark inactive part then print some instructions? Or provide a multi-stage `rotate --create`, `rotate --deactivate` and `rotate --delete` or something and keep a record of the rotating in progress, printing instructions at each step?

lox:
I think how you've done it is the sensible default. A future addition might be to add a `--invalidate` or something similar.

sj26:
:tada:

PR #63, opened 2016-05-03
Add a basic file backend

lox:
This adds a file backend for storing keyring items encrypted in ~/.awsvault/keys/{keyname}. Keys are encrypted as JWT tokens.

The usecase for this is for command-line linux usage. It also opens up the possibility of using it on servers to store credentials.

### DIFF
--- a/keyring/file.go
+++ b/keyring/file.go
@@ -1,0 +1,155 @@
+package keyring
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/user"
+	"path/filepath"
+	"time"
+
+	jose "github.com/dvsekhvalnov/jose2go"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+type passwordFunc func(string) (string, error)
+
+func terminalPrompt(prompt string) (string, error) {
+	fmt.Printf("%s: ", prompt)
+	b, err := terminal.ReadPassword(1)
+	if err != nil {
+		return "", err
+	}
+	fmt.Println()
+	return string(b), nil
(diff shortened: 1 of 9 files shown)


neumayer:
I started working on my own simple file backend before I saw this :-). I think your commit should be rather easy to extend further to use gpg, in the future I mean.

Anyway, I tested locally and this works for me. It might be nice to add some tests for non-happy paths.

Any plans on whether you'll merge this?

lox:
Thoughts @pda @mtibben?

mtibben:
👍

keyring/file.go
```diff
@@ -0,0 +1,155 @@
...
+	fmt.Println()
+	return string(b), nil
+}
+
+func init() {
+	supportedBackends[FileBackend] = opener(func(name string) (Keyring, error) {
+		return &fileKeyring{
+			PasswordFunc: terminalPrompt,
+		}, nil
+	})
```
mtibben: It may be better to keep use of the `supportedBackends` in `keyring.go` as the scope of how this works isn't immediately obvious

pda:
> Keys are encrypted as JWT tokens.

Why JWT? Does the encryption layer not provide authentication?

pda:
Oh right, it's encrypting the data using the Javascript Object Signing and Encryption (JOSE) and/or JWE (JSON Web Encryption) things that are related to JWT.

pda:
> The usecase for this is for command-line linux usage.

Seems good 👍 

> It also opens up the possibility of using it on servers to store credentials.

Sounds dangerously out of scope.

Issue #59, opened 2016-02-16
Assume role session expires in 15 mins


ptolemy428:
Hi @lox,  

Currently, when running aws-vault with exec, it creates a session (for assuming role), which times out after 15 mins. AWS assume role API allows up to 1 hour as the duration of the session.

Is there a way to configure this setting with aws-vault (i.e. env vars, config files, etc)?

~/development/ca/data$ aws-vault --debug exec dev-admin
DEBUG 2016/02/16 14:26:38 Parsing config file /Users/dev/.aws/config
DEBUG 2016/02/16 14:26:38 Looking up keyring for default
DEBUG 2016/02/16 14:26:38 Opening keychain /Users/dev/Library/Keychains/aws-vault.keychain
DEBUG 2016/02/16 14:26:40 Opening keychain /Users/dev/Library/Keychains/aws-vault.keychain
DEBUG 2016/02/16 14:26:43 Using session ****************LVXQ, expires in 3h41m40.575404173s
DEBUG 2016/02/16 14:26:43 Assuming role arn:aws:iam::xxxxxxxxxxxxxx:role/xxxxxxxxxxxx
DEBUG 2016/02/16 14:26:45 Using role ****************XSZA, **expires in 14m59.24446076s**
DEBUG 2016/02/16 14:26:45 Parsing config file /Users/dev/.aws/config
DEBUG 2016/02/16 14:26:45 Writing temporary credentials to ENV
cmd has been run. exit

Some background information, we use aws-vault to create sessions on local machine to run aws command. some of the commands take more than 15 mins to complete. (i.e. such as baking AMIs). The session timed out after 15 mins, which leavings everything hanging.

lox:
Interesting, I wonder why I did that. The best way to solve this problem is to use the `--server` parameter to `aws-vault exec` which will run an ec2 metadata service so that your credentials can be refreshed. I'll look into why the role is using the minimum expiry time.

lox:
I believe I just never got around to adding a flag for setting the `assume-role-ttl`, there is on for the `session-ttl`. I'd happily accept a PR, or I can look into this when I've got a moment.

mtibben:
Fixed in #64

warrenseen:
@lox the default lifetime for STS role tokens is the maximum, 1 hour (ref: http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html). 

[...]

PR #64, opened 2016-05-26
Add an Assumed Role duration flag.

warrenseen:
Adds a flag (--assume-role-ttl) and support for passing through a custom duration for the assumed role to the VaultProvider.

(Edited to add: this is the first time I've done anything with go - I think I've done this right, and it works for me (TM), but happy to take suggestions on board.)

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -15,16 +15,17 @@ import (
 )
 
 type ExecCommandInput struct {
-	Profile     string
-	Command     string
-	Args        []string
-	Keyring     keyring.Keyring
-	Duration    time.Duration
-	MfaToken    string
-	MfaPrompt   prompt.PromptFunc
-	StartServer bool
-	Signals     chan os.Signal
-	NoSession   bool
+	Profile      string
+	Command      string
+	Args         []string
+	Keyring      keyring.Keyring
+	Duration     time.Duration
+	RoleDuration time.Duration
+	MfaToken     string
+	MfaPrompt    prompt.PromptFunc
+	StartServer  bool
+	Signals      chan os.Signal
+	NoSession    bool
 }
 
 func ExecCommand(ui Ui, input ExecCommandInput) {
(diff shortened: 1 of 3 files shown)


simpsora:
Would love to have this.  :+1:

joho:
The code looks fine to me. Just out of curiosity, what's the use case for the flag?

warrenseen:
@joho Was initially mentioned in this issue: https://github.com/99designs/aws-vault/issues/59, but we've been seeing long running tasks failing with the message: 

`AWS::RDS::Errors::ExpiredToken: The security token included in the request is expired` as the default token lifetime is 15m.

joho:
Sounds good to me.

mtibben:
lgtm

lox:
Out of interest @warrenseen, why not use the `--server`? We should probably document its use better.

warrenseen:
@lox that was the first thing we tried, but even using the server, we were running into the same issue. I think whatever operation was causing the timeout was not refreshing the role token when it was expired (building an RDS instance with CloudFormation seemed a common trigger).

lox:
If that is the case, I'd be inclined to bump the default to 30 mins.

lox:
I'm generally inclined to keep knobs and dials to a minimum with this sort of thing, I'm going to reopen #59. Thoughts @mtibben?

mtibben:
The default in aws-vault has always been the minimum 15 minutes. The [AWS API defaults](http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) to the maximum 1 hour. 

I guess if you changed the aws-vault default to 1 hour, then this flag might not be as useful.

But still it doesn't hurt to expose this option for those who want it, given that this tool is all about issuing temporary creds. This was also [previously discussed](https://github.com/99designs/aws-vault/issues/59#issuecomment-184509946)

PR #66, opened 2016-06-06
fix build due to non-backward compatible AWS SDK changes

achille-roussel:
Hello,

I was facing an issue while trying to build the latest version of aws-vault which I think were caused by non-backward compatible changes that have been introduced to the AWS SDK:
```
$ go build .
# github.com/99designs/aws-vault
./provider.go:197: cannot use aws.Config literal (type *aws.Config) as type client.ConfigProvider in argument to sts.New:
	*aws.Config does not implement client.ConfigProvider (missing ClientConfig method)
./provider.go:217: cannot use aws.Config literal (type *aws.Config) as type client.ConfigProvider in argument to sts.New:
	*aws.Config does not implement client.ConfigProvider (missing ClientConfig method)
./rotate.go:51: cannot use aws.Config literal (type *aws.Config) as type client.ConfigProvider in argument to iam.New:
	*aws.Config does not implement client.ConfigProvider (missing ClientConfig method)
./rotate.go:100: cannot use aws.Config literal (type *aws.Config) as type client.ConfigProvider in argument to iam.New:
	*aws.Config does not implement client.ConfigProvider (missing ClientConfig method)
```

The changes in this PR fix the issue and allow compiling with the latest release of the AWS SDK.

Please let me know if you want to see anything changed.

PS: The README mentions that aws-vault should be built with `GO15VENDOREXPERIMENT=1`, I was using go 1.6 which has this turned on by default, but there are no `vendor` directory in the repo so dependencies aren't matched on a specific release which is why the issue happened (I didn't know what version of the AWS SDK aws-vault should be built with tho that's why I had to fix the sources).

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -11,6 +11,7 @@ import (
 	"github.com/99designs/aws-vault/prompt"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/sts"
 )
 
@@ -191,11 +192,11 @@ func (p *VaultProvider) getSessionToken(creds *credentials.Value) (sts.Credentia
 
 	client := p.client
 	if client == nil {
-		client = sts.New(&aws.Config{
+		client = sts.New(session.New(&aws.Config{
 			Credentials: credentials.NewCredentials(&credentials.StaticProvider{
 				Value: *creds,
 			}),
-		})
+		}))
 	}
 
 	log.Printf("Getting new session token for profile %s", p.profiles.sourceProfile(p.profile))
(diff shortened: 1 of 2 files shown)


lox:
Thanks, this repo is using https://github.com/Masterminds/glide for managing the `vendor/` dir, which is not checked in. You can checkout the correct dependencies with `glide install` in the working dir. This needs to be better documented, and probably the fact that the `vendor/` dir isn't checked in is a bug.

achille-roussel:
OK, thanks for the info. I just saw that the `vendor` directory is listed in the `.gitignore` is that expected?

Also, is there any reason not to use the default vendoring mechanism available in Go? (like just adding git submodules to the `vendor` directory and letting `go get` do the job)

lox:
There are reasons, but none of them probably compelling any more. I'll put in a PR that includes them for master and then we can evaluate bumping the aws sdk version in this PR.

lox:
Mind updating this now that vendors is committed?

lox:
I'm going to close this due to inactivity.

PR #67, opened 2016-06-07
Commit vendor directory

lox:
This simplifies things.


lox:
See https://github.com/99designs/aws-vault/pull/66#issuecomment-224032358

lox:
Thoughts @pda @mtibben

pda:
Yep I think committing vendored deps is the least worst approach.
I'm also curious if Glide is worthwhile in a post-go-1.6 world.

PR #74, opened 2016-07-25
Use a more modular kingpin cli configuration

lox:
This PR cleans up the quite complicated cli setup. It's longer vertically, but less horizontally dense and easier to figure out what's going on. It also sets the stage for more verbose help.


pda:
This seems to have broken things. The `PreAction` that sets `keyringImpl` happens *after* the `configureLoginCommand` etc commands are called, each of which passes `Keyring: keyringImpl` (nil at that stage) to the specific input e.g. `LoginCommandInput`.

Removing the `PreAction` wrapper of course means the `globals.Backend` doesn't exist yet, and without passing a pointer to a pointer to keyringImpl to all the input structs I couldn't see a quick fix, but maybe there is one?

lox:
Ugh. Apologies @pda. Looking now.

lox:
Really need some tests that exercise this stuff.

PR #75, opened 2016-08-03
Stdout parameter to not open login URL in default browser

porty:
New flag for login: `aws-vault login --stdout xxx`: prints the login URL to stdout instead of opening in default browser.

### DIFF
--- a/login.go
+++ b/login.go
@@ -20,6 +20,7 @@ type LoginCommandInput struct {
 	Keyring   keyring.Keyring
 	MfaToken  string
 	MfaPrompt prompt.PromptFunc
+	UseStdout bool
 }
 
 func LoginCommand(ui Ui, input LoginCommandInput) {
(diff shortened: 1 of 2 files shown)


joho:
LGTM

pda:
Looks sensible to me.

For background, this solves cases like wanting to open in Firefox despite default being Chrome, or wanting to login on a system without `open`/`start`/`xdg-open`.

lox:
I'd probably have gone with --print rather than --stdout=xxx

pda:
The `xxx` part was a placeholder for the profile name in the command, unrelated to the `--stdout` boolean flag. But yeah `--print` would work just as well.

FernandoMiguel:
in case someone was looking for the same thing i was 
`$ open -a "Google Chrome Canary" --args --no-first-run -disk-cache-dir=$(mktemp -d /tmp/chromecanary.XXXXXX) --user-data-dir=$(mktemp -d /tmp/chromecanary.XXXXXX) "$(aws-vault --debug login <PROFILE> --stdout)"`

FernandoMiguel:
added an alias option
https://github.com/FernandoMiguel/kb/blob/master/Chrome/aws-vault.md

PR #80, opened 2016-08-09
Basic tests and Travis-CI integration

lox:
This project is woefully short on tests and CI. This PR will add some basic tests and CI via Travis. Feels @joho @mtibben @pda?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ FLAGS=-X main.Version=$(VERSION) -s -w
 CERT="Developer ID Application: 99designs Inc (NRM9HVJ62Z)"
 SRC=$(shell find . -name '*.go')
 
+test:
+	go test $(shell go list ./... | grep -v /vendor/)
+
 build:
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -8,6 +8,9 @@ FLAGS=-X main.Version=$(VERSION) -s -w
 CERT="Developer ID Application: 99designs Inc (NRM9HVJ62Z)"
 SRC=$(shell find . -name '*.go')
 
+test:
+	go test $(shell go list ./... | grep -v /vendor/)
```
joho: There's a glide command you can run here instead of using shell

joho: `go test $(shell glide nv)`

lox: Indeed, but then it would depends on glide :)

lox: Which would mean we'd need to install glide on Travis. Because it has vendors committed glide is just needed for updating.

joho: ah, of course, i saw the use of glide and just assumed from there. travis != our buildkite setup

Issue #71, opened 2016-07-24
Add binary to homebrew


lox:
Looks like @createdbypete added a PR to homebrew, but it was rejected. I've commented.

https://github.com/Homebrew/homebrew-binary/pull/317

dgoodlad:
Fixed in #81

PR #81, opened 2016-08-15
Add `brew cask` install instructions for macOS

dgoodlad:
I added a formula for `aws-vault` to homebrew cask in https://github.com/caskroom/homebrew-cask/pull/23637. I _propose_ that the README should talk about that as an installation method since many folks just aren't used to or comfortable with manually dropping binaries in place, setting permissions correctly, etc.

I'm not wedded to this at all, though, so feel free to close if you disagree :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ Currently OSX/Keychain and Linux/KWallet are supported, support for Linux's libs
 
 Download the [latest release](https://github.com/99designs/aws-vault/releases). The OSX release is code-signed, and you can verify this with `codesign -dvvv aws-vault`.
 
+On macOS, you may instead use [homebrew cask](https://github.com/caskroom/homebrew-cask) to install that same code-signed binary:
+
+    brew cask install aws-vault
+    # Verify the code-signed binary
+    codesign -dvvv $(which aws-vault)
+
 ## Usage
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -15,7 +15,8 @@ On macOS, you may instead use [homebrew cask](https://github.com/caskroom/homebr
 
 The macOS release is code-signed, and you can verify this with `codesign`:
 
-    codesign -dvvv $(which aws-vault)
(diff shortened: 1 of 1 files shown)


joho:
I don't feel strongly either way. Will defer to @lox

pda:
Seems good to me; as you mentioned in the linked PR there were various issues around getting aws-vault into homebrew with code signing, and you seem to have got it in the right place for those issues to be resolved now.

* Linked PR: https://github.com/caskroom/homebrew-cask/pull/23637
* Previous one: https://github.com/Homebrew/homebrew-core/pull/3385#issuecomment-236302100
* Previous previous one: https://github.com/Homebrew/homebrew-binary/pull/317

Might be premature on the “macOS” vs “OS X” but at least it's more future proof. Maybe `s/OSX/macOS/` the two existing references for consistency?

joho:
Actually we should reference both osx & macos in the readme so that anyone googling "osx amazon credentials" or "macos aws key" etc will match against the library.

dgoodlad:
@pda thanks for linking to the history

> Might be premature on the “macOS” vs “OS X” but at least it's more future proof. Maybe `s/OSX/macOS/` the two existing references for consistency?

Yeah I've seen many projects making this change, and installed the beta yesterday so it was just in my head as the default.  I agree with @joho, it's probably a good idea to have both in the docs for some hot SEO action 😉

pda:
Then change one of the “OSX”[sic] to “Mac OS X” for even hotter SEO action? ;)

dgoodlad:
Just pushed a couple commits, clarifying both the macOS/OSX naming and split up the code signature verification from `brew cask install`

lox:
Awesome! Nice one @dgoodlad! This will also close #71. 

If we are going to recommend people check the code signing (which we should), we should also recommend that they verify it's signed by 99designs by looking for:

`Authority=3rd Party Mac Developer Application: 99designs Inc (NRM9HVJ62Z)`

dgoodlad:
@lox done!

:zap:

PR #85, opened 2016-08-21
Use `.keychain-db` extension on macOS Sierra

### LINKED ISSUE #84: Keychain file names changed on macOS 10.12 (beta)
The names given to keychains in `~/Library/Keychains` have changed to have the extension `.keychain-db` rather than `.keychain` on beta versions of macOS 10.12.

While running `aws-vault add` for the first time works correctly, subsequent operations (e.g. `aws-vault exec`) will not find credentials, and will never even request access to the keychain file. The file that was created is `~/Library/Keychains/aws-vault.keychain-db`, despite debug output during `aws-vault add` saying it was opening `~/Library/Keychains/aws-vault.keychain`.

I had a poke around and modified [keychain.go:44](https://github.com/99designs/aws-vault/blob/42854c4d7d2af53a0978244ee20db25038312989/keyring/keychain.go#L44) to open the `.keychain-db` file instead. This works, but not consistently. `aws-vault add` now explodes.

To work around the issue, I ended up reverting that change, and symlinked `.keychain-db` to `.keychain`:

```
$ ls -l ~/Library/Keychains/aws-vault*
lrwxr-xr-x  1 dgoodlad  staff     21 20 Aug 14:07 /Users/dgoodlad/Library/Keychains/aws-vault.keychain -> aws-vault.keychain-db
-rw-r--r--  1 dgoodlad  staff  24668 20 Aug 14:50 /Users/dgoodlad/Library/Keychains/aws-vault.keychain-db
```

I'm not nearly familiar enough with what's going on here to work out the issue quickly, so I'm opening this issue to at least document what I know.

dgoodlad:
As of macOS Sierra, Keychain files are stored with the extension `.keychain-db` rather than `.keychain`. This is a quick kernel-version check so that we use the correct extension on the user's OS.

Doing the version check via the `kern.osrelease` sysctl call isn't the "most correct" thing to do on macOS. The options are:

* Use the objective-c `isOperatingSystemAtLeastVersion` function from the Foundation framework. This would need me to know how to allocate and fill a `NSOperatingSystemVersion` struct full of `NSNumber` fields, properly allocate those fields, and manually memory manage them, so after fumbling with it for a bit I went looking elsewhere
* Shell out to `sw_vers` 😐 

I'm happy enough with this sysctl version; it is reliable, just not "obvious" that kernel version 16.x.x is macOS 10.12

Fixes #84

### DIFF
--- a/keyring/keychain.go
+++ b/keyring/keychain.go
@@ -20,6 +20,9 @@ import (
 	"log"
 	"os"
 	"os/user"
+	"strconv"
+	"strings"
+	"syscall"
 	"unicode/utf8"
 	"unsafe"
 )
@@ -30,18 +33,44 @@ type keychain struct {
 	Passphrase string
 }
 
+func keychainPath(name string) (string, error) {
+	usr, err := user.Current()
+	if err != nil {
+		return "", err
+	}
+
+	// As of macOS Sierra, Keychain files are stored with the `.keychain-db`
+	// extension, rather than `.keychain`. The result of the kern.osrelease
+	// sysctl is the kernel version number in the form "major.minor.patch",
+	// and Sierra is the first macOS release to use kernel major version 16
+	osver, err := syscall.Sysctl("kern.osrelease")
+	if err != nil {
+		return "", err
+	}
+
+	major, err := strconv.Atoi(strings.Split(osver, ".")[0])
+	if err != nil {
+		return "", err
+	}
+
+	if major >= 16 {
+		return usr.HomeDir + "/Library/Keychains/" + name + ".keychain-db", nil
+	} else {
+		return usr.HomeDir + "/Library/Keychains/" + name + ".keychain", nil
+	}
+}
+
 func init() {
 	supportedBackends[KeychainBackend] = opener(func(name string) (Keyring, error) {
 		if name == "" {
 			name = "login"
 		}
 
-		usr, err := user.Current()
+		path, err := keychainPath(name)
 		if err != nil {
 			return nil, err
 		}
-
-		return &keychain{Path: usr.HomeDir + "/Library/Keychains/" + name + ".keychain", Service: name}, nil
+		return &keychain{Path: path, Service: name}, nil
 	})
 
 	DefaultBackend = KeychainBackend


joho:
LGTM. Should we hold off a little while on cutting a release, wait and see if any other Sierra issues shake out?

dgoodlad:
@joho I'm easy; my local build's working fine for me, without any codesigning either.

It's up to you folks as the Signing-Keymasters when you want to release 😉  My feeling is that we wait for Apple to cut an RC of Sierra, then we can be pretty confident there won't be any major breakages for `aws-vault`, and we can ship then.

joho:
Yeah I reckon RC time is a good point to cut a release. Though this is backwards compatible, so if we released for other reasons it's NBD.

dgoodlad:
> Though this is backwards compatible, so if we released for other reasons it's NBD.

Yeah; the way I see it, these fixes can land in master because they're safe to release if some other work needs to ship, but don't need to trigger a release on their own.

dgoodlad:
Ping @lox @pda for another set of 👀  on my obviously Top Quality™ hacks

pda:
LGTM and Works On My Machine. OS X 10.11.6 (15G31), Darwin Kernel Version 15.6.0.

lox:
I'm not stoked at the idea of shelling out to `sw_vers`. I'd like to understand the root cause a bit before rolling this into a release. If we absolutely have to, I'd prefer to simply check for the `.keychain-db` version first and then look for the older `.keychain` if it doesn't exist.

dgoodlad:
@lox shelling out was an option, but one I discarded. The implementation I went with is a `sysctl` call to `kern.osrelease`, instead...

You're right, we _could_ look for `~/Library/Keychains/login.keychain{,-db}` to work it out, too. I just don't know how much it's worth doing it that way over the os version check. If you still vote to do it that way over the `sysctl` method, it's easy to do, but want to make sure you still feel as strongly when it's not comparing to shelling out.

lox:
Yeah, fair enough. I can live with that @dgoodlad. Probably best to get this merged before the new version drops soon.

dgoodlad:
🚢ed

I'll need one of you fine 99designs folks to sign & tag a new release, though...

mtibben:
So just update to sierra and started using aws-vault master to get this change. But I get
```
$ aws-vault ls
2016/09/21 12:05:35 Opening keychain /Users/michael/Library/Keychains/aws-vault.keychain-db
No credentials found
```

In the keychain directory we have both `.keychain` and `.keychain-db` files.

Opening either file in the Keychain Access app shows the same thing. The aws-vault keychain is also still visible in the Keychain Access App. 

Using the previous 3.4 release of aws-vault works fine for me with the existing keychain. 

So I assume the problem you were solving is only related to _new_ keychains @dgoodlad.

We're gonna need a migration path for existing keychains. Either move the existing keychain file to the new file extension? Or just continue to use the existing?

mtibben:
Nope, new keychains are working fine for me using 3.4 aws-vault. @dgoodlad can you explain what issue this PR was trying to solve?

pda:
Seems it's a leaky abstraction to use `os.Stat` to see if a keychain exists but then [`SecKeychainOpen`](https://developer.apple.com/reference/security/1396431-seckeychainopen) to open it — I'm guessing Apple added some backwards-compatibility to `SecKeychainOpen` since @dgoodlad wrote this patch against the Sierra beta, and now the 1:1 mapping between paths on disk and paths that `SecKeychainOpen` can open is broken. There doesn't seem to be a `SecKeychainStat` or equivalent. There is a [`SecKeychainGetPath`](https://developer.apple.com/reference/security/1401073-seckeychaingetpath) but I believe that's only useful after you have opened the keychain.

Rather than testing for existence with `os.Stat` can we just call `SecKeychainOpen` and correctly handle the [result codes](https://developer.apple.com/reference/security/1658642-keychain_services#1664539) it returns?

mtibben:
yep gotcha - it's the `os.Stat`s that @dgoodlad would have been running into.

lox:
So revert and new release?

mtibben:
well the problem still remains - we're using `os.Stat` to determine whether the keychain exists, however macOS Sierra will map the filename internally in the api if the keychain path doesn't end with `-db`, depending on whether the given path exists or not. We can either do the same thing (if the `.keychain` file exists, use it rather than `.keychain-db`) OR we could remove our usage of `os.Stat` and depend on the keychain API to tell us when a keychain does or doesn't exist

lox:
The latter would seem to make the most sense.

mtibben:
Yeah, who is familiar with that code? looks a bit scary.

Here's a branch with the first quick-fix option: [use-keychain-if-exists](https://github.com/99designs/aws-vault/commit/use-keychain-if-exists)

lox:
I mean, I wrote it... but familiar is a stretch.

lox:
I'll dive in now.

lox:
Reading this lovely code: http://opensource.apple.com/source/libsecurity_keychain/libsecurity_keychain-37184/lib/TrustAdditions.cpp?txt

Looks like we need to implement [`SecKeychainGetStatus`](https://developer.apple.com/reference/security/1399085-seckeychaingetstatus?language=objc)

ndaversa:
This fixes seems broken in the 3.5.0 release.

I built from this point and it works fine on Sierra

PR #86, opened 2016-09-13
Makefile check doesn't handle golang 1.6 and 1.7

lox:
This leads to broken binaries if codesigning is used after building with golang 1.6 or 1.7.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,21 +4,15 @@ PREFIX=github.com/99designs/aws-vault
 GOVERSION=$(shell go version)
 GOBIN=$(shell go env GOBIN)
 VERSION=$(shell git describe --tags --candidates=1 --dirty)
-FLAGS=-v -X main.Version=$(VERSION) -s
+FLAGS=-X main.Version=$(VERSION) -s
 CERT="3rd Party Mac Developer Application: 99designs Inc (NRM9HVJ62Z)"
 
 build:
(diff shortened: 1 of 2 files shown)


lox:
This means `make build` will no longer for for golang 1.4.

pda:
Oops.

pda:
> This means make build will no longer for for golang 1.4.

Seems fine.

Tangent: it bugs me that `Makefile` unconditionally attempts code signing on Darwin; most users of `Makefile` aren't going to have the cert, but it can be useful to `make build` instead of `go …` to get the LDFLAGS / version number etc. Perhaps code signing could be a separate make target, or enabled with an env var?

lox:
You make an excellent point. See most recent commit.

lox:
Merge?

pda:
Works for me on go 1.6.2 and OS X 10.11.6:

```
pda@paulbook ~/Projects/99designs/aws-vault ❯ make
go build -o aws-vault -ldflags="-X main.Version=v3.4.0-19-gc7c781d -s" github.com/99designs/aws-vault
```

and `make sign` seems to work other than not having the cert;

```
pda@paulbook ~/Projects/99designs/aws-vault ❯ make sign
codesign -s "3rd Party Mac Developer Application: 99designs Inc (NRM9HVJ62Z)" ./aws-vault
3rd Party Mac Developer Application: 99designs Inc (NRM9HVJ62Z): no identity found
make: *** [sign] Error 1
```

👍

PR #88, opened 2016-09-21
Check keychain exists with SecKeychainGetStatus

lox:
This reverts #85, because it turns out to not actually work under macOS Sierra. 

This change uses the `SecKeychainGetStatus` to see if the keychain exists. I suspect we might also need to drop extensions from paths.

### DIFF
--- a/keyring/keychain.go
+++ b/keyring/keychain.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"os"
 	"os/user"
 	"unicode/utf8"
 	"unsafe"
@@ -48,10 +47,6 @@ func init() {
 }
 
 func (k *keychain) Get(key string) (Item, error) {
-	if _, err := os.Stat(k.Path); os.IsNotExist(err) {
-		return Item{}, ErrKeyNotFound
-	}
-
 	serviceRef, err := _UTF8StringToCFString(k.Service)
 	if err != nil {
 		return Item{}, err
@@ -121,7 +116,7 @@ func (k *keychain) Set(item Item) error {
 	var kref C.SecKeychainRef
 	var err error
 
-	if _, err := os.Stat(k.Path); os.IsNotExist(err) {
+	if exists, _ := keychainExists(k.Path); !exists {
 		var prompt = true
 		if k.Passphrase != "" {
 			prompt = false
@@ -135,6 +130,7 @@ func (k *keychain) Set(item Item) error {
 	} else {
 		kref, err = openKeychain(k.Path)
 		if err != nil {
+			log.Printf("error opening %v, %v", err, k.Path)
 			return err
 		}
 		defer C.CFRelease(C.CFTypeRef(kref))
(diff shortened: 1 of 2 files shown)


lox:
Feels @dgoodlad @mtibben @pda?

lox:
After this I can use my previously created keychains just fine, haven't tested creating them.

keyring/keychain.go
```diff
@@ -323,6 +286,27 @@ func (k *keychain) Keys() ([]string, error) {
...
+func keychainExists(path string) (bool, error) {
+	// returns no error even if it doesn't exist
+	k, err := openKeychain(path)
+	if err != nil {
+		log.Printf("Error opening %v", err)
+		return false, err
+	}
+	defer C.CFRelease(C.CFTypeRef(k))
+
+	var status C.SecKeychainStatus
```
lox: I'm pretty sure that SecKeychainStatus doesn't need to be explicitly released here, it's not a ref.

keyring/keychain.go
```diff
@@ -33,54 +29,24 @@ type keychain struct {
...
 		}
 
-		path, err := keychainPath(name)
+		usr, err := user.Current()
 		if err != nil {
 			return nil, err
 		}
-		return &keychain{Path: path, Service: name}, nil
+
+		return &keychain{Path: usr.HomeDir + "/Library/Keychains/" + name + ".keychain", Service: name}, nil
```
mtibben: Just making sure you realise new keychains will still be created with `.keychain-db` on Sierra. So the `Path` won't actually be accurate, might be a future source of bugs

lox: Yup, should change it so it's not exposed.

lox: Done.

lox:
If anyone had the beta still installed, would be awesome to check this works.

dgoodlad:
@lox works for me!

lox:
Ship this at your leisure @mtibben.

mtibben:
Released [3.5.0](https://github.com/99designs/aws-vault/releases/tag/v3.5.0)

stevehodgkiss:
3.5.0 isn't working on Sierra for me. This PR seems like it could be the likely cause. Steps taken (after removing the old keychain file):

`aws-vault add my-profile` asks for a new password for the vault, and appears to work fine. It creates `aws-vault.keychain-db` (not `aws-vault.keychain` like the log output indicates):

```
❯ aws-vault add my-profile
Enter Access Key ID: [snip]
Enter Secret Access Key: [snip]
2016/09/26 14:11:34 Opening keychain /Users/steve/Library/Keychains/aws-vault.keychain
2016/09/26 14:11:34 Creating keychain /Users/steve/Library/Keychains/aws-vault.keychain (prompt true)
2016/09/26 14:11:42 Adding service="aws-vault", account="my-profile" to osx keychain /Users/steve/Library/Keychains/aws-vault.keychain
Added credentials to profile "my-profile" in vault
2016/09/26 14:11:42 Parsing config file /Users/steve/.aws/config
2016/09/26 14:11:42 Opening keychain /Users/steve/Library/Keychains/aws-vault.keychain
```

Then trying to use it fails. It doesn't ask for my vault password at this stage:

```
❯ aws-vault exec a-profile bash
2016/09/26 14:11:48 Parsing config file /Users/steve/.aws/config
2016/09/26 14:11:48 Looking up keyring for my-profile
2016/09/26 14:11:48 Opening keychain /Users/steve/Library/Keychains/aws-vault.keychain
2016/09/26 14:11:48 Error from keyring The user name or passphrase you entered is not correct.
No credentials found for profile "a-profile"
```

I've build `1c5c5f2` locally and that works fine (after https://github.com/99designs/aws-vault/pull/85 was merged).

Issue #70, opened 2016-07-24
bash completion of exec command


jedi4ever:
we use `make` a lot at work. It has nice bashcompletion for it's tasks.

Execing it with `aws-vault exec <profile> make` sure works great! thanks for the plugin!
But it breaks the completion.

I researched and came across `_command_offset`in bash_completion.
It is used by `sudo` to achieve something similar.

Maybe it can inspire you :)

lox:
Yup, I've wanted to do this for a while. Unfortunately I'm a zsh user, but I think it can be done such that it's portable.

lox:
Apparently this is somewhat helped by the cli library we use: https://github.com/alecthomas/kingpin#bashzsh-shell-completion

jedi4ever:
@lox this build in completion will only help for the aws-vault options
afaik this will not pass commandcompletion to the execed command
for that the command offset is needed , but I guess it's a good start to hack on

jedi4ever:
I tried the build in `eval "$(aws-vault --completion-script-bash)"`
but it says --completion-script-bash option is not found.

I've made some progress on the completion of the exec command.

`aws-vault exec profile1 -- make <TAB>` now expands correctly the make tasks

I'll continue to enhance it with the aws-vault options.

```
# bash completion for aws-vault                             -*- shell-script -*-

_aws_vault()
{
    local cur prev words cword

    local i

    cur="${COMP_WORDS[COMP_CWORD]}"
    prev="${COMP_WORDS[COMP_CWORD-1]}"

    for (( i=1; i <= COMP_CWORD; i++ )); do
        if [[ ${COMP_WORDS[i]} == -- ]]; then
            local root_command=${COMP_WORDS[i]}
            _command_offset $i+1
            return
        fi
    done

} && complete -F _aws_vault aws-vault
```

lox:
Nice! I've been working on getting the latest cli stuff into aws-vault so that we can at least use kingpin to output the above script when `--completion-script-bash` is passed.

PR #89, opened 2016-09-21
WIP: Bash and ZSH autocompletion templates

lox:
See #70 for background, this adds custom templates for `aws-vault --completion-script-bash`.

See https://github.com/alecthomas/kingpin#bashzsh-shell-completion for more details.

### DIFF
--- a/config.go
+++ b/config.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


lox:
@jedi4ever are you still using the completions that you wrote in #70?

lox:
Gonna close this due to lack of interest.

pecigonzalo:
@lox I would also be really interested on this, whats missing? seems like completion sort of works, but not for sub-commands.

Issue #77, opened 2016-08-08
AWS Console login SessionDuration


pda:
I've been trying to add `--session-duration` flag to `aws-vault login`, but hit a snag where adding the `SessionDuration` parameter to the federated login causes the following error:

> message Invalid credentials parameter
> description The request sent by the client was syntactically incorrect.

It's exactly as described a few days ago by somebody else, so perhaps a recent problem with AWS API:
https://forums.aws.amazon.com/thread.jspa?threadID=236788

The documentation at http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_enable-console-custom-url.html suggests it should be possible to to set SessionDuration to up to 12 hours, regardless of the duration of the assumed role token being used to get the login token.

Below is my WIP, run out of time, and not ready to PR it without seeing it working against the API. Note also I based it on the older v3.4.0 config structure because `master` @ 69bf23b921017cde9f147d6e1241d3664ea666a1 is broken by https://github.com/99designs/aws-vault/pull/74

```diff
diff --git a/login.go b/login.go
index 033210b..57e239a 100644
--- a/login.go
+++ b/login.go
@@ -7,6 +7,8 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"strconv"
+	"time"
 
 	"github.com/99designs/aws-vault/keyring"
 	"github.com/99designs/aws-vault/prompt"
@@ -16,11 +18,12 @@ import (
 )
 
 type LoginCommandInput struct {
-	Profile   string
-	Keyring   keyring.Keyring
-	MfaToken  string
-	MfaPrompt prompt.PromptFunc
-	UseStdout bool
+	Profile         string
+	Keyring         keyring.Keyring
+	MfaToken        string
+	MfaPrompt       prompt.PromptFunc
+	UseStdout       bool
+	SessionDuration time.Duration
 }
 
 func LoginCommand(ui Ui, input LoginCommandInput) {
@@ -59,14 +62,23 @@ func LoginCommand(ui Ui, input LoginCommandInput) {
 
 	q := req.URL.Query()
 	q.Add("Action", "getSigninToken")
+	if input.SessionDuration != 0 {
+		q.Add("SessionDuration", strconv.Itoa(int(input.SessionDuration.Seconds())))
+	}
 	q.Add("Session", string(jsonBytes))
 	req.URL.RawQuery = q.Encode()
 
+	ui.Debug.Printf("GET %s\n", req.URL)
+
[...]

PR #92, opened 2016-09-22
Use iam credentials for login, allowing session expiry of 12h

lox:
Based on feedback in this [aws forum thread](https://forums.aws.amazon.com/thread.jspa?messageID=742905&#742905), it became obvious that the new longer federated session expiry could be used if IAM credentials were used to assume the role, vs our standard iam -> session token -> assume role -> federation tokens. 

I'm not sure why I never realized that the session token step was unneeded. I suspect it was because it adds complexity to the code, as you need to conditionally apply MFA to the AssumeRole call if there isn't a session with MFA already applied to it.

Regardless, this works, although I've been drinking a lot of wine, so I might be imagining it all.

### DIFF
--- a/main.go
+++ b/main.go
@@ -170,6 +170,16 @@ func configureLoginCommand(app *kingpin.Application, ui Ui, g *globalFlags) {
 		Short('t').
 		StringVar(&input.MfaToken)
 
+	cmd.Flag("federation-token-ttl", "Expiration time for aws console session").
+		Default("12h").
+		OverrideDefaultFromEnvar("AWS_FEDERATION_TOKEN_TTL").
+		Short('f').
+		DurationVar(&input.FederationTokenDuration)
+
+	cmd.Flag("assume-role-ttl", "Expiration time for aws assumed role").
+		Default("15m").
+		DurationVar(&input.AssumeRoleDuration)
+
 	cmd.Flag("stdout", "Print login URL to stdout instead of opening in default browser").
 		Short('s').
 		BoolVar(&input.UseStdout)
--- a/exec.go
+++ b/exec.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"log"
 	"os"
 	"os/exec"
 	"strings"
@@ -11,7 +10,6 @@ import (
 	"github.com/99designs/aws-vault/keyring"
 	"github.com/99designs/aws-vault/prompt"
 	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/aws/aws-sdk-go/aws/credentials"
 )
 
 type ExecCommandInput struct {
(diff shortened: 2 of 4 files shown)


lox:
Fixes #77.

mtibben:
👍  lgtm

lox:
I don't have MFA for my test accounts, so would appreciate some testing with that.

lox:
For posterity, this was my monologue last night:

```

Y’know, when it comes down to it, why do we even create Sessions if there isn’t an MFA on it?

[7:09]  
Basically the aim of aws-vault is to just not expose IAM credentials to the console or downstream apps.

[7:09]  
The mechanism it does that by is creating temporary credentials with STS. (edited)

[7:10]  
So the options are either GetSessionToken, or AssumeRole.

[7:10]  
If there is a role to be assumed, it should just use AssumeRole, otherwise GetSessionToken.

[7:11]  
I see no good reason to go IAM -> GetSessionToken -> AssumeRole

[7:11]  
I guess the only reason might be that you might want to use MFA to create a long-ish session, which you can then assume roles from, because AssumeRole is capped at 1 hour.

[7:13]  
Currently aws-vault caches session tokens in the keychain, but not assumed role sessions.

[7:13]  
So it’s probably worth keeping that session creating behaviour, except for in login.

[7:13]  
Good talk.
```

main.go
```diff
@@ -170,6 +170,16 @@ func configureLoginCommand(app *kingpin.Application, ui Ui, g *globalFlags) {
...
 		StringVar(&input.MfaToken)
 
+	cmd.Flag("federation-token-ttl", "Expiration time for aws console session").
+		Default("12h").
+		OverrideDefaultFromEnvar("AWS_FEDERATION_TOKEN_TTL").
+		Short('f').
+		DurationVar(&input.FederationTokenDuration)
+
+	cmd.Flag("assume-role-ttl", "Expiration time for aws assumed role").
+		Default("15m").
```
mtibben: This may have been discussed previously - the aws cli defaults to 1 hour, should we match it?

> The temporary security credentials are valid for the duration that you specified when calling assume-role , which can be from 900 seconds (15 minutes) to a maximum of 3600 seconds (1 hour). The default is 1 hour.

\- http://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html

mtibben:
I've tested with MFA and also `--server` and haven't had any issues

main.go
lox: Yeah, I think that probably makes sense.

lox:
Nice.

pda:
Seems to work for me.

```
❯ ./aws-vault login redacted-admin
2016/09/23 14:23:39 Parsing config file /Users/pda/.aws/config
2016/09/23 14:23:39 Looking up keyring for 99designs
2016/09/23 14:23:39 Opening keychain /Users/pda/Library/Keychains/aws-vault.keychain
Enter token for arn:aws:iam::redacted:mfa/redacted: 
2016/09/23 14:23:46 Assuming role arn:aws:iam::redacted:role/redacted with iam credentials
2016/09/23 14:23:48 Using role ****************redacted, expires in 14m59.534745467s
2016/09/23 14:23:48 Creating federation login token, expires in 12h0m0s
```

lox:
Ship it at your leisure then.

PR #98, opened 2016-10-01
Added the freedesktop.org's Secret Service backend.

gsterjov:
This PR adds support for the secrets service used by Gnome (and KDE as well I think).
I pulled the actual communication with DBus into a separate go package since it seemed useful as a library in itself (https://github.com/gsterjov/go-libsecret).

Its working great for me, so I'd love to see this go into upstream :smile:

### DIFF
--- a/keyring/keyring.go
+++ b/keyring/keyring.go
@@ -4,6 +4,7 @@ import "errors"
 
 const (
 	KeychainBackend string = "keychain"
+	SecretsBackend  string = "secrets"
 	KWalletBackend  string = "kwallet"
 	FileBackend     string = "file"
 )
--- a/keyring/keyring.go
+++ b/keyring/keyring.go
@@ -3,10 +3,10 @@ package keyring
 import "errors"
 
 const (
-	KeychainBackend string = "keychain"
-	SecretsBackend  string = "secrets"
-	KWalletBackend  string = "kwallet"
-	FileBackend     string = "file"
+	SecretServiceBackend  string = "secret-service"
+	KeychainBackend       string = "keychain"
+	KWalletBackend        string = "kwallet"
+	FileBackend           string = "file"
 )
 
 var DefaultBackend = FileBackend
--- a/keyring/secrets.go
+++ b/keyring/secrets.go
@@ -9,19 +9,19 @@ import (
   "encoding/json"
 )
 
-const SECRETS_COLLECTION = "awsvault"
+const secretsCollection = "awsvault"
 
 
 func init() {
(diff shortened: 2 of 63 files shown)


gsterjov:
Related issue: https://github.com/99designs/aws-vault/issues/23

keyring/keyring.go
```diff
@@ -4,6 +4,7 @@ import "errors"
 
 const (
 	KeychainBackend string = "keychain"
+	SecretsBackend  string = "secrets"
```
lox: LibSecretBackend might be clearer here?

keyring/secrets.go
```diff
@@ -0,0 +1,239 @@
...
+package keyring
+
+import (
+  "github.com/gsterjov/go-libsecret"
+  "fmt"
+  "os"
+  "encoding/json"
+)
+
+const SECRETS_COLLECTION = "awsvault"
```
lox: Go style is `secretsCollection` or `SecretsCollection` if it's supposed to be exported.

keyring/secrets.go
```diff
@@ -0,0 +1,239 @@
...
+
+
+func init() {
+  service, err := libsecret.NewService()
```
lox: This should be moved into the opener below, otherwise it will result in errors on any system without libsecret.

keyring/secrets.go
```diff
@@ -0,0 +1,239 @@
...
+
+type secretsKeyring struct {
+  name        string
+  service     *libsecret.Service
+  collection  *libsecret.Collection
+  session     *libsecret.Session
+}
+
+
+type SecretsError struct {
```
lox: Does this type need to be exported?

lox:
For comparison, here is Docker's implementation: https://github.com/docker/docker-credential-helpers/tree/master/secretservice

keyring/keyring.go
gsterjov: Was thinking about it but its not actually using libsecret since it goes through dbus via godbus. Couldn't think of a better name than libsecret for the separate go package since there are already so many named *secret

lox: Fair point, might `SecretServiceBackend` be better then?

gsterjov: yea it would, nice one

gsterjov:
Thanks for the review. Made the changes.
I'm guessing I'd have to add `github.com/gsterjov/go-libsecret` to the glide.yml for the linux test to pass?

lox:
Yup, and commit it to the vendors folder.

PR #99, opened 2016-10-07
add.go, rm.go: use log.Printf vs fmt.Printf to ensure newline.

pda:
```
$ aws-vault rm --sessions-only redacted
Deleted 11 sessions.💀
```

The skull (if it renders properly here) is my shell saying there was no newline at the end of the command output.  `log.Println` and `log.Printf` seem to be used elsewhere, but `fmt.Printf` was used here; fixed.

### DIFF
--- a/add.go
+++ b/add.go
@@ -48,7 +48,7 @@ func AddCommand(app *kingpin.Application, input AddCommandInput) {
 		return
 	}
 
-	fmt.Printf("Added credentials to profile %q in vault", input.Profile)
+	fmt.Printf("Added credentials to profile %q in vault\n", input.Profile)
 
 	profiles, err := awsConfigFile.Parse()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lox:
https://github.com/99designs/aws-vault/blob/master/main.go#L220 :(

pda:
Actually no, that moves it from `stdout` to `--debug` output. I'll just put some `\n`s into the `fmt` output.

lox:
Tempted to ditch that behaviour, it's got a high WTF factor.

pda:
Fixed to use `fmt.Printf("…\n")`

> Tempted to ditch that behaviour, it's got a high WTF factor.

Yeah a more explicit debug vs warning vs log output mechanism would perhaps be good. But using a `log` singleton is never going to work for all cases (unless it has priority-flagged logging).

PR #100, opened 2016-10-17
Replace glide with govendor

lox:
I'm moving away from glide due to some major issues with it's workflow:

1. Atomically adding a new package isn't possible at this stage, other packages get updated
2. It's very difficult to reconcile what is in your vendor/ and what isn't, which makes for things getting out of date
3. It's optimized for non-committed vendor/ dirs, which are dead to me
4. It's another set of dot files at the project root

[Govendor](https://github.com/kardianos/govendor) on the other hand is much more go-like and addresses all of the above issues gracefully.


joho:
I just read through https://github.com/kardianos/govendor/blob/master/doc/whitepaper.md and have no beef with this change.

Issue #104, opened 2016-12-11
Login fails for IAM credential based profiles


lox:
It seems the creating a federated login only works when you have assumed a role.

lox:
Fixed in #105

PR #105, opened 2016-12-11
Show error when using a profile without source_profile

lox:
Show a useful error rather than "400 Bad Request"


lox:
Until #104 is resolved, this at least provides an error.

PR #106, opened 2016-12-21
Allow console login for IAM Credentials without AssumeRole

lox:
Currently trying to login with IAM credentials doesn't work, it fails with Bad Request. 

This uses GetFederationToken to make it work.


lox:
Closes #104.

PR #107, opened 2016-12-28
exec: Exit with exit code of subcommand

hraban:
Code was already there but the order of if blocks made this block
unreachable.

Demo:

```
$ ~/go3d/bin/aws-vault exec mykey -- bash -c 'exit 2'
$ echo $?
2
$ /usr/local/bin/aws-vault exec mykey -- bash -c 'exit 2'
aws-vault: error: exit status 2
$ echo $?
0
```


lox:
Many thanks, will get this in the next release

Issue #112, opened 2017-02-01
Invalid handle error on Windows


porty:
# Problem

When entering access credentials, `aws-vault` exits with the error message "The handle is invalid"

```
C:\Users\shorty>aws-vault add test
Enter Access Key ID: AKIAAAAAAAAAAAAAAAAAAAAA
Enter Secret Access Key: zzzzzzzzzzzzzzzzzzzzzzz
Enter passphrase to unlock C:\Users\shorty\.awsvault\keys: aws-vault: error: The handle is invalid.
```

# Expected Result

aws-vault accepts password input and exists successfully.

# Environment Details

OS: Windows 10 x64 build 15019
Go version: 1.7.3
Built from tag v3.7.0
Using standard Windows Command prompt

lox:
Did you fix this in #113 @porty?

porty:
Yep, said so up the top

> Fix for #112

zhkostadinov:
The issue is still there on Windows when you try to do that with bash windows. 
The workaround is to use PowerShell with admin privileges.

PR #113, opened 2017-02-01
Fix terminal input in windows

porty:
Fix for #112 

The file descriptor passed to `terminal.ReadPassword` is hardcoded to `1` which works fine for UNIX systems but is not correct for a Windows environment. Setting this to the actual file descriptor of stdout fixes the problem.

```
C:\Users\shorty>aws-vault add test
Enter Access Key ID: AKIAAAAAAAAAAAAAAAAAAAAA
Enter Secret Access Key: zzzzzzzzzzzzzzzzzzzzzzz
Enter passphrase to unlock C:\Users\shorty\.awsvault\keys:
Added credentials to profile "test" in vault
```

Also fixed nearby code that did nothing with error values.

### DIFF
--- a/keyring/file.go
+++ b/keyring/file.go
@@ -55,12 +55,12 @@ func (k *fileKeyring) dir() (string, error) {
 
 	stat, err := os.Stat(dir)
 	if os.IsNotExist(err) {
-		os.MkdirAll(dir, 0700)
+		err = os.MkdirAll(dir, 0700)
 	} else if err != nil && !stat.IsDir() {
 		err = fmt.Errorf("%s is a file, not a directory", dir)
 	}
 
-	return dir, nil
+	return dir, err
 }
 
 func (k *fileKeyring) unlock() error {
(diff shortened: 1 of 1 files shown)


joho:
This looks like a good change to me. Right now the travis builds do mac & linux, should we add in a windows build too? Does travis even do windows?

porty:
Windows CI would be good, but in this case it wouldn't have picked anything up anyway

lox:
No, Travis doesn't do Windows. AppVeyor is the only one I know of.

lox:
Nice work btw @porty

lox:
I'm going to merge this.

proffesboris:
Guys, there is still bug when you try to login via Git Bash

umairdojki:
Thanks for suggesting the workaround @domhdp !

Are there any plans on fixing this proper for Git Bash? A colleague is on 5.4.4 and they don't get this error but I do on 6.3.0

PR #117, opened 2017-03-14
Update jose2go dep

porty:
Update the jose2go dep based on http://blogs.adobe.com/security/2017/03/critical-vulnerability-uncovered-in-json-encryption.html

Not sure if Elliptic Curve cryptography is actually used, but better safe than sorry.

```shell
go get -u github.com/dvsekhvalnov/jose2go
govendor update github.com/dvsekhvalnov/jose2go
```


mtibben:
might be a good opportunity to bump all our deps

PR #118, opened 2017-03-16
Support linux for exec --server

adrienkohlbecker:
Hello!

This PR makes `exec`'s server mode compatible with linux. The syntax for the alias installation is different on this platform.
To do so, I extracted the `installNetworkAlias` function in platform specific files. I also added an explicit error message on windows.

Best,

### DIFF
--- a/alias_darwin.go
+++ b/alias_darwin.go
@@ -1,0 +1,9 @@
+// build: darwin
+
+package main
+
+import "os/exec"
+
+func installNetworkAlias() ([]byte, error) {
+	return exec.Command("ifconfig", "lo0", "alias", "169.254.169.254").CombinedOutput()
+}
--- a/alias_linux.go
+++ b/alias_linux.go
@@ -5,5 +5,5 @@ package main
 import "os/exec"
 
 func installNetworkAlias() ([]byte, error) {
-	return exec.Command("ifconfig", "lo:0", "169.254.169.254").CombinedOutput()
+	return exec.Command("ip", "addr", "add", "169.254.169.254/24", "dev", "lo", "label", "lo:0").CombinedOutput()
 }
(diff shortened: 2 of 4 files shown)


joho:
Thanks, that looks really helpful.

@Vektah reckon you could do a build/test on your linux setup when you get in this morning just to sanity check it?

Vektah:
I get
```
roci :: ~ » aws-vault server
aws-vault: error:
aws-vault: error: exec: "ifconfig": executable file not found in $PATH
```

Like most things on linux there are at least 2 ways to do things:
 - `ipconfig`, old `net-tools` super binary that's being removed from more and more distros (centos 7 and arch that I know of)
 - `ip`, newer modular tool with new syntax that not many understand.

I think with `ip` the command would be `ip addr add 169.254.169.254 dev lo label lo:0`

I'm not sure of the exact history but `ip` has been around since centos 5, so its probably a better target:
```
roci :: 99designs/contests » docker run -it centos:5 bash
[root@12691d7431c8 /]# ip
Usage: ip [ OPTIONS ] OBJECT { COMMAND | help }
       ip [ -force ] [-batch filename
where  OBJECT := { link | addr | addrlabel | route | rule | neigh | ntable | tunnel |
                   maddr | mroute | monitor | xfrm }
       OPTIONS := { -V[ersion] | -s[tatistics] | -r[esolve] |
                    -f[amily] { inet | inet6 | ipx | dnet | link } |
                    -o[neline] | -t[imestamp] }
```

adrienkohlbecker:
Thanks for the review!
Updated the PR to use `ip`

joho:
No problem, thanks again for the submission.

adrienkohlbecker:
Any chance for an official release?

PR #119, opened 2017-03-23
Verify vendor

Vektah:
Currently PR's like https://github.com/99designs/aws-vault/pull/117/files are an easy target for people to sneak code into aws-vault and steal secrets. 

Anyone can easily put up a PR saying they are updating vendor, github collapses the comments and its assumed the changes included are the output of running govendor, but in reality anything could be in there.

This adds a CI install step that fetches a clean vendor directly from the respective github repos. Of course each dep could also contain malicious code, but at least its a little harder.

For some reason travis isn't updating PR's on success/failure any more, but it is working:
image


joho:
We still need a longer term answer for CI & github integration but I'm inclined to merge this in the meantime as it's still a step forward on security.

Any objections?

joho:
No objections.

PR #127, opened 2017-06-07
Adds some better error messages

lox:
Primarily:

```bash
aws-vault login llamas
aws-vault: error: No credentials found for profile llamas.
Use 'aws-vault add llamas' to set up credentials or 'aws-vault list' to see what credentials exist
```

/cc @toolmantim


toolmantim:
Muchos better! 👍

PR #128, opened 2017-07-03
Add configurable ENV support for assume-role-ttl

davidjbarton:
I used aws-vault heavily with Terraform. Some operations (especially RDS-related) can take in excess of 15m to execute. It would be useful to be able to set this as an environment variable rather than remember to pass it as a flag for each run.

### DIFF
--- a/main.go
+++ b/main.go
@@ -97,6 +97,7 @@ func configureExecCommand(app *kingpin.Application, g *globalFlags) {
 
 	cmd.Flag("assume-role-ttl", "Expiration time for aws assumed role").
 		Default("15m").
+		OverrideDefaultFromEnvar("AWS_ASSUME_ROLE_TTL").
 		DurationVar(&input.RoleDuration)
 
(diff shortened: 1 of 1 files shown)


joho:
Seems reasonable to me. Could you also add something to the readme about it. Only way a user could discover this is configurable is by reading the source (something I noticed is also true about `AWS_FEDERATION_TOKEN_TTL`)

lox:
Does that show up in the help text?

joho:
I just checked `aws-vault exec --help` and `aws-vault exec --help-long` and there's no mention of ENV variables. Maybe I'll just merge this and we can address documenting env overrides separately.

lox:
The thing is all options have inferred env :) You can use AWS_VAULT_xxxx
for any argument. It's just incredibly unobvious.

joho:
Sounds _great_. Merging.

Don't think I'll schedule a point release for this yet, let a few other things build up first.

PR #130, opened 2017-08-05
Replace keychain with go keychain

lox:
I've been meaning to do this for ages, but finally got around to putting in the PR's to https://github.com/keybase/go-keychain that will let us entirely replace our long-ago-forked custom keychain interaction code. So much joy in this diff.

https://github.com/keybase/go-keychain/pull/12
https://github.com/keybase/go-keychain/pull/13

### DIFF
--- a/keyring/keychain.go
+++ b/keyring/keychain.go
@@ -69,20 +69,20 @@ func (k *keychain) Set(item Item) error {
 		return err
 	}
 
-	newItem := gokeychain.NewItem()
-	newItem.SetSecClass(gokeychain.SecClassGenericPassword)
-	newItem.SetService(k.service)
-	newItem.SetAccount(item.Key)
-	newItem.SetLabel(item.Label)
-	newItem.SetDescription(item.Description)
-	newItem.SetData(item.Data)
-	newItem.SetSynchronizable(gokeychain.SynchronizableNo)
-	newItem.SetAccessible(gokeychain.AccessibleWhenUnlocked)
-	newItem.UseKeychain(kc)
-	newItem.SetAccess(gokeychain.NoApplicationsTrusted)
+	kcItem := gokeychain.NewItem()
+	kcItem.SetSecClass(gokeychain.SecClassGenericPassword)
+	kcItem.SetService(k.service)
+	kcItem.SetAccount(item.Key)
+	kcItem.SetLabel(item.Label)
+	kcItem.SetDescription(item.Description)
+	kcItem.SetData(item.Data)
+	kcItem.SetSynchronizable(gokeychain.SynchronizableNo)
+	kcItem.SetAccessible(gokeychain.AccessibleWhenUnlocked)
+	kcItem.UseKeychain(kc)
+	kcItem.SetAccess(gokeychain.NoApplicationsTrusted)
 
 	log.Printf("Adding service=%q, account=%q to osx keychain %s", k.service, item.Key, k.path)
-	return gokeychain.AddItem(item)
+	return gokeychain.AddItem(kcItem)
 }
 
 func (k *keychain) Remove(key string) error {
--- a/keyring/keychain.go
+++ b/keyring/keychain.go
@@ -140,7 +140,7 @@ func (k *keychain) createOrOpen() (gokeychain.Keychain, error) {
 		return gokeychain.Keychain{}, err
 	}
 
-	if k.passphrase != "" {
+	if k.passphrase == "" {
 		log.Printf("Creating keychain %s with prompt", k.path)
 		return gokeychain.NewKeychainWithPrompt(k.path)
 	}
--- a/keyring/keychain.go
+++ b/keyring/keychain.go
@@ -2,27 +2,15 @@
 
 package keyring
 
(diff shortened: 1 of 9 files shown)


keyring/keychain.go
```diff
@@ -2,27 +2,15 @@
...
-	"path/filepath"
-	"unicode/utf8"
-	"unsafe"
 
-	"github.com/mitchellh/go-homedir"
+	gokeychain "github.com/keybase/go-keychain"
+)
+
+const (
+	keychainAccessGroup = "ACE1234DEF.com.99designs.aws-vault"
```
pda: I don't fully understand what this is, but looks like it maps to [`kSecAttrAccessGroup`](https://developer.apple.com/documentation/security/ksecattraccessgroup) which doesn't sound necessary in this use-case, and `ACE1234DEF` looks suspiciously like sample data.

> Access groups are used to share keychain items among two or more apps.

It's being passed to `gokeychain.NewGenericPassword()` which calls [`SetAccessGroup()`](https://github.com/keybase/go-keychain/blob/69c518e5bfff89ad62db81a185634d3aeac3c453/keychain.go#L226) which calls [`k.SetString(AccessGroupKey, ag)`](https://github.com/keybase/go-keychain/blob/69c518e5bfff89ad62db81a185634d3aeac3c453/keychain.go#L188) which treats empty string as unset. So perhaps we should just pass an empty string to `gokeychain.NewGenericPassword()`? Or is this actually doing something?

lox: Yup, agreed. It's for sharing items between a family of apps, which we don't need.

keyring/keychain.go
```diff
@@ -37,483 +20,131 @@ func init() {
 			name = "login"
 		}
 
-		home, err := homedir.Dir()
-		if err != nil {
-			return nil, err
-		}
-
 		return &keychain{
-			path:    filepath.Join(home, "/Library/Keychains/"+name+".keychain"),
```
lox: This is one change that probably needs testing. The new code uses simply name +".keychain", vs the old one that references the full path. It works fine on my machine which is 10.12.6.

keyring/keychain.go
```diff
@@ -37,483 +20,131 @@ func init() {
...
-	queryDict := mapToCFDictionary(query)
-	defer C.CFRelease(C.CFTypeRef(queryDict))
+	newItem := gokeychain.NewItem()
+	newItem.SetSecClass(gokeychain.SecClassGenericPassword)
+	newItem.SetService(k.service)
+	newItem.SetAccount(item.Key)
+	newItem.SetLabel(item.Label)
+	newItem.SetDescription(item.Description)
+	newItem.SetData(item.Data)
+	newItem.SetSynchronizable(gokeychain.SynchronizableNo)
```
lox: I like that this is more specific. I wonder what the old default was.

keyring/keychain.go
```diff
@@ -37,483 +20,131 @@ func init() {
...
-
-	if err == errDuplicateItem {
-		if err = k.Remove(item.Key); err != nil {
-			return err
-		}
-		err = newKeychainError(C.SecItemAdd(queryDict, nil))
-	}
-
-	return err
+	return gokeychain.AddItem(item)
```
mtibben: I'm getting a compile error?
```
$ go build github.com/99designs/aws-vault
# github.com/99designs/aws-vault/keyring
keyring/keychain.go:85: cannot use item (type Item) as type keychain.Item in argument to keychain.AddItem
```

mtibben: `s/item/newItem/` works

mtibben:
All looks good to me

lox:
We really need some sort of CI for this project.

mtibben:
Actually, I have spotted a difference in behaviour. When creating a keychain, v3.7.1 prompts for a password. This version does not. Is that intentional?

lox:
Nope, not intentional.

lox:
Good spotting.

keyring/keychain.go
```diff
@@ -37,483 +20,131 @@ func init() {
...
-	}
-	return C.CFArrayCreate(nil, valuesPointer, C.CFIndex(numValues), &C.kCFTypeArrayCallBacks)
-}
 
-func _CFArrayToArray(cfArray C.CFArrayRef) (a []C.CFTypeRef) {
-	count := C.CFArrayGetCount(cfArray)
-	if count > 0 {
-		a = make([]C.CFTypeRef, count)
-		C.CFArrayGetValues(cfArray, C.CFRange{0, count}, (*unsafe.Pointer)(&a[0]))
+	if k.passphrase != "" {
```
lox: 🤦‍♂️

lox:
Verified this works.

joho:
Hey, I was just going through documenting/tidying stuff in the keyring extraction I did.

Looks like https://github.com/keybase/go-keychain/pull/13 never merged. I'm wondering if there's a bug hidden somewhere and it's no longer doing what we think it does.

lox:
Nah, those folks just take a while to respond sometimes.

PR #131, opened 2017-08-05
WIP: Support macos touchid

lox:
This is my first pass at touchid support. Passing in the `--biometrics` flag will use touchid on compatible macOS hosts. The first time you do this we prompt for the master passphrase and then store it in the login keychain, along with the rights to access that item in future. Subsequent access is blocked by touch id, but after that we lookup the keychain password and use it to unlock the aws-vault keychain. 

I don't love the approach, but it closely matches 1password. Need to implement the extra security bits that they [do too](https://support.1password.com/touch-id-security-mac/):

 * [ ] When your fingerprint isn’t recognized three times in a row
 * [ ] Encrypt the stored master password with locally stored random key to ensure that it only works on your current host. Alternately with hardware id?

Also depends on #130.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: go
 
+osx_image: xcode8.3
 install: ./scripts/ci_install.sh
 
 go:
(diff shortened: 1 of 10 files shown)


lox:
This needs to be upstreamed to `999designs/keyring` I think.

scottrigby:
@lox I know this was a while ago, but any further thoughts on / interest in this? I looked at https://github.com/99designs/keyring but there doesn't seem to be any issue there yet. Could probably pitch in a little work on this but would want to connect about approach 😄 

Not critical, mainly an annoyance from what seems similar to https://github.com/99designs/aws-vault/issues/219

scottrigby:
Also see #273

mtibben:
This has bit-rotted

PR #133, opened 2017-08-07
Use the original credentials for key rotation, not a session

lox:
This should address @pda's problem in#132.

PR #134, opened 2017-08-07
Support profiles in rotate

lox:
Based on #132 and #133, key rotation was broken in a number of ways:

* It used a session, which means IAM calls are only possible if there is an MFA
* It didn't support profiles, so if your master creds didn't have permission to do IAM stuff and you needed to assume a role to do so it wouldn't work
* It dumped your creds to the console on error

This fixes all of the above as well as reducing the amount of prompts you get from Keychain. 

Damn you for nerd sniping me just before I fell asleep @pda.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -33,6 +33,7 @@ type VaultOptions struct {
 	MfaPrompt          prompt.PromptFunc
 	NoSession          bool
 	Profiles           profiles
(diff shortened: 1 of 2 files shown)


lox:
Beyond that, that method needs to be broken into a few, but damned if I could figure out a neat way to do it.

PR #138, opened 2017-08-09
Fixed regression that prompted for keychain access for sessions

lox:
I noticed that we were back to displaying three prompts for every access, even when there is a session. Should only prompt once now and not at all if there is an active session.

### DIFF
--- a/config.go
+++ b/config.go
@@ -83,5 +83,5 @@ func formatCredentialError(p string, from profiles, err error) string {
 			sourceDescr, source)
 	}
 
-	return fmt.Sprintf("Failed to get credentials for %s: %v", err, sourceDescr)
+	return fmt.Sprintf("Failed to get credentials for %s: %v", sourceDescr, err)
 }
--- a/keyring/keychain.go
+++ b/keyring/keychain.go
@@ -79,9 +79,12 @@ func (k *keychain) Set(item Item) error {
 	kcItem.SetSynchronizable(gokeychain.SynchronizableNo)
 	kcItem.SetAccessible(gokeychain.AccessibleWhenUnlocked)
 	kcItem.UseKeychain(kc)
-	kcItem.SetAccess(gokeychain.NoApplicationsTrusted)
 
-	log.Printf("Adding service=%q, account=%q to osx keychain %s", k.service, item.Key, k.path)
+	if !item.TrustSelf {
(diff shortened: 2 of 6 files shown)


pda:
Any idea which commit/PR the regression came in?
Either I haven't experienced it, or I've grown used to it; on `v3.7.1-11-g6dea313`

lox:
I actually think it's been there forever. Try clearing your sessions and running:

```
aws-vault exec my-profile-with-mfa -- time
aws-vault exec my-profile-with-mfa -- time
```

You should get exactly 1 Keychain prompt the first time, and one MFA prompt the second time.

pda:
With v3.7.1-11-g6dea313:

* First command
    * keychain password prompt
    * keychain “allow” button prompt (I never click “always allow”)
    * MFA prompt
* Second command
    * keychain “allow” button prompt

With `v3.7.1-36-g6fcd647` (current master):

* First command
    * keychain “allow prompt”
    * MFA prompt
* Second command
    * *(nothing)*

I guess I had just grown accustomed to all the extra button prompts, and hitting spacebar to “allow”. Nice improvement (as long as it's not harming the security model…?)

lox:
My view is that it's basically the design of sessions. I'd always intended subsequent calls to a session to work without a prompt.

pda:
In my mind, sessions primarily avoid repeated MFA, and keychain unlock/auto-lock time avoid repeated password entry, and both are worthwhile security/usability trade-offs.

Within a session, clicking (or pressing spacebar) to allow each `aws-vault exec …` command is a price I'm happy to pay to know that background processes (e.g. malicious npm modules) can't invisibly obtain AWS credentials.

I'd lean towards keeping that behaviour. And I guess people can and will click Always Allow if they want to make that further security/usability trade-off.

I wonder how this question maps to non-macOS backends.

lox:
I shouldn't have done this, but this PR also included some fixes for the go-keychain stuff: https://github.com/99designs/aws-vault/pull/138/commits/4868b434f590a42a321da7a3e657be651a101190

And it also dropped support for reading credentials from env and shared credentials files, which merited it's own PR too https://github.com/99designs/aws-vault/pull/138/commits/d84f6636db253a52f9b273170b103538e782da32

I regret this now.

lox:
#140 is part of this.

lox:
As is #141.

lox:
And #142. I feel much better now.

PR #141, opened 2017-08-10
Refactor order in which access to keychain is requested

lox:
When a session is requested from the keychain, the master credentials shouldn't be requested. Instead we should be prompting for access to the session keychain item.

The net effect is the same, requesting master credentials pops a prompt, and then the next time you request access to a session it pops a prompt, but it's the correct prompt, for the correct item, the session. 

I guess the downside of this, and it's probably a sizeable downside is that you can't hit "Always trust" the second time and never see a prompt again. You might be able to do that for the master credentials, but you would need to do it for every new session key that is stored.

Bug, feature, regression, improvement, not sure. Maybe all?

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -83,5 +83,5 @@ func FormatCredentialError(profileKey string, from Profiles, err error) string {
 			sourceDescr, source)
 	}
 
-	return fmt.Sprintf("Failed to get credentials for %s: %v", err, sourceDescr)
+	return fmt.Sprintf("Failed to get credentials for %s: %v", sourceDescr, err)
 }
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -151,6 +151,48 @@ func (p *VaultProvider) Retrieve() (credentials.Value, error) {
 	return value, nil
 }
 
+// RetrieveWithoutSessionToken returns credentials that are either the master credentials or
+// a session created with AssumeRole. This allows for usecases where a token created with AssumeRole
+// wouldn't work.
+func (p *VaultProvider) RetrieveWithoutSessionToken() (credentials.Value, error) {
+	log.Println("Skipping session token and using master credentials directly")
+
+	creds, err := p.getMasterCreds()
+	if err != nil {
+		return credentials.Value{}, err
+	}
+
+	if role, ok := p.profiles[p.profile]["role_arn"]; ok {
+		session, err := p.assumeRole(creds, role)
+		if err != nil {
+			return credentials.Value{}, err
+		}
+
+		log.Printf("Using role ****************%s, expires in %s",
+			(*session.AccessKeyId)[len(*session.AccessKeyId)-4:],
+			session.Expiration.Sub(time.Now()).String())
+
+		window := p.ExpiryWindow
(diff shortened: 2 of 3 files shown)


provider.go
```diff
@@ -86,42 +86,12 @@ func NewVaultProvider(k keyring.Keyring, profile string, opts VaultOptions) (*Va
 	}, nil
 }
 
+// Retrieve returns credentials protected by a GetSessionToken. If there is an associated
+// role in the profile then AssumeRole is applied. The benefit of a session is that it doesn't
+// require MFA or a user prompt to access the keychain item, much like sudo.
 func (p *VaultProvider) Retrieve() (credentials.Value, error) {
-	creds, err := p.getMasterCreds()
```
lox: Previously I was accessing these up front, regardless of whether it was a `NoSession` call, or whether there was a session lookup later.

lox: I'm not entirely sure if I meant to do that, but it seems the wrong way around.

sessions.go
```diff
@@ -64,10 +64,11 @@ func (s *KeyringSessions) Store(profile string, session sts.Credentials, duratio
 
 	log.Printf("Writing session for %s to keyring", profile)
 	s.Keyring.Set(keyring.Item{
-		Key:       s.key(profile, duration),
-		Label:     "aws-vault session for " + profile,
-		Data:      bytes,
-		TrustSelf: true,
+		Key:         s.key(profile, duration),
+		Label:       "aws-vault session for " + profile,
+		Description: "aws-vault session for " + profile,
```
lox: Description doesn't serve any purpose here.

sessions.go
```diff
@@ -64,10 +64,11 @@ func (s *KeyringSessions) Store(profile string, session sts.Credentials, duratio
...
+		Data:        bytes,
+		TrustSelf:   false,
```
lox: This is the meaningful change WRT to sessions.

config.go
```diff
@@ -83,5 +83,5 @@ func formatCredentialError(p string, from profiles, err error) string {
 			sourceDescr, source)
 	}
 
-	return fmt.Sprintf("Failed to get credentials for %s: %v", err, sourceDescr)
```
lox: Ooops!

provider.go
```diff
@@ -169,6 +150,48 @@ func (p *VaultProvider) Retrieve() (credentials.Value, error) {
 	return value, nil
 }
 
+// RetrieveWithoutSessionToken returns credentials that are either the master credentials or
+// a session created with AssumeRole. This allows for usecases where a token created with AssumeRole
+// wouldn't work.
```
pda: Should this conditional be lifted up into the caller? It feels dangerous having a conditional based on some global state which defaults to exposing master credentials if something goes wrong looking for the role. Two clearly named functions might be nicer;

```go
if role, ok := p.profiles[p.profile]["role_arn"]; ok {
  credentials, err := GetAssumedRoleCredentials(role)
} else {
  credentials, err := GetMasterCredentials()
}
```

lox: Unfortunately `Retrieve` is the interface method here.

PR #143, opened 2017-08-10
Allow the keyring name to be customized

lox:
This allows you to use the login keychain on macOS.

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,10 +12,6 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
-const (
-	KeyringName = "aws-vault"
-)
-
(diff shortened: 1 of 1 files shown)


pda:
If the idea is to evaluate using the login keychain as the default/only supported keychain, this is a nice transitional option. But maintaining it forever might just add confusion/optionitis and another factor in the many ways aws-vault can be working for different people. Perhaps it could be simplified to `--use-login-keychain` so people can't invent too many new ways of using it? Easier to turn that flag into a noop later if we decide to only use the login keychain or go back to a dedicated one.

lox:
Superseded by #144

Issue #49, opened 2015-10-16
No credentials in profile after restart


coen-hyde:
It appears my access credentials got lost from the key chain after a restart. My profile appears in aws-vault but I get the message "No credentials found for profile "development"" when trying to use it. Any idea what has happened?

lox:
Hrm, that's very odd. What do you get for `aws-vault --debug ls`?

coen-hyde:
weird indeed. It's working now and i didn't do anything. Still have my bash history from when it wasn't working. Here is the output of ```aws-vault --debug ls```

```
DEBUG 2015/10/16 15:50:39 Opening keychain /Users/coen-hyde/Library/Keychains/aws-vault.keychain
development
default session
development session
```

I also verified the existence of the aws-vault keychain file when it wasn't working.

coen-hyde:
hmm, it's doing it again

```
$ aws-vault exec development --write-env -- env | grep AWS
No credentials found for profile "development"
```

```
$ aws-vault --debug ls
DEBUG 2015/10/19 16:26:41 Opening keychain /Users/coen-hyde/Library/Keychains/aws-vault.keychain
development
default session
development session
```

pda:
Are you accidentally landing in an aws-vault subshell by running `aws-vault exec <profile>` without a command?

```sh
$ ./aws-vault-2.4.1 exec fooprofile -- env | grep AWS
AWS_CONFIG_FILE=/var/folders/ck/tx_xkpf54x30fbh9__ycyn340000gn/T/aws-vault064985698
AWS_DEFAULT_PROFILE=fooprofile
AWS_PROFILE=fooprofile
AWS_DEFAULT_REGION=us-east-1
AWS_REGION=us-east-1

$ ./aws-vault-2.4.1 exec fooprofile    # seems to do nothing, actually launches subshell

$ ./aws-vault-2.4.1 exec fooprofile -- env | grep AWS
No credentials found for profile "fooprofile"

$ exit

$ ./aws-vault-2.4.1 exec fooprofile -- env | grep AWS
AWS_CONFIG_FILE=/var/folders/ck/tx_xkpg54x30dbh9__ycyn320000dn/T/aws-vault115037450
AWS_DEFAULT_PROFILE=fooprofile
AWS_PROFILE=fooprofile
AWS_DEFAULT_REGION=us-east-1
AWS_REGION=us-east-1
```

[...]

PR #144, opened 2017-08-10
Support using the login keychain

lox:
This adds `--use-login-keychain` which will use your default `login` keychain on macOS rather than creating an `aws-vault` keychain.

### DIFF
--- a/cli/global.go
+++ b/cli/global.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	KeyringName = "aws-vault"
+	DefaultKeyringName = "aws-vault"
 )
 
 var (
@@ -23,9 +23,10 @@ var (
 )
 
 var GlobalFlags struct {
-	Debug        bool
-	Backend      string
-	PromptDriver string
+	Debug            bool
+	Backend          string
+	PromptDriver     string
+	UseLoginKeychain bool
 }
 
 func ConfigureGlobals(app *kingpin.Application) {
@@ -42,17 +43,24 @@ func ConfigureGlobals(app *kingpin.Application) {
 		OverrideDefaultFromEnvar("AWS_VAULT_PROMPT").
 		EnumVar(&GlobalFlags.PromptDriver, promptsAvailable...)
 
+	app.Flag("use-login-keychain", "Uses your default login keychain rather a new one").
+		OverrideDefaultFromEnvar("AWS_VAULT_USE_LOGIN_KEYCHAIN").
+		BoolVar(&GlobalFlags.UseLoginKeychain)
+
 	app.PreAction(func(c *kingpin.ParseContext) (err error) {
+		var keyringName = DefaultKeyringName
+		if GlobalFlags.UseLoginKeychain && GlobalFlags.Backend == keyring.KeychainBackend {
+			keyringName = "login"
+		}
 		if !GlobalFlags.Debug {
 			log.SetOutput(ioutil.Discard)
 		}
 		if keyringImpl == nil {
-			keyringImpl, err = keyring.Open(KeyringName, GlobalFlags.Backend)
+			keyringImpl, err = keyring.Open(keyringName, GlobalFlags.Backend)
 		}
 		if awsConfigFile == nil {
 			awsConfigFile, err = vault.NewConfigFromEnv()
 		}
 		return err
 	})
-
 }


mtibben:
I think I'm missing some context. What's the end goal, are we evaluating whether to use the login keychain as the default? And if so what's the migration path, any security tradeoffs?

lox:
I guess let's talk about the opposite, what are the goals of a separate keychain? As I've spent more time with aws-vault, it just seems like a reasonable trade-off for UX to allow people to store (and eventually synchronise to iCloud) credentials in their primary login keychain. I see most of the benefit in aws-vault being the prompt for credential use. The seperate keychain offers much shorter timeouts, which in theory protects you if you leave you computer unlocked and unattended for more than 5 minutes, but it seems like a pretty marginal benefit. 

In the future, I'd like aws-vault to use the `login` keychain as the default as this is what most apps do and it's expected behaviour for most people. I'd rather push people to rotate credentials and use MFA liberally.

lox:
Interested in @sj26 and @pda's take on this too.

cli/global.go
```diff
@@ -42,17 +43,24 @@ func ConfigureGlobals(app *kingpin.Application) {
...
 		EnumVar(&GlobalFlags.PromptDriver, promptsAvailable...)
 
+	app.Flag("use-login-keychain", "Uses your default login keychain rather a new one").
+		OverrideDefaultFromEnvar("AWS_VAULT_USE_LOGIN_KEYCHAIN").
+		BoolVar(&GlobalFlags.UseLoginKeychain)
+
 	app.PreAction(func(c *kingpin.ParseContext) (err error) {
+		var keyringName = DefaultKeyringName
+		if GlobalFlags.UseLoginKeychain && GlobalFlags.Backend == keyring.KeychainBackend {
+			keyringName = "login"
```
mtibben: Is `login` what we want here? Seems like it's possible for the "default" keychain to be named differently

mtibben:
So to be clear the ACLs only allowing access to `aws-vault` are still set on the individual item right? This doesn't mean any old process can access the credentials

lox:
Yup, correct. It's just which keychain it's located in.

cli/global.go
lox: Yeah, fair point. I will investigate that. Perhaps "default" keychain is the phrasing we should be using.

mtibben:
That makes sense to me. Although I have found it useful in the past to back up / delete the aws-vault keychain directly when debugging aws-vault. I think I would want to retain the ability to specify the keychain in an env var as you had in #143 😬   Or do you have any other suggestions for that use case? I'm not convinced having `AWS_VAULT_KEYRING_NAME ` adds much complexity. Thoughts @pda ?

lox:
Part of this is removing some of the complexity too. I'd like it to just use the default keychain eventually which should eliminate https://github.com/99designs/aws-vault/issues/49

lox:
Perhaps an export or import facility would be better for that @mtibben?

lox:
Could read/write to `~/.aws/credentials`.

lox:
Deleting entries in the default keychain is still pretty easy.

cli/global.go
sj26: Looks a bit like you can use 'NULL' to mean 'default keychain':

https://developer.apple.com/documentation/security/1393225-seckeychainitemcreatefromcontent#topic-content

sj26:
I'm on-board with default keychain :+1:

I kinda already have a patched version of aws-vault which changes KeychainName to login >.>

heph:
I would love to see this as the default behavior to make the typical user experience as straightforward as possible, while keeping a flag or env var to override for people who require the flexibility.

I'm supporting 100+ devs who need to securely manage their AWS credentials with a minimum of fuss. Keeping credentials in a non-login keychain has and will result in them forgetting the password when "Always Allow" expires (as we recently found 356 days after first onboarding users with gpg keys). The login keychain is a perfect default as it's unlockable with the user's account password so cannot be forgotten.

lox:
That makes a lot of sense @heph. We'll try and get this in the next major version. Thanks for the feedback!

lox:
I'll progress this after #181, otherwise I don't have a good way of moving creds between keychains.

lox:
Ended up deciding the import/export was a bad idea. Moving this to #197.

PR #146, opened 2017-08-14
Extract keyring into a separate package

joho:
@lox needed the keyring logic for something outside the aws-vault project so we've split it out to a new package at https://github.com/99designs/keyring

This PR switches aws-vault over to the now vendored version of the keyring logic.


joho:
Build passes - shipit?

lox:
Thanks! Awesome.

PR #148, opened 2017-08-26
Move go files into top level packages

lox:
I got sick of never being able to find anything, so I split the massive main package into some sub-packages:

* cli - All the cli commands
* vault - The guts of the credentials stuff
* server - The EC2 metadata server implementation

I made no functional changes here, things just got shuffled around.

Yes, this is a big change. Whatevs. I also coded it in bed. Deal with it. 🕶


lox:
Yeah, I've tried a few ways to deal with that @joho in the past, I can never find a layer of abstraction that isn't more work than it's worth. I kind of like that the tests can now declare just the commands under test though.

Issue #114, opened 2017-02-12
panic on login


mtibben:
Using aws-vault 3.7.0
```
$ aws-vault login home
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x4006cd2]

goroutine 1 [running]:
panic(0x447f2a0, 0xc4200120d0)
	/usr/local/Cellar/go/1.7.5/libexec/src/runtime/panic.go:500 +0x1a1
main.getUserName(0xc420131520, 0x14, 0xc420136690, 0x28, 0x0, 0x0, 0x0, 0x0, 0xc420116a00, 0x47, ...)
	/Users/michael/src/github.com/99designs/aws-vault/login.go:204 +0x202
main.getFederationToken(0xc420131520, 0x14, 0xc420136690, 0x28, 0x0, 0x0, 0x0, 0x0, 0x274a48a78000, 0x0, ...)
	/Users/michael/src/github.com/99designs/aws-vault/login.go:182 +0x364
main.LoginCommand(0xc4200b8870, 0x7fff5fbff4c0, 0x4, 0x476c920, 0xc420113860, 0x0, 0x0, 0x45465b0, 0x0, 0x274a48a78000, ...)
	/Users/michael/src/github.com/99designs/aws-vault/login.go:76 +0x1967
main.configureLoginCommand.func1(0xc42009ce10, 0x402349a, 0x448c7a0)
	/Users/michael/src/github.com/99designs/aws-vault/main.go:181 +0xb5
github.com/99designs/aws-vault/vendor/gopkg.in/alecthomas/kingpin%2ev2.(*actionMixin).applyActions(0xc4200b6498, 0xc42009ce10, 0x0, 0x0)
	/Users/michael/src/github.com/99designs/aws-vault/vendor/gopkg.in/alecthomas/kingpin.v2/actions.go:28 +0x58
github.com/99designs/aws-vault/vendor/gopkg.in/alecthomas/kingpin%2ev2.(*Application).applyActions(0xc4200b8870, 0xc42009ce10, 0x0, 0x0)
	/Users/michael/src/github.com/99designs/aws-vault/vendor/gopkg.in/alecthomas/kingpin.v2/app.go:554 +0x126
[...]

PR #152, opened 2017-09-17
Fallback to Arn when Username doesn't exist

mtibben:
It seems `User.UserName` isn't always present in a [`GetUser`](http://docs.aws.amazon.com/IAM/latest/APIReference/API_GetUser.html) response. I experienced this while using aws-vault with root credentials

So this now falls back to parsing the `User.Arn`.

Fixes https://github.com/99designs/aws-vault/issues/114

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/99designs/aws-vault/prompt"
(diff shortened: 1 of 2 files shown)


joho:
LGTM.

The gitignore change required?

mtibben:
Huh I just noticed https://github.com/99designs/aws-vault/commit/74160d01f6c10b342472b2311945e90296fa2b69, should probably generalise that.. Is root the only circumstance where `Username` won't be present @lox ?

lox:
Your approach is better. Should have PR'd.

Issue #160, opened 2017-10-11
Which profiles does aws-vault manage?


wgj:
Possible duplicate of #137 

aws-vault v3.7.1
macOS 10.12.6

I have a lot of profiles (roles, actually) that I switch between. I tried to use one with aws-vault that I obviously forgot to add. In order to solve this problem, I wanted to see all the profiles I've already added to aws-vault, so that I know which profiles I still need to add.

However, when I run `aws-vault --list`, it doesn't seem to do what the help text describes.
Similar to #137, I would have expected to see a list of profiles that have been added (i.e., `aws-vault add foo`).
```
~ $ aws-vault exec foo -- /usr/local/bin/aws s3 ls
aws-vault: error: No credentials found for profile "foo"

~ $ aws-vault --help ls
usage: aws-vault list

List profiles
...

~ $ aws-vault ls
WestonJ0
WestonJ0 session (12...34)
WestonJ0 session (56...78)
WestonJ0 session (87...65)
WestonJ0 session (43...21)
~ $ 
```

wgj:
Disregard.

After investigating more I found some pretty obvious flaws in my logic (What's the access key for a role? ;), and after finding that there were some roles which I can use with `aws-vault` that weren't _added_ (again, an impossible action), I found that the issue was a typo in the profile's name within `.aws/config`.

Although the use case is shaky, I think it still stands in that `aws-vault ls` should include assumed roles. Closing in favor of #137.

PR #161, opened 2017-10-11
Clarify credentials, sessions and profiles

### LINKED ISSUE #137: What should `aws-vault list` show?
Currently `aws-vault list` will show all keys in the aws-vault keychain. This includes sessions. 

I've been thinking of breaking out: `aws-vault list-profiles`, `aws-vault list-sessions` and then have `aws-vault list` show only the credentials that are stored. The alternative would be to have `aws-vault list` have a `--profiles` and `--session` flag. 

Feels?

lox:
Closes #137. Relates to #160.

General cleanup of profile / config parsing code. Implements the `list` command with tabular output:

```
Profile                   Credentials              Sessions
=======                   ===========              ========
ljd                       ljd                      61343032633337323134, 32663735326133373631, 61633665646639303539
ljd-admin                 ljd                      61343032633337323134, 32663735326133373631, 61633665646639303539
xyz                       ljd                      61343032633337323134, 32663735326133373631, 61633665646639303539
```

This of course begs the question, why do we have so many old sessions hanging around?


mtibben:
So if I understand correctly - Profile is what's in config, Credentials is the ID for the credentials aws-vault uses to generate the session, and will correspond to the `profile` name or the `source_profile` if that's being used

It seems the challenge is working out the right UI to disambiguate between profiles/credentials and make users aware of this nuance, so this is definitely an improvement on what we currently have.

So because it seems this command is using profiles as it's main starting point, does this command list all profiles in config, or just profiles we have credentials stored for?

mtibben:
Ha I just realised you asked the same question https://github.com/99designs/aws-vault/issues/137#issuecomment-331605356

lox:
Yeah, I decided c) all of the above. It will show profiles in your config and it will show credentials without a profile in the config. The only thing it won't do is show sessions that don't have a config entry.

lox:
I'm going to automatically add a config entry with the `add` command to keep things consistent.

Issue #164, opened 2017-10-17
mfa_serial isn't inherited from source_profile?


dnch:
Took me a while to figure this one out. 

We've got an AWS organisation with several accounts, with access to each one determined by an individual policy. Additionally, all our IAM users have a policy that has an explicit deny on nearly everything if MFA isn't present. 

My AWS config looks like this:

```
[profile company]
region = ap-southeast-2
mfa_serial = arn:aws:iam::123456789012:mfa/user.name
output = json

[profile build]
role_arn = arn:aws:iam::987654321098:role/OrganizationAccountAccessRole
source_profile = company
```

I was unable to perform any actions involving the `build` profile, basically getting the "user.name not authorized to perform: sts:AssumeRole on resource $ROLE` error.

It wasn't until I added the `mfa_serial` line to each profile that it worked as expected.

Is it possible to have profiles "inherit" mfa_serial from the designated source profile, or is this a limitation of the AWS SDK?

lox:
Yeah, that always confused me too. Source_profile only inherits credentials, not config. We could easily change that, but it would confuse people that expected the standard AWS cli behaviour.

FernandoMiguel:
yeah, I've always expected the MFA to not be in every profile, but seems it's required
this is my skeleton config
https://github.com/FernandoMiguel/AWS-Trust-CrossAccounts/blob/master/dot_aws/config

FYI, this is the UPDATED aws iam policy for users enforcing MFA
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_delegate-permissions_examples.html#creds-policies-mfa-console
it's updated cause they forgot that the attacker could remove the MFA

lox:
The reason it's needed is that STS tokens can't make IAM calls unless they have an MFA associated.

lox:
Thanks for the docs update @dnch, going to close this one. Feel free to re-open if needed.

PR #165, opened 2017-10-17
add note about mfa_serial / assuming roles

dnch:
in response to the discussion in #164


joho:
Many thanks. One of those little quality of life docs tweaks that makes all the difference!

lox:
Thanks @dnch! 😍

PR #168, opened 2017-10-19
Use app.Fatal instead of app.Error

### LINKED ISSUE #167: aws-vault exec exits 0 when trying to run nonexistent command
If I run `aws-vault exec dev -- nonexistentcommand`, I get a 0 return code. Same if cmd exists but is not executable. Investigating now.

I would expect it to return a nonzero code.

kevinburke:
app.Error merely logs, instead of quitting the program with a non-zero
return code. This means some subset of programs fail to exec the
underlying binary and then exit 0, which might incorrectly indicate to
a caller (like me) that they exited successfully.

Fixes #167.


lox:
Thanks!

PR #170, opened 2017-10-23
IAM takes a while to propagate new keys

### LINKED ISSUE #132: Key rotation: documentation and/or bug
I'm attempting to `aws-vault rotate <profile>` where profile is an IAM user with AdministratorAccess policy attached, and I'm seeing something like https://github.com/99designs/aws-vault/issues/15:

```
$ aws-vault rotate pda
Enter passphrase to unlock /home/pda/.awsvault/keys:
Using old credentials to create a new access key
aws-vault: error: InvalidClientTokenId: The security token included in the request is invalid
        status code: 403, request id: ...
```

Using a `v3.7.1-23-g8c008d3` build of aws-vault.

Same error for any `aws iam ...` operation, but all other AWS services are working;

```
$ aws-vault exec pda -- aws iam list-access-keys
Enter passphrase to unlock /home/pda/.awsvault/keys:
An error occurred (InvalidClientTokenId) when calling the ListAccessKeys operation: The security token included in the request is invalid
```

Is this expected? Is there an IAM setting/policy somewhere that needs to be changed? Or does rotation only work for non-IAM root account credentials? Should we be using the stored access key rather than a session token for the `rotate` command?

I've used `aws-vault rotate` before on work AWS accounts, but never with this personal AWS account.

As https://github.com/99designs/aws-vault/issues/129 pointed out, there's no mention of `aws-vault rotate` in the README etc; I'm happy to add that if I can figure out how to use it :)

lox:
This is why key rotation was failing, new IAM keys take a while to propagate. This adds a terrible sleep and then poll. 

Fixes #132.

### DIFF
--- a/cli/add.go
+++ b/cli/add.go
@@ -38,7 +38,7 @@ func ConfigureAddCommand(app *kingpin.Application) {
 func AddCommand(app *kingpin.Application, input AddCommandInput) {
 	var accessKeyId, secretKey string
 
-	if source, ok := awsConfig.SourceProfile(input.Profile); ok {
+	if source, _ := awsConfig.SourceProfile(input.Profile); source.Name != input.Profile {
 		app.Fatalf("Your profile has a source_profile of %s, adding credentials to %s won't have any effect",
 			source.Name, input.Profile)
 		return
--- a/cli/login.go
+++ b/cli/login.go
@@ -7,8 +7,6 @@ import (
 	"log"
 	"net/http"
 	"net/url"
-	"regexp"
-	"strings"
 	"time"
 
 	"github.com/99designs/aws-vault/prompt"
@@ -17,7 +15,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/skratchdot/open-golang/open"
 	"gopkg.in/alecthomas/kingpin.v2"
@@ -201,7 +198,7 @@ func getFederationToken(creds credentials.Value, d time.Duration) (*sts.Credenti
 	})
 	client := sts.New(sess)
 
-	currentUsername, err := getCurrentUserName(sess)
+	currentUsername, err := vault.GetUsernameFromSession(sess)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


cli/rotate.go
```diff
@@ -142,20 +150,26 @@ func RotateCommand(app *kingpin.Application, input RotateCommandInput) {
 		return
 	}
 
-	newVal, err := newSessionCreds.Get()
-	if err != nil {
-		app.Fatalf(awsConfig.FormatCredentialError(err, input.Profile))
-		return
-	}
+	log.Printf("Waiting for new IAM credentials to propagate")
+	time.Sleep(time.Second * 8)
```
mtibben: 8 seconds seems a while, should we just start polling immediately?

cli/login.go
```diff
@@ -219,11 +220,20 @@ func getFederationToken(creds credentials.Value, d time.Duration) (*sts.Credenti
...
+
 func getCurrentUserName(sess *session.Session) (string, error) {
 	client := iam.New(sess)
 
 	resp, err := client.GetUser(&iam.GetUserInput{})
 	if err != nil {
+		matches := getUserErrorRegexp.FindAllStringSubmatch(err.Error(), -1)
+		if len(matches) > 0 {
+			log.Printf("%#v", matches)
+		}
```
mtibben: What is this logging? does it need a prefix?

cli/rotate.go
lox: It seems pretty consistently to take about 8 seconds :)

lox: We could suppress errors and poll regularly though.

cli/login.go
lox: Good catch, was purely debugging. Feels on pattern matching this error? It seems epic-ly el grosso.

cli/add.go
```diff
@@ -38,7 +38,7 @@ func ConfigureAddCommand(app *kingpin.Application) {
 func AddCommand(app *kingpin.Application, input AddCommandInput) {
 	var accessKeyId, secretKey string
 
-	if source, ok := awsConfig.SourceProfile(input.Profile); ok {
+	if source, _ := awsConfig.SourceProfile(input.Profile); source.Name != input.Profile {
```
lox: This change is unrelated, and is just a bug I introduced in master.

cli/rotate.go
```diff
@@ -92,13 +94,19 @@ func RotateCommand(app *kingpin.Application, input RotateCommandInput) {
 		return
 	}
 
+	var iamUserName *string
+
+	// A username is needed for some IAM calls if the credentials have assumed a role
```
lox: Context here @mtibben is that for the root user this doesn't exist and needs to be null.

lox:
I've moved everything to `vault.Rotator` @mtibben and cleaned things up a bit.

mtibben:
lgtm

damonmaria:
Thanks for the fix 👍 Been missing my key rotation :)

Would love an update in homebrew when someone with privileges gets the chance.

Issue #166, opened 2017-10-18
Metadata Server on Windows


williamdenton:
My primary dev environment is windows, so I had a look to see what it would take to get the server running on windows.

Currently windows server is explicitly disabled as there is no code to install the network alias:

https://github.com/99designs/aws-vault/blob/eb8b1dcd07a88a16b1ec889976927122a798c583/server/alias_windows.go#L7-L9

if I update that to:

```go

func installNetworkAlias() ([]byte, error) {
	exec.Command("netsh", "interface", "ipv4", "add", "address", "Loopback Pseudo-Interface 1", "169.254.169.254", "255.255.0.0")
	return []byte{}, nil
}

```

installs the interface (when run from an elevated (administrator/sudo) prompt. So far so good. Its a hack as it doesn't respect the output of that command and im not 💯 on the interface name being portable to other machines also running it a second time will fail (but that doesn't matter)

Next problem is the internals of the server mode try to start the server in the background using `sudo`, this is no use on windows:

https://github.com/99designs/aws-vault/blob/eb8b1dcd07a88a16b1ec889976927122a798c583/server/server.go#L72-L81

If I comment out that block and then call `aws-vault server` from exec the server is started, albeit nested

```ps
aws-vault exec -s profile -- aws-vault server
```

It works!

But not in a way that I'd be comfortable creating a PR for. So I'm looking for feedback on how to approach this to get it running on windows in a way that can be merged. I've got lots of people that would love to use this in my team and my workaround https://github.com/williamdenton/awslocalmetadata that I was using before trying to get this working can be killed.

lox:
This looks awesome @williamdenton! If you wanted to put this into a PR, we'd be happy to accept it. Partial windows support is better than none for the metadata server!

PR #174, opened 2017-10-30
Metadata server running on windows

williamdenton:
Issue https://github.com/99designs/aws-vault/issues/166 describes whats going on here

This is a breaking change for the way the metadata server is started

```sh
aws-vault exec <environment> --server
```

is now

```sh
aws-vault exec <environment> -s -- aws-vault server
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -76,30 +76,42 @@ func checkServerRunning(bind string) bool {
 	return err == nil
 }
 
+func StartCredentialProxyOnWindows() error {
+	log.Printf("Starting `aws-vault server` in the background")
+	cmd := exec.Command(os.Args[0], "server")
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	time.Sleep(time.Second * 1)
+	if !checkServerRunning(metadataBind) {
+		return errors.New("The credential proxy server isn't running. Run aws-vault server as Administrator in the background and then try this command again")
+	}
+	return nil
+}
+
+func StartCredentialProxyWithSudo() error {
+	log.Printf("Starting `aws-vault server` as root in the background")
+	cmd := exec.Command("sudo", "-b", os.Args[0], "server")
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func StartCredentialProxy() error {
+	if runtime.GOOS == "windows" {
+		return StartCredentialProxyOnWindows()
+	}
+	return StartCredentialProxyWithSudo()
+}
+
 func StartCredentialsServer(creds *vault.VaultCredentials) error {
 	if !checkServerRunning(metadataBind) {
-		if runtime.GOOS == "windows" {
-			log.Printf("Starting `aws-vault server` in the background")
-			cmd := exec.Command(os.Args[0], "server")
-			cmd.Stdin = os.Stdin
-			cmd.Stdout = os.Stdout
-			cmd.Stderr = os.Stderr
-			if err := cmd.Start(); err != nil {
-				return err
-			}
-			time.Sleep(time.Second * 1)
-			if !checkServerRunning(metadataBind) {
-				return errors.New("The credential proxy server isn't running. Run aws-vault server as Administrator in the background and then try this command again")
-			}
-		} else {
-			log.Printf("Starting `aws-vault server` as root in the background")
-			cmd := exec.Command("sudo", "-b", os.Args[0], "server")
-			cmd.Stdin = os.Stdin
-			cmd.Stdout = os.Stdout
-			cmd.Stderr = os.Stderr
-			if err := cmd.Run(); err != nil {
-				return err
-			}
+		if err := StartCredentialProxy(); err != nil {
+			return err
 		}
 	}
 
--- a/server/server.go
+++ b/server/server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -85,6 +86,10 @@ func StartCredentialsServer(creds *vault.VaultCredentials) error {
 			cmd.Stderr = os.Stderr
 			if err := cmd.Start(); err != nil {
 				return err
+			}
+			time.Sleep(time.Second * 1)
+			if !checkServerRunning(metadataBind) {
+				return errors.New("The credential proxy server isn't running. Run aws-vault server as Administrator in the background and then try this command again")
 			}
 		} else {
 			log.Printf("Starting `aws-vault server` as root in the background")
--- a/server/server.go
+++ b/server/server.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"runtime"
 	"time"
 
 	"github.com/99designs/aws-vault/vault"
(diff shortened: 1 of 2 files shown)


williamdenton:
I'm an experienced developer but this is the first Go code I've ever written, so assume I'm a complete newbie.  I'm totally happy to have my approach/style critiqued

williamdenton:
One thing i didn't fully understand is why there are two servers running. One on 9099 which generates the security credentials payload and presumably token management and another than exposes it on the metadata ip address.

Could the two not be combined? If they could be combined that would solve the nested calls

lox:
I wish I'd documented it better, as I wrote it several years ago, but the two part design has some advantages. 

The part that fetches credentials and has access to the keyring is run in the foreground on localhost:9099 and only lasts as long as your interactive `aws-vault exec` session does, whilst the long-running root privileged portion runs in the background. 

It's a stronger design from a security perspective, but it also makes it possible to prompt for MFA. 

I'd definitely consider keeping things simple and consolidating to a  `aws-vault credential-server` or similar, but we'd need to address the MFA prompt. It would work fine with the applescript mfa prompt, but not sure how good the usability would be on other platforms.

I don't mind the idea of requiring users to run it in another console tab. It takes away some of the magic and will perhaps mean people adopt it more. 

Thoughts @mtibben / @pda?

lox:
I generally dislike the fact that we call sudo ourselves anyway. We could drop that and leave it up to the user to use sudo as makes sense.

williamdenton:
I've tried automatically starting the `server` in the background in the same way it does now but without `sudo -b`. As its not background anymore i need to change `cmd.Run()` to `cmd.Start()`

https://github.com/99designs/aws-vault/blob/master/server/server.go#L73-L80

however exec is still expecting to execute another command and crashes when it cant

this makes it happy, but clearly im missing something
```sh
aws-vault exec playpen --server -- cmd /C  pause
```
now the process that is bound to port 80 is long lived and continues to run once started. Obviously this requires an elevated shell to run the first time (windows equivalent of sudo).


I feel like I'm missing a key bit of knowledge about how the metadata server is supposed to be executed.

lox:
So there are two http servers:

* The "proxy" server starts up in the background and binds to `169.254.169.254:80`, which is what the aws tooling connects to. It needs root permissions to create the iface and bind to a port < 1024. This is what is triggered by `aws-vault server`. It's meant to be pretty useless on it's own, and simply proxies requests to `localhost:9099` if it exists.

* The "credential" server is started at the beginning of an `aws-vault exec --server` invocation and it only runs for the duration of that command. It's the one with the keys to the kingdom, so it binds to `localhost:9099`. 

So with those two servers in place, the aws cli tools can request IAM creds, and refresh them as they go (refreshing them is important, is it keeps the lifetime low). 

The lines you referred to (https://github.com/99designs/aws-vault/blob/master/server/server.go#L73-L80) just check that the "proxy" is running, as the "credential" server isn't much use without it.

Does that help?

lox:
It seems like the hard part on windows is going to be the "proxy" server. I'd be tempted to require that windows users manually invoke the proxy server component:

```go
	if !checkServerRunning(metadataBind) {
                if runtime.GOOS == "windows" {
                    app.Fatal("The credential proxy server isn't running. Run aws-vault server as Administrator in the background and then try this command again")
                }
		log.Printf("Starting `aws-vault server` as root in the background")
		cmd := exec.Command("sudo", "-b", os.Args[0], "server")
		cmd.Stdin = os.Stdin
		cmd.Stdout = os.Stdout
		cmd.Stderr = os.Stderr
		if err := cmd.Run(); err != nil {
			return err
		}
	}
```

lox:
The more I think about it, I think the two part server design is better. It prevents people from running long lived credential servers, which basically negate the security value of aws-vault.

lox:
Have we lost you @williamdenton? :)

williamdenton:
Sorry, no, just busy with another project currently, will return here shortly.

Will try the suggestion above and update PR soon

mtibben:
Yep think I agree @lox, 2 part makes a lot of sense.

I did some googling, and on Windows, [the runas command might work](https://stackoverflow.com/a/5953967) to run the command as Administrator, although I haven't tried this myself.

I have a few suggestions on the UX around the server starting 

1. If the credential service hasn't been detected but we can start it ourselves, tell the user what we're doing
   ```
   $ aws-vault exec --server home -- aws s3 ls
   Starting the aws-vault credential service.
   + sudo aws-vault server
   Password:
   ```

2. As a fallback if sudo/runas isn't present or if we're on windows we can prompt the user to start the service 
   ```
   $ aws-vault exec --server -- aws s3 ls
   In order to run in --server mode, run the following command in your console
   to start the aws-vault credential service

       sudo aws-vault server
       runas /user:Administrator "aws-vault server"

   ```

williamdenton:
this solution has very similar behavior to other platforms now:
* it starts the server in the background if not started (but requires the main process to be as admin)
* once the background server is running the actual credential server does not require administrative privileges

I think this is the best option for keeping the behaviour consistent on all platforms 

```bash
.\aws-vault.exe exec playpen -s -- aws s3 ls
Enter passphrase to unlock ...\.awsvault\keys:
Creation of network alias for server mode requires elevated permissions (Run as administrator).
Unable to locate credentials. You can configure credentials by running "aws configure".
```

the `Unable to locate credentials...`  comes from the aws command in the exec


Because I'm using `cmd.Start()` to launch the background server process i can't tell if it failed to launch, i can't see a way to inspect the cmd object to tell if it has terminated. there is `Wait()` but it doesn't appear to take a timeout...

williamdenton:
@lox this seems to get it functional and usable on windows.

lox:
I'll review in detail shortly! Thanks so much @williamdenton!

lox:
I wonder if we could use `start /B "" program` to run it in the background?

server/server.go
```diff
@@ -76,13 +78,28 @@ func checkServerRunning(bind string) bool {
...
+			}
+		} else {
+			log.Printf("Starting `aws-vault server` as root in the background")
+			cmd := exec.Command("sudo", "-b", os.Args[0], "server")
+			cmd.Stdin = os.Stdin
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			if err := cmd.Run(); err != nil {
+				return err
+			}
```
mtibben: Might read a bit nicer if we pull this logic out into 2 functions, `StartCredentialsServer` and `StartCredentialsServerWithSudo`? And might be worth doing the `checkServerRunning` in both cases - there are failure scenarios if sudo doesn't exist for example

williamdenton: if sudo doesn't exist then we still get to see the error as we start sudo with [`cmd.Run()`](https://golang.org/pkg/os/exec/#Cmd.Run). This launches sudo and waits for it to finish - which it does as soon as it starts the process in the background `-b`

williamdenton:
a couple of questions about merging:
Do I need to squash my commits before this gets merged?
or Do you merge with all the commits? 
Do you want signed commits?

lox:
> a couple of questions about merging:
> Do I need to squash my commits before this gets merged?
> or Do you merge with all the commits?
> Do you want signed commits?

It's fine how it is! Basically as soon as it's a 👍 from two of us, we'll merge it.

server/server.go
```diff
@@ -74,14 +76,41 @@ func checkServerRunning(bind string) bool {
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func StartCredentialProxy() error {
+	if runtime.GOOS == "windows" {
+		return StartCredentialProxyOnWindows()
+	}
+	return StartCredentialProxyWithSudo()
```
mtibben: @williamdenton Actually another way to do this is put the arch specific functions into conditional build files, rather than using the runtime check. What's preferred @lox ?

lox: It gets complicated, let's stick to this switch. LGTM.

PR #181, opened 2017-11-08
Import and Export commands

lox:
This adds a command for importing plain text credentials from `~/.aws/credentials`, and a command to export credentials to plain text credentials. 

This is useful for migrating between vaults, or migrating from other tools to aws-vault.

### DIFF
--- a/cli/add.go
+++ b/cli/add.go
@@ -44,12 +44,6 @@ func ConfigureAddCommand(app *kingpin.Application) {
 func AddCommand(app *kingpin.Application, input AddCommandInput) {
 	var accessKeyId, secretKey string
 
-	if source, _ := awsConfig.SourceProfile(input.Profile); source.Name != input.Profile {
-		app.Fatalf("Your profile has a source_profile of %s, adding credentials to %s won't have any effect",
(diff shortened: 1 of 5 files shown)


sj26:
Nice!

lox:
I can't decide if `export` is an anti-pattern in terms of security/UX trade-off. It makes it very easy to dump somone's credentials if they aren't logged in and they encourage persisting credentials to disk.

Might a single `copy` command be better? Something like:

```bash
aws-vault copy ~/.aws/credentials keychain:default
aws-vault copy keychain:aws-vault keychain:login --profile my-profile
```

FernandoMiguel:
@lox i have to agree.. exporting to clear text would make the entire tools less secure

lox:
Yup, ok. Gonna close this one out.

acdha:
I agree with the decision not to have an export feature but wouldn't it still be useful to have an import feature? My use-case would be having a bunch of developers who have `~/.aws/credentials` files with multiple profiles and wanting a relatively simple way to switch over to using `aws-vault` without an error-prone copy-and-paste step.

PR #185, opened 2017-11-13
spoof an instance ID so we can pass the go aws-sdk Available() call

### LINKED ISSUE #184: Go SDK doesn't detect credentials are available when running in server mode
The [AWS Go SDK](https://github.com/aws/aws-sdk-go/blob/master/aws/ec2metadata/api.go#L128) checks 169.254.169.254/latest/meta-data/instance-id to see if the metadata service is available. aws-vault does not provide this endpoint, and so any libraries using the EC2Metadata.Available() call will return an error.

PR to follow!

sjauld:
Resolves #184 - the Go AWS SDK can now hit the instance-id endpoint when testing for the availability of the metadata endpoint.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -29,6 +29,8 @@ func StartMetadataServer() error {
 	router := http.NewServeMux()
 	router.HandleFunc("/latest/meta-data/iam/security-credentials/", indexHandler)
 	router.HandleFunc("/latest/meta-data/iam/security-credentials/local-credentials", credentialsHandler)
+	// The AWS Go SDK checks the instance-id endpoint to validate the existence of EC2 Metadata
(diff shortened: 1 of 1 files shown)


sjauld:
I was trying to run Terraform acceptance tests in Docker without having to pass in AWS_ACCESS_KEY_ID etc. as envs. I figured server mode was a real winner... Sadly Terraform uses the EC2Metadata.Available() call while setting up AWS creds.

lox:
Could I trouble you to leave a comment explaining why we have that method to prevent it being removed in future?

sjauld:
Commented!

pda:
I like it, and good idea returning `aws-vault` as the instance ID; I can imagine that saving somebody much confusion.

PR #186, opened 2017-11-16
set AWS_SECRET_KEY

bandyson:
Support AWS's inconsistent naming convention by setting a `AWS_SECRET_KEY` variable in addition to `AWS_SECRET_ACCESS_KEY`.

giphy 21

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -132,9 +132,10 @@ func ExecCommand(app *kingpin.Application, input ExecCommandInput) {
 	}
 
 	if setEnv {
-		log.Println("Setting subprocess env: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY")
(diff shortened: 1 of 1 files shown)


pda:
Any examples of programs/libraries that use `AWS_SECRET_KEY` but not `AWS_SECRET_ACCESS_KEY`?

lox:
@bandyson the SDK's have standardized on `AWS_SECRET_ACCESS_KEY`, I'd prefer not to add this if we don't have to, what app are you having problems with?

bandyson:
hi @lox. i was having trouble building a java app via maven but on further examination i don't think the problem is related o this issue. 

i'll go ahead and close this PR.

lox:
I did a bit of a search, and it looks like the only common one is the AWS Elastic Beanstalk client.

PR #191, opened 2017-11-20
Bump go-ini version to supported "nested config"

### LINKED ISSUE #187: `aws-vault add` borks ~/.aws/config when using s3v4 signatures
This one is a little bit difficult because the [AWS documentation around s3v4 signatures](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingAWSSDK.html) and [cli configuration](http://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html) is not amazing.

To reproduce:-
1. Check that it works as expected
    ```
    $ aws configure set default.s3.signature_version s3v4
    $ cat ~/.aws/config
    [default]
    s3 =
        signature_version = s3v4
    ...
    $ aws-vault exec my-profile -- aws s3 ls my-bucket
        PRE my-folder/
        PRE my-other-folder/
    ```
2. See what happens when aws-vault modifies the config file
    ``` 
    $ aws-vault add bob
    Enter Access Key ID: mygreatkey
    Enter Secret Access Key: itsasecret
    Added credentials to profile "bob" in vault 
    $ cat ~/.aws/config
    [default]
    s3=
    signature_version=s3v4
    ...
    $ aws-vault exec my-profile -- aws s3 ls my-bucket
    'str' object has no attribute 'get'
    ```

Trial and error seems to indicate that a whitespace character is required before signature_version for this config to work. Not sure s3v4 is the only configuration setting this affects or if other AWS services have similar config settings. I'll see if I can put a fix into Config.

lox:
Closes #187.

Thanks so much to @Unknwon for https://github.com/go-ini/ini/issues/131! 🙇🏼 🙌🏻 💪🏼

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -66,7 +66,9 @@ func LoadConfigFromEnv() (*Config, error) {
 
 func (c *Config) parseFile() error {
 	log.Printf("Parsing config file %s", c.Path)
-	f, err := ini.Load(c.Path)
+	f, err := ini.LoadSources(ini.LoadOptions{
+		AllowNestedValues: true,
+	}, c.Path)
 	if err != nil {
 		return fmt.Errorf("Error parsing config file %q: %v", c.Path, err)
 	}
--- a/vault/config_test.go
+++ b/vault/config_test.go
@@ -1,6 +1,7 @@
 package vault_test
 
 import (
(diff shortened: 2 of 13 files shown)


mtibben:
lgtm

PR #193, opened 2017-11-26
Bump keyring version

### LINKED ISSUE #188: aws-vault exec → 4 password prompts
On macOS Sierra 10.12.6 (16G1036) I'm currently getting four password prompts from aws-vault 4.1.0 (latest) for the first `aws-vault exec $role …` of the day:

image

Also, for subsequent runs within the session TTL, I previously just got a button prompt, but now I get a password prompt:

screen shot 2017-11-17 at 9 55 20 am

Perhaps some would go away if I clicked Always Allow but I've never done that for aws-vault. I'm accustomed to one or two password prompts for a non-existing-session `aws-vault exec`.

What are other people seeing/doing?

lox:
This updates to the latest https://github.com/99designs/keyring API, which is a fairly major change. I've checked that we are setting the correct defaults so that it's backwards compatible on everything but kwallet and libsecret, but have eyeballed them.

Closes #188.

### DIFF
--- a/cli/add_test.go
+++ b/cli/add_test.go
@@ -29,7 +29,7 @@ func ExampleAddCommand() {
 	app := kingpin.New(`aws-vault`, ``)
 	ConfigureGlobals(app)
 	ConfigureAddCommand(app)
-	kingpin.MustParse(app.Parse([]string{"add", "--env", "foo"}))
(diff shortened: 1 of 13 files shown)


lox:
I'm going to merge this next week, let me know if there are any objections @Vektah / @mtibben

pda:
Nice work 🎉

PR #197, opened 2017-12-03
Allow for different macOS Keychain to be used

lox:
This supersedes #144. It allows for a specific macOS keychain to be used, which means that either the default/ login keychain can be used rather than the previous `aws-vault.keychain` that was created. 

The default remains `aws-vault`, so it's backwards compatible.


mtibben:
Yep I think this is the best first step, we could potentially switch to default `login` keychain at the next major version

PR #199, opened 2017-12-05
searching for legacy keychain if not specified

heph:
if the keychain isn't specified, searches for `aws-vault` followed by `login` to support legacy installations while pushing new users towards the login keychain.

### DIFF
--- a/cli/global.go
+++ b/cli/global.go
@@ -9,12 +9,13 @@ import (
 	"github.com/99designs/aws-vault/prompt"
 	"github.com/99designs/aws-vault/vault"
 	"github.com/99designs/keyring"
+	gokeychain "github.com/keybase/go-keychain"
 	"golang.org/x/crypto/ssh/terminal"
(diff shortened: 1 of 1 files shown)


heph:
@lox plx review

lox:
Neat! I'm just in the process of trying to get my latest few PR's merged in, Travis is giving me all sorts of problems. Let me settle them in and then I'll provide some feedback.

lox:
@heph given we've gone back to defaulting to aws-vault and letting people opt-in to a new vault, I reckon we might hold off on this check until we change the default. Thoughts?

mtibben:
So leave default at aws-vault for upcoming v5 (with option and env var for login?), and switch to login for v6?  We probably want to think about migrating creds between the keychains in v6 also

lox:
Yeah I think so @mtibben. We'll add a `copy` command in 5.1 that will prepare for that too.

Issue #198, opened 2017-12-04
Enhancement: Allow whitelisting of senders to metadata service with the --server option


clintoncampbell:
As it is defined right now, the local metadata service stub is attached to 169.254.169.254 and addressable by any host on the local link (leveraging static ARP entries to associate the target's MAC address). It's quite easy, on most networks, to identify potential targets and enumerate the service across multiple hosts.

While I can implement some local firewall rules to lock this down, that provision is out of reach for many developers. Ideally, the server would default to rejecting any outside traffic and provide a basic level of protection for the simple use case.

lox:
Isn't `169.254.169.254` non-routable?

lox:
I'd love to figure out a way to do this if possible!

lox:
It also binds to the loopback interface doesn't it? I couldn't think of any way for network traffic to access it when I was implementing it. Hostile web pages could possibly interact with it with CSRF though.

clintoncampbell:
169.254/16 is not routable but it is available on the LAN unless there is some sort of client isolation in use. Client isolation is pretty rare outside larger enterprises and campuses. Inbound firewall ought to provide some protection, but that's not as ubiquitous as it ought to be, e.g., off by default on macOS and some linux distros. 

At least on macOS, the metadata address doesn't seem to be tightly bound to loopback, and I could imagine this behavior being somewhat platform dependent. The easy test is to have a colleague on the same wireless or ethernet set a static arp entry binding the metadata address to your NIC before hitting the path in a browser or curl. 

The upside is that the machine (at least on macOS) does not appear to respond to ARP or MDNS requests, so it does require knowledge of the target's MAC. However, most machines tend to be fairly chatty when they join the network as they try to identify shared resources.

My first thought on a cross-platform solution would be to look at the sender IP in the server and reject anything that doesn't come from 169.254.169.254. This should allow anything (VM or container) on the localhost to communicate as long as NAT is running. Outside users may be able to spoof inbound traffic, but the response won't (or shouldn't) leave the machine.

The more robust solution would be to use native firewall functionality, but I don't think there's a cross platform solution (though it would be really freaking cool). Not sure I can get to it for a couple of days, but I'll let you know if I come up with a good _pf_ option for macOS. _iptables_ on Linux ought to be fairly easy as well. Will post back if I think of any other options!

clintoncampbell:
[...]

PR #205, opened 2017-12-16
Prevent local network segment from accessing credential server

lox:
Via #198 we were made aware that it was possible to route traffic to `169.254.169.254:80` even if it is aliased to `lo0`. This means anyone on the same network segment could access the credential server whilst an `aws-vault exec --server blah` command was running. 

This adds a check for a `RemoteAddr` of `127.0.0.1`. I don't love this approach, would much prefer something lower down, but it seems like a pragmatic mitigation in the short term. Any thoughts @mtibben @clintoncampbell?

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -122,6 +122,20 @@ func StartCredentialsServer(creds *vault.VaultCredentials) error {
 
 	log.Printf("Local instance role server running on %s", l.Addr())
 	go http.Serve(l, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ip, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		// Must make sure the remote ip is localhost, otherwise clients on the same network segment could
+		// potentially route traffic via 169.254.169.254:80
+		if ip != `127.0.0.1` {
+			http.Error(w, "Access denied from non-localhost address", http.StatusUnauthorized)
+			return
+		}
+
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -122,6 +122,20 @@ func StartCredentialsServer(creds *vault.VaultCredentials) error {
...
 	go http.Serve(l, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ip, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		// Must make sure the remote ip is localhost, otherwise clients on the same network segment could
+		// potentially route traffic via 169.254.169.254:80
+		if ip != `127.0.0.1` {
```
mtibben: You could possibly use [IP.IsLoopback](https://golang.org/pkg/net/#IP.IsLoopback) here

mtibben:
From what I've been seeing, it looks like the ability to route link-local IPs on any interface (even loopback interfaces) seems like a "feature"... 

> No configuration is necessary

 \- https://developer.apple.com/library/content/qa/qa1357/_index.html

I think the only lower-level fix would be to either put some routing rules or firewall rules in place - however getting those to persist may be fiddly.

Agree this is a pragmatic mitigation for now 

:shipit:

clintoncampbell:
Yes, the behavior is fairly standard across *nix distributions although
it's rare for anyone outside of networking and operations professions to
encounter it. Given the capabilities of Go's SDK, I agree that this is the
right solution.

Not sure how quickly I can get to it, but I'll look at putting together
some firewall configuration scripts for common macOS and Linux that could
be referenced in the documentation.

PR #206, opened 2017-12-18
Route builds back to Xcode 7.3 image

cotsog:
See https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce


lox:
Thanks @cotsog! Perplexing what causes the error, I build this locally on XCode 8.3

PR #208, opened 2017-12-22
Remove server subcommand from `--help`

### LINKED ISSUE #207: aws-vault: error: Server failed: exit status 1
$ aws-vault server --debug
2017/12/20 11:49:21 Loading config file /Users/fernando/.aws/config
2017/12/20 11:49:21 Parsing config file /Users/fernando/.aws/config

$ aws-vault --version
v4.1.0

MacOS 10.13.3 Beta (17D20a)

lox:
Some notes:

> The "proxy" server starts up in the background and binds to 169.254.169.254:80, which is what the aws tooling connects to. It needs root permissions to create the iface and bind to a port < 1024. This is what is triggered by aws-vault server. It's meant to be pretty useless on it's own, and simply proxies requests to localhost:9099 if it exists.
> 
> The "credential" server is started at the beginning of an aws-vault exec --server invocation and it only runs for the duration of that command. It's the one with the keys to the kingdom, so it binds to localhost:9099.
https://github.com/99designs/aws-vault/pull/174#issuecomment-340921440

Closes #207.

### DIFF
--- a/cli/server.go
+++ b/cli/server.go
@@ -11,7 +11,9 @@ type ServerCommandInput struct {
 func ConfigureServerCommand(app *kingpin.Application) {
 	input := ServerCommandInput{}
 
-	cmd := app.Command("server", "Run an ec2 instance role server locally")
+	cmd := app.Command("server", "Run an ec2 instance role server locally").
+		Hidden()
+
 	cmd.Action(func(c *kingpin.ParseContext) error {
 		ServerCommand(app, input)
 		return nil


avram:
Is there a reason this PR is marked WIP? looks like the right thing to do-- and I've been confused many times by the tempting command that is in `aws-vault --help` that I'm not supposed to run.

lox:
@avram the docs are still in progress, but this is probably a useful change on it's own.

avram:
Right-- the actual docs could be improved as well but removing it from `aws-vault --help` is a big improvement as that is the only place someone would see that this subcommand even exists.

PR #212, opened 2018-01-17
Metadata endpoints for dotnet aws clients

williamdenton:
My current project is working with SQS Queues, i have found that these two additional metadata endpoints are required to allow the SQSClient to configure itself correctly when using the metadata endpoint.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -23,10 +23,15 @@ const (
 	localServerBind = "127.0.0.1:9099"
 )
 
-func StartMetadataServer() error {
+var (
+	metadataRegion string
+)
+
+func StartMetadataServer(region string) error {
 	if _, err := installNetworkAlias(); err != nil {
 		return err
 	}
+	metadataRegion = region
 
 	router := http.NewServeMux()
 	router.HandleFunc("/latest/meta-data/iam/security-credentials/", indexHandler)
@@ -76,9 +81,8 @@ func instanceIdHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func regionDiscoveryHandler(w http.ResponseWriter, r *http.Request) {
-	//the config source profile has a region defined, this should be used here
 	json.NewEncoder(w).Encode(map[string]interface{}{
-		"region": "us-west-2",
+		"region": metadataRegion,
 	})
 }
 
--- a/server/server.go
+++ b/server/server.go
@@ -33,6 +33,8 @@ func StartMetadataServer() error {
 	router.HandleFunc("/latest/meta-data/iam/security-credentials/local-credentials", credentialsHandler)
 	// The AWS Go SDK checks the instance-id endpoint to validate the existence of EC2 Metadata
 	router.HandleFunc("/latest/meta-data/instance-id/", instanceIdHandler)
+	// The AWS DotNet SDK checks the instance-id endpoint to validate the existence of EC2 Metadata
+	router.HandleFunc("/latest/meta-data/iam/info", iamInfoHandler)
 
 	l, err := net.Listen("tcp", metadataBind)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


lox:
I wish AWS would actually publish these things. Thanks @williamdenton!

server/server.go
```diff
@@ -73,6 +75,13 @@ func instanceIdHandler(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprintf(w, "aws-vault")
 }
 
+func regionDiscoveryHandler(w http.ResponseWriter, r *http.Request) {
+	//the config source profile has a region defined, this should be used here
+	json.NewEncoder(w).Encode(map[string]interface{}{
+		"region": "us-west-2",
```
williamdenton: i can't figure out how to get the config imported to the server module that contains the region from the config file, obviously we can't merge this hard coded version...

williamdenton: this code is a separate process to the actual credentials, so unless it gets proxied through as the credentials do its a bit stuck. Even if we pass its as a param on startup people may want to run different profiles with different regions and as this process is long lived thats not going to work.

urgh.

lox:
Huh. TIL https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html

server/server.go
lox: I reckon checking the AWS_REGION and AWS_DEFAULT_REGION environment would be legit, or falling back to us-east-1.

williamdenton:
rebased on master due to long time with activity, added region env var as suggested

williamdenton:
(this isn't working yet btw)

williamdenton:
it's a bit weird that `/server/server.go` is responsible for both running the metadata http server, but also launching the metadata service. These two things are related, but happen in different processes.

A future refactor might be to split these two responsibilities.

This shows up as a little bit of weirdness in that the var `metadataRegion` is used for two purposes, to pass the region through to the start arguments, but also in the server to populate the response.

lox:
Agreed @williamdenton. Perhaps we could move them into a `metadata` subpackage at some point.

williamdenton:
Let me know if there is anything you'd like me to change about my approach. I'm just hacking my way through as I'm not experienced in Go.

williamdenton:
Hey @lox Let me know if there is anything you'd like me to change about my approach. I'm just hacking my way through as I'm not experienced in Go.

server/server.go
```diff
@@ -23,16 +23,25 @@ const (
 	localServerBind = "127.0.0.1:9099"
 )
 
-func StartMetadataServer() error {
+var (
+	metadataRegion string
```
lox: Ooof. A global here is a bummer. We might need to rethink this approach.

williamdenton: yeah, i wasn't sure. Th server is initialized with this value, and i need somewhere to keep it so it can be used in the http handlers. The handlers are funcs so i can't pass it normally. 

How could we solve this?

williamdenton: im being lazy and using this same variable in the "start server" code further down (Line 135). It could be easily passed normally between the functions to avoid using this global

server/server.go
```diff
@@ -108,7 +131,8 @@ func StartCredentialProxy() error {
 	return StartCredentialProxyWithSudo()
 }
 
-func StartCredentialsServer(creds *vault.VaultCredentials) error {
+func StartCredentialsServer(creds *vault.VaultCredentials, region string) error {
+	metadataRegion = region
```
williamdenton: instead of using the global here it could easily be passed through the function calls as an argument

server/server.go
lox: I'm going to think on it a bit. We're talking about refactoring this file into some smaller pieces and I want to think about how that might play into things.

williamdenton:
Have you put any thought into how this code could be improved? Happy to rework stuff for the better.

williamdenton:
this is way out of date now - closing in favour of a rewrite of the relevant bits (possibly just region discovery)

Issue #222, opened 2018-02-13
Server as feature


astanciu:
The server part of this tool (--server) is extremely useful when doing local development that requires AWS services. Being able to simulate the metadata service with auto-renewing tokens is a blessing.

Having said this, the experience can be improved just a tad. Currently, the server only works if passed some other command, which requires doing something like
`aws-vault exec phenix --server sh`
or
`aws-vault exec phenix --server node -- -e 'console.log(\"Running\");setTimeout(()=>{},1000*60*60*24*10);'`

but even here, there are some very strange behaviors. For ex, if I start this in the morning, it will happily run all day w/o issues. If I forget to turn it off and my laptop goes to sleep overnight, the next day, i have to completely kill the terminal where this is running as stdin gets really messed up, then also kill the aws-vault processes.

It's such a useful feature, but it requires daily tinkering. Why not clean this up a bit and allow it as a standalone feature? Something like
`aws-vault server [profile]` 
that allows clean interrupts and maybe outputs some logs when it's used, or token is renewed?

FernandoMiguel:
you dont need to pass commands, as it will run as a daemon
`aws-vault --debug exec phenix --server`
is all you need. anywhere in your laptop will reach the metadata server.

it is a terrible security risk to keep this long running (like next day). please make sure you terminate your session when you dont need it

astanciu:
> as it will run as a daemon

Does it though? It seems to open a new shell, and the server is only working while that shell's process is running. If you exit that shell, the server stops.

FernandoMiguel:
that is the current behaviour.
but you can access it from anywhere

astanciu:
Right, that's what this issue was suggesting, improving this feature and making it a bit friendlier, more prominent.. for ex, running --server would just display the output of the server, allowing me to kill it with ctrl-c. In its current state, I have to take extra care not to lose track of this shell and accidentally close it (when you have 10+ tabs in iTerm, this can be an issue)

FernandoMiguel:
given that for security reasons you shouldnt be long-running this, if you kill it, should be a problem... just launch it again.

astanciu:
This app shouldn't be prescriptive in how I choose to manage my security. All I'm saying is that the current behavior is very unnatural, for any cli command. Its just weird to run a command which just gives you a new shell, without any indication of what's happening behind the scenes. It's just not good UX, imo.

lox:
[...]

PR #224, opened 2018-02-13
Allow running just server

astanciu:
Allows running just 
`aws-vault exec --server [PROFILE]`

 without spawning a new shell, waits for ctrl-c to end. In reference to #222

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	"fmt"
 
 	"github.com/99designs/aws-vault/prompt"
 	"github.com/99designs/aws-vault/server"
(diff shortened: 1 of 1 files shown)


lox:
Thanks @astanciu! Unfortunately this would be a breaking change, without a command `aws-vault exec [PROFILE]` should execute `$SHELL`. Special casing this for `--server` would be unexpected I reckon.

One approach to this would be to add a `--server-only` argument that did what you have implemented here. 

Beyond that, we should just figure out why the process doesn't cleanly die on suspend.

lox:
>To me aws-vault exec means “execute this command” and aws-vault exec --server means “execute this command with a credential server available to it”. Then the assumption is the credential server ends when the command ends.

Me too, funnily enough.

astanciu:
>  “run a credential server available to the whole system until the user tells it to stop”

This is exactly the desired feature here, and I fully agree with the "exec" semantics. Using --server-only, or updating what `aws-vault server` do seem like more reasonable options (the latter of which makes most sense).

lox:
Ok, let's aim to update `aws-vault server` so that it can run on it's own, and perhaps make the current behaviour something like `aws-vault server --bind-only`?

The only complication is that we need sudo for that port bind.

PR #226, opened 2018-02-21
Make secret-service use collection 'awsvault'

gunnaringe:
This fixes a regression from v4.1.0 to v4.2.0, where collection
was renamed to 'aws-vault'

This is caused by that github.com/99designs/keyring previously
used the collection name 'awsvault' by default.
This was changed in commit 5a0d04e 'Add backend specific config'.


lox:
Many thanks!

PR #227, opened 2018-02-26
Add a HintAction to exec to let KingPin autocomplete profile names.

stoggi:
KingPin can provide dynamic values to it's bash-completion with the HintAction() API
https://github.com/alecthomas/kingpin#additional-api

Using this, you can have auto-completion on the profile names in ~/.aws/config when running:
```
aws-vault exec <tab>
```

You can see what will be fed to bash autocompletion with:
```
aws-vault --completion-bash exec
```

I tried to add a test in exec_test.go, but ran into trouble trying to generate a test config like in config_test.go. Some help here would be appreciated.

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -59,6 +59,7 @@ func ConfigureExecCommand(app *kingpin.Application) {
 
 	cmd.Arg("profile", "Name of the profile").
 		Required().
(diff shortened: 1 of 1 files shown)


lox:
Sounds like a good idea! Thanks!

PR #236, opened 2018-03-15
Move the signal setup closer to cmd.Start()

Lowercases:
aws-vault captures SIGINT in order to forward it to the spawned process.

Since the signal capturing is set up way before the child process is
spawned, in particular before getting the credentials, the signals are
captured at the mfa device prompt making it impossible to exit the CLI
using Ctrl-C at that point.

This moves the signal set up closer to when the child process is spawned,
making it possible to exit using Ctrl-C at the mfa prompt.

Tested on linux/KDE/kdewallet backend


joho:
This seems like a sensible change. @lox can you forsee any unexpected issues?

lox:
Seems very sensible, thanks @Lowercases

Lowercases:
Thanks for looking into it!

PR #240, opened 2018-04-03
Increase MaxAssumeRoleDuration to 12 hours

xmonkee:
Now allowed

https://aws.amazon.com/blogs/security/enable-federated-api-access-to-your-aws-resources-for-up-to-12-hours-using-iam-roles/

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -20,7 +20,7 @@ const (
 	MaxSessionDuration    = time.Hour * 36
 	MinSessionDuration    = time.Minute * 15
 	MinAssumeRoleDuration = time.Minute * 15
-	MaxAssumeRoleDuration = time.Hour
+	MaxAssumeRoleDuration = time.Hour * 12
 
 	DefaultSessionDuration    = time.Hour * 4
 	DefaultAssumeRoleDuration = time.Minute * 15


lox:
Have you tried this out @xmonkee?

lox:
Looks like the role needs to be updated:

`aws iam update-role -–role-name ADFS-Production -–max-session-duration 14400`

xmonkee:
Hmm, I seem to have jumped the gun a bit. I'm getting the following error:

`The requested DurationSeconds exceeds the 1 hour session limit for roles assumed by role chaining.`

I'm not exactly sure where the chaining is occuring.

From https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_terms-and-concepts.html:

```
Role chaining
Role chaining occurs when you use a role to assume a second role through the AWS CLI or API. For example, assume that User1 has permission to assume RoleA and RoleB. Additionally, RoleA has permission to assume RoleB. You can assume RoleA by using User1's long-term user credentials in the AssumeRole API operation. This operation returns RoleA's short-term credentials. To engage in role chaining, you can use RoleA's short-term credentials to assume RoleB.

Role chaining limits your AWS CLI or API role session to a maximum of one hour. Given the previous example, if you use the AssumeRole API operation to assume RoleB using the temporary credentials for RoleA, you can specify the duration of your role session with the DurationSeconds parameter. You can specify a parameter value of up to 43200 seconds (12 hours), depending on the maximum session duration setting for your role. Because of this limit, if you provide a DurationSeconds parameter value greater than one hour when you assume the role using chaining, the operation fails.
```

lox:
Heh, yup, that was what I thought, but appreciate you diving into a PR! It's unfortunately going to be more complicated than this as folks have different combinations of assumed roles, federated logins and STS session.

FernandoMiguel:
Changing aws-vault is not enough.
As I mentioned before, the role needs to be setup to allow more than the default. 
So either the user does that manually, and aws-vault tests for it,
Or
Aws-vault needs to make an api call to change the role max value

lox:
I'm going to close this in the meantime.

FernandoMiguel:
That's sad news.
Terraform already released this change, so I can easily modify my roles to support this. 
I had already changed the main ones via cli 
All it's missing is aws vault support

rdonkin:
I would be fine to configure the role for longer max duration outside aws-vault - in fact I think it's better to use Terraform / CloudFormation to manage roles, rather than have aws-vault making changes. The (AWS blog mentioned above)[https://aws.amazon.com/blogs/security/enable-federated-api-access-to-your-aws-resources-for-up-to-12-hours-using-iam-roles/] covers how to do this.

However, there's clearly some work for aws-vault to support this by changing its API calls.

lox:
Let's just merge this and we can figure out how to make it easier to modify the roles later.

FernandoMiguel:
Agreed it's the mvp approach.

PR #243, opened 2018-04-09
Allow exposing master credentials if asked

mtparet:
### Description
Add the option `--use-master-keys` to use master keys.

### Why ?
I know that this is not a good security practice. 

But we are going to use `aws-vault` withing server for long living application. 
For long living application we need to provide ec2 metadata so the aws sdk can refresh itself the credentials.
The way to provide these ec2 metadata is to use the option --server. To correctly integrates everything in our infrastructure, we need to be able to run this server within a container but this does not work today.

As it will takes times to fix all these issues, I add this workaround so we could even use `aws-vault` right now and have a smoothly migration in the future. 

I will open an issue about the several current issues using `aws-vault exec --server` inside docker

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -27,7 +27,14 @@ $ aws-vault exec work -- aws s3 ls
 another_bucket
 ```
 
+## Using master credentials
+In case you have a long living application and the server solution could
+not work for you, you can use master credentials. 
 
+It reduces security level as it exposes permanent credentials so we recommend using it only if needed and rotating the keys on a regular basis. 
+```bash
+$ aws-vault exec work --use-master-keys -- <long living application, ie: rails s >
+```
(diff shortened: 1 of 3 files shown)


FernandoMiguel:
isnt that the idea of --no-session?

FernandoMiguel:
https://github.com/99designs/aws-vault/blob/master/cli/exec.go#L37

mtparet:
Indeed, I will check this... thanks!

myoung34:
Isnt [this a better fit](https://github.com/lyft/metadataproxy) ? Im having a hard time following what this solves. It seems like you have a unique edge case and want the tooling to account for it...

mtparet:
Yes, it could be a better fit :)  But I can't find anything about master credentials encrypted and a way to easily rotating these.

PR #264, opened 2018-06-05
Read external_id into Profile and pass to assume role requests

AlexRudd:
A quick fix to support the [external_id option](https://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html#cli-roles-xaccount) in profile config.

Have tested this against one of my own roles.


iamjoeker:
Downloaded and built it this evening. I've been using it for hours across multiple accounts. Looking forward to seeing this merged!

PR #271, opened 2018-08-01
Update go-keychain to resolve Go 1.11 compilation issue

kevinburkeomg:
Previously compiling aws-vault on Go 1.11 beta 2 would fail with a
compilation error. For more information see:

https://github.com/golang/go/commit/94076feef
https://github.com/golang/go/issues/26721
https://github.com/keybase/go-keychain/pull/30

Also, add tip to the Travis CI build matrix. And run tests on the latest point
release for Go 1.8, 1.9, and 1.10, so 1.10.3 instead of 1.10 for example.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,18 @@
 language: go
+go_import_path: github.com/99designs/aws-vault
 
 install:
   - go get -u github.com/kardianos/govendor
   - govendor status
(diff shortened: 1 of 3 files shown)


lox:
Thanks for your hard work on this @kevinburkeomg!

Issue #277, opened 2018-08-21
Add option to open a specific AWS service directly 


maishsk:
When using `aws-vault login` the command by default generates a URL that takes you to the main console - which is almost never ever useful - because it requires even further drilling down into the specific service I want to access. 

What would be awesome would be another flag in the `login` that would be passed to the specify the service I want to go to, for example:

`aws-vault login MYPROFILE --service ec2` and would take me directly to the EC2 console.

PR #278, opened 2018-08-21
Add service option

maishsk:
This is an enhancement that addresses [#277](https://github.com/99designs/aws-vault/issues/277)

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -30,6 +30,7 @@ type LoginCommandInput struct {
 	UseStdout               bool
 	FederationTokenDuration time.Duration
 	AssumeRoleDuration      time.Duration
+	Service                 string
 }
 
 func ConfigureLoginCommand(app *kingpin.Application) {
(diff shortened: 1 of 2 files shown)


cli/login.go
```diff
@@ -169,14 +174,21 @@ func LoginCommand(app *kingpin.Application, input LoginCommandInput) {
...
-	}
+  
+  destination := "https://console.aws.amazon.com/"
+  if input.Service != "" {
+  	destination = fmt.Sprintf(
+  			"https://console.aws.amazon.com/%s",
+  			input.Service,
+  		)
+	  } else {
+  	if profile, _ := awsConfig.Profile(input.Profile); profile.Region != "" {
```
lox: So service won't be handled for specific regions?

maishsk: :) I was waiting for this question. 
There are certain endpoints - that are global across all AWS - and there are others that are region specific.

The behavior that currently happens is that the browser will go into the region that you last worked in. (I assume it is a cookie that is used - AWSCRegion or noflush_Region) where the region is stored

Do you prefer having it region specific? I do not think that it is necessary

mtibben:
Fixed in #291

Issue #282, opened 2018-09-02
Server: Add handler for `latest/meta-data/iam/info`


amitsaha:
AWS.NET SDK's KMS client (and may be others) tries to access the above endpoint during a `decrypt` (and perhaps other) operations. The result is the API request fails with a 404 and the request is aborted. I think, it's not specific to KMS client, but I haven't looked deeply into it. 

To reproduce:

- Setup an IP address alias for 169.254.169.254 for the loopback interface
- Create a new directory `test` somewhere in your file system
- Create the following directory structure inside `test`:

```
└── latest
    └── meta-data
        └── iam
            └── security-credentials
                └── testrole
```
(where `testrole` is a file with contents `testrole`)

- Run a HTTP server from inside the `test` directory: `sudo python3 -m http.server 80 --bind 169.254.169.254 `

Now, get the code at `https://github.com/amitsaha/dotnet-demos/tree/master/AWSSDKDemo` and
`dotnet run`.

On the terminal where the HTTP server was started you will see:

```
169.254.169.254 - - [02/Sep/2018 10:32:49] "GET /latest/meta-data/iam/security-credentials/ HTTP/1.1" 200 -
169.254.169.254 - - [02/Sep/2018 10:32:49] code 404, message File not found
169.254.169.254 - - [02/Sep/2018 10:32:49] "GET /latest/meta-data/iam/info HTTP/1.1" 404 -
169.254.169.254 - - [02/Sep/2018 10:32:49] code 404, message File not found
169.254.169.254 - - [02/Sep/2018 10:32:49] "GET /latest/meta-data/iam/info HTTP/1.1" 404 -
```

On the client side, you will see a traceback.

PR in https://github.com/99designs/aws-vault/pull/283

PR #283, opened 2018-09-02
Add handler for iam/info

amitsaha:
Fix for https://github.com/99designs/aws-vault/issues/282

The response is intentionally set to a minimal valid JSON string.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -29,6 +29,8 @@ func StartMetadataServer() error {
 	}
 
 	router := http.NewServeMux()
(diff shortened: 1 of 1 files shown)


amitsaha:
MAINTAINERS - please take a look and if it looks good, can we have a new release please? I introduced this to our dev team and would be massively useful, if I have a new release to point them to rather than distributing binaries myself :)

amitsaha:
Thanks @mtibben !

PR #285, opened 2018-09-03
login endpoints for region

frezbo:
This PR does the following:
    
- Fixes: https://github.com/99designs/aws-vault/issues/258
- Pulls in changes from: https://github.com/99designs/aws-vault/pull/259 and addressed the PR comments
- Pass the region to the client
- Fix a failing `go vet`

Special thanks to @sprice-janrain
    
Signed-off-by: Noel Georgi <git@frezbo.com>

PR #291, opened 2018-09-23
Add service option initially implemented by @maishsk

frezbo:
This is a feature enhance to #278 by @maishsk. Thanks a lot for the idea. I tested this both in commercial and gov-cloud region and seems to work like a charm. This also takes cares of the region issue as per #285 .

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -32,6 +32,7 @@ type LoginCommandInput struct {
 	FederationTokenDuration time.Duration
 	AssumeRoleDuration      time.Duration
 	Region                  string
+	Service                 string
 }
 
 func ConfigureLoginCommand(app *kingpin.Application) {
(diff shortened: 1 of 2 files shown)


mtibben:
Seems like most, but not all, AWS services have the URL form `https://{host}/{service}/home`, though `/{service}` seems to redirect

Is `--service` right here, or would something like `--url-path` be clearer about what this is for? It's possible to deep link into a particular URL in aws web console using this flag too, which `service` doesn't capture

frezbo:
@mtibben so you want `service` along woth `--url-path` for nested services. BTW nice idea. I will try to see if I can spend some time on this.

mtibben:
I was thinking just `--url-path`?

maishsk:
👍 
LGTM

PR #297, opened 2018-10-05
log when login command ignores MFA

q3aiml:
Hi!

First, this might just be me being completely dense. If so I will not be at all offended if this is simply closed. I am putting this out for your thoughts because it took me more time than I would like to admit to figure out why the login command cannot support MFA without assuming a role. Something closer to a warning actually feels more appropriate to me, since I assumed profiles configured for MFA would use it unless `--no-session` is specified, but I can also understand if that would be useless noise for others.

Thanks for looking.

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -89,6 +89,15 @@ func LoginCommand(app *kingpin.Application, input LoginCommandInput) {
 		return
 	}
 
+	if provider.HasMfa() && !provider.HasRole() {
+		// Federated login requires GetFederationToken or AssumeRole tokens. Only AssumeRole and GetSessionToken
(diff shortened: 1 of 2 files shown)


lox:
Wow, sorry it's taken so long to review this 😓This seems like a pretty sensible change to me. @mtibben any objections to merging this in?

mtibben:
@q3aiml can you please rebase on master?

q3aiml:
Thanks for your time taking a look! Unfortunately my tech stack has changed. Sorry for not cleaning this up sooner. I'll go ahead and close this but if anyone finds it useful who is able to verify a rebased version please feel free to revive in a new PR.

PR #305, opened 2018-10-17
Document stdout flag

QuinnyPig:


pda:
Nice. Here's where the flag is defined: https://github.com/99designs/aws-vault/blob/master/cli/login.go#L63-L65

PR #306, opened 2018-10-30
Add custom shell completion scripts

### LINKED ISSUE #70: bash completion of exec command

segevfiner:
They are based on the shell completion scripts generated by kingpin in alecthomas/kingpin#252. Modified to complete commands after '--'.

Fixes #70

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-aws-vault
-aws-vault-*
+/aws-vault
+/aws-vault-*
 .DS_Store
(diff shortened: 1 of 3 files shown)


lox:
Awesome, I've been hoping someone would implement this for ages! ❤️

Issue #219, opened 2018-02-07
Constantly have to re-enter keychain password, part 2


solsglasses:
I think that this issue https://github.com/99designs/aws-vault/issues/211 is still happening.  That issue was closed after it was believed to be fixed, but myself and another person are reporting that we are still seeing the problem.

lox:
Could you tell us a bit more about your config? Are you assuming a role, or just using creds? Does your profile have MFA?

At a minimum you will have your root credentials in your keychain and a session, so if you aren't ticking "Don't ask me again" you should expect 2 prompts and then 1 prompt. 

Are you running an official build? If so, what version?

seanscottking:
> Could you tell us a bit more about your config?
Absolutely, I am at your disposal, ask me anything you need to know

> Are you assuming a role, or just using creds?
Yes I originally experienced this problem (and decided to create this github issue) while I was assuming a role.

> Does your profile have MFA?
I am not using MFA

> Are you running an official build? If so, what version?
Installed via brew:
```
$ aws-vault --version
v4.2.0
```

seanscottking:
Hey @lox and team, I want to help, is there any other questions I can answer/things I can try out?

conzy:
If its any help I experience the same thing on OSX Sierra. Brew version `aws-vault --version
v4.2.0`

tomquisel:
I have the same issue, same version as the last poster. I use an assumed role, but actually not for the profile that has an issue. It's a vanilla profile with no MFA or roles.

stevencarpenter:
I am also experiencing this issue. I am running on mac 10.13.3. It is very frustrating to get repeatedly asked when assuming a role. I am running aws-vault 4.2.0. I am using MFA.

lox:
How many times are you prompted for a single usage? Both on the first time you use it (which creates and stores a session) and then for when you are using the session?

Use something like `aws-vault --debug exec <profile> -- aws s3 ls` to test.

stevencarpenter:
One time for a single run of the command you provided, and then if I run it again right after I have to authenticate again. Occasionally I will auth once and the keychain password will be asked for, then I will be asked to auth again in the same command, but this time it will have an Always Accept option. Despite clicking that, I will be asked to re-auth again in like 5 minutes.

lox:
I believe the default timeout of the keychain vault we create is 5 minutes. So you should be re-prompted for your Keychain password every five minutes.

lox:
[...]

PR #318, opened 2018-12-04
 Default to trusting sessions, but not root credentials to address #219

StevenACoffman:
Address #219 as follow up to 99designs/keyring#16

 Default to trusting sessions, but not root credentials. This reduces the need to re-enter password and click "Allow" (or "Always Allow").

Users would not be prompted for passwords for session access/modification, but they would still need to be prompted when the root credentials were accessed. This change in behavior is likely welcome by all, but might warrant a release note in the next release.

### DIFF
--- a/vault/sessions.go
+++ b/vault/sessions.go
@@ -139,7 +139,7 @@ func (s *KeyringSessions) Store(profile string, session sts.Credentials, expires
 		Data:        bytes,
 
 		// specific Keychain settings
-		KeychainNotTrustApplication: true,
+		KeychainNotTrustApplication: false,
 	})
 }
 
--- a/cli/global.go
+++ b/cli/global.go
@@ -80,17 +80,18 @@ func ConfigureGlobals(app *kingpin.Application) {
 				allowedBackends = append(allowedBackends, keyring.BackendType(GlobalFlags.Backend))
 			}
 			keyringImpl, err = keyring.Open(keyring.Config{
-				ServiceName:             "aws-vault",
-				AllowedBackends:         allowedBackends,
-				KeychainName:            GlobalFlags.KeychainName,
-				FileDir:                 "~/.awsvault/keys/",
-				FilePasswordFunc:        fileKeyringPassphrasePrompt,
-				PassDir:                 GlobalFlags.PassDir,
-				PassCmd:                 GlobalFlags.PassCmd,
-				PassPrefix:              GlobalFlags.PassPrefix,
-				LibSecretCollectionName: "awsvault",
-				KWalletAppID:            "aws-vault",
-				KWalletFolder:           "aws-vault",
+				ServiceName:              "aws-vault",
+				AllowedBackends:          allowedBackends,
+				KeychainName:             GlobalFlags.KeychainName,
+				FileDir:                  "~/.awsvault/keys/",
+				FilePasswordFunc:         fileKeyringPassphrasePrompt,
+				PassDir:                  GlobalFlags.PassDir,
+				PassCmd:                  GlobalFlags.PassCmd,
+				PassPrefix:               GlobalFlags.PassPrefix,
+				LibSecretCollectionName:  "awsvault",
+				KWalletAppID:             "aws-vault",
+				KWalletFolder:            "aws-vault",
+				KeychainTrustApplication: true,
 			})
 			if err != nil {
 				return err
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -384,6 +384,9 @@ func (p *KeyringProvider) Store(val credentials.Value) error {
 		Key:   p.Profile,
 		Label: fmt.Sprintf("aws-vault (%s)", p.Profile),
 		Data:  bytes,
+
+		// specific Keychain settings
+		KeychainNotTrustApplication: true,
 	})
 }
 


StevenACoffman:
So. Anyone have a chance to look at this?

lox:
Apologies for the delay @StevenACoffman. My preference for this would be as per my comment at https://github.com/99designs/aws-vault/issues/219#issuecomment-438496156, which is to default to trusting sessions, but not root credentials.

StevenACoffman:
@lox It seems to me you just want to have the root credentials keyring item stored with `KeychainNotTrustApplication: true,` while having the session storage `KeychainNotTrustApplication: false`
Do you really need it to be configurable? 
https://github.com/99designs/aws-vault/blob/master/vault/sessions.go#L142
https://github.com/99designs/aws-vault/blob/master/vault/provider.go#L386

lox:
This looks great @StevenACoffman! Really appreciate the clear and detailed explanation! I think this is super close.

My main concern here is that it requires users to understand how this works and what the config means. I'm kind of tempted to just make this the default and ditch the config option.

What do you think?

StevenACoffman:
I would *love* to ditch the config and make this the default behavior. I won't be able to get to it for a bit, unfortunately.

vault/sessions.go
```diff
@@ -139,7 +139,7 @@ func (s *KeyringSessions) Store(profile string, session sts.Credentials, expires
 		Data:        bytes,
 
 		// specific Keychain settings
-		KeychainNotTrustApplication: true,
+		KeychainNotTrustApplication: false,
```
StevenACoffman: While boolean default value is false, so I believe this line could be omitted, I have left it to be explicit. I am open to removal.

Also, I haven't tested removing it. :)

StevenACoffman:
@lox I altered the code as you suggested, which simplifies things and also avoids needing to document another option. 

After applying, Users would not be prompted for passwords for session access/modification, but they would *still* need to be prompted when the root credentials were accessed. This change in behavior is likely welcome by all, but might warrant a release note in the next release.

+ The `keyring.Open` now has `keyring.Config` option of `KeychainTrustApplication: true`, which allows some items to trust the application.
+ The root credentials keyring item will be stored with `KeychainNotTrustApplication: true,` **requiring a password** to access.
+ The session keyring item will be stored with `KeychainNotTrustApplication: false` so **no password** will be required to access.

+ See here for current master session storage:
https://github.com/99designs/aws-vault/blob/master/vault/sessions.go#L142
+ See here for current master root credential storage:
https://github.com/99designs/aws-vault/blob/master/vault/provider.go#L386
+ See here for current master item trust evaluation:
https://github.com/99designs/keyring/blob/master/keychain.go#L107

lox:
@mtibben and @pda any strong objections? I think this is a pragmatic compromise.

StevenACoffman:
@mtibben @pda I appreciate your work on this project, and it's entirely acceptable if you are not able to do so, but if you do have a minute to look this PR over, I would really appreciate any feedback.

lox:
Let's just ship it. We'll need @mtibben's help for a release though.

lox:
Thanks for your work @StevenACoffman, and apologies we've been sluggish on feedback.

StevenACoffman:
No worries at all. Thanks for your work on a wonderful tool.  By the way, I found the directions for self-signing to be missing a few steps (or perhaps outdated?), so this is what I provided to my co-workers:
#### Install Golang (skip if already done)
If you don't have golang installed, you need to do that and export GOPATH 
```
if [[ ! $(go version 2>/dev/null) ]]
then
  brew install go --cross-compile-common
fi

export GOPATH=${GOPATH:-$HOME/go}
mkdir -p "$GOPATH"

LINE=$(cat << EOF
export GOPATH=$GOPATH
EOF
)

FILE=$HOME/.bash_profile
# Only append LINE if file does not already contain it
grep -qF "$LINE" "$FILE" || echo "$LINE" >> "$FILE"
```

#### Deal with signing aws-vault

1. Start keychain access
2. Keychain access -> Certificate assistance -> Create a certificate
3. Name it whatever you like (I choose AWSVAULTSIGNER)
4. Check "Let me override defaults"
5. Just click continue until you reach "Extended Key Usage Extension" select "Code Signing"
6. Just click continue until you reach the point to select Keychain, select the *System* keychain
7. type `codesign --force --sign name-of-certificate /Users/${HOME}/go/src/github.com/99designs/aws-vault/aws-vault`  in your console
8. You can then sign your binary like this:

```bash
go get -u -v github.com/99designs/aws-vault
cd $GOPATH/src/github.com/99designs/aws-vault
make build
codesign --force --sign name-of-certificate ${GOPATH}/src/github.com/99designs/aws-vault/aws-vault
export PATH="$(PWD):$PATH"
# verify signature
codesign -dvv $(which aws-vault) 2>&1 | grep Authority
```

### Verify signature
    $ codesign -dvv $(which aws-vault) 2>&1 | grep Authority
Check out Apple's guide on it [here](http://web.archive.org/web/20090119080759/http://developer.apple.com/documentation/Security/Conceptual/CodeSigningGuide/Procedures/chapter_3_section_2.html), or find it in `Keychain Access > Certificate Assistant > Create Certificate > Code Signing Certificate`.

Issue #321, opened 2019-01-08
detect presence of "pass" executable


kevinburkenotion:
When I run commands, I'm getting:

```
aws-vault: error: Failed to get credentials for myprofile: exec: "pass": executable file not found in $PATH
```

I believe this is trying to use the pass functionality, but I don't have that set up, I use the keychain. It would be good if aws-vault could detect whether the command exists and try a different backend if so.

(Oddly, if I run it again, a second later, it uses the keychain properly. I don't understand why that is happening, either)

I'm running aws-vault tip with Go tip on a Mac, High Sierra.

mikes-blend:
I'm having a similar issue where I've had an installation script fail because its trying to use pass as the default backend and its not installed on my host. Previously its always selected "file" as the default backend. I was able to work around this by specifying --backend=file on my aws-vault commands in my installation scripts, but just wondering if this was an intentional change?

lox:
Sorry about this folks! @rekahsoft any ideas?

mikes-blend:
Thanks for the quick action guys!

One other question: Lets say you have an aws-vault v4.4.1 installation on a host with pass available on the PATH, and it is currently using a keychain or file backend. Once this change is released, and aws-vault is upgraded on that host, will the the new aws-vault distribution be able to find your existing keys? Or will you need to take explicit action such as specifying --backend=file in your aws-vault exec commands or setting the AWS_VAULT_BACKEND env var?

rekahsoft:
@mikes-blend you would need to explicity set `--backend=file` or the environment variable `AWS_VAULT_BACKEND`. Perhaps there is a nicer way to do this. @lox thoughts?

PR #323, opened 2019-01-10
Update govendor - Ensure passcmd is on the callers PATH

rekahsoft:
Update govendor with resolution to issue https://github.com/99designs/aws-vault/issues/321 where the pass backend was being selecting instead of file, and then failing because the pass executable was not available. 

See: https://github.com/99designs/keyring/pull/34.


rekahsoft:
@lox this is ready to go.

PR #340, opened 2019-02-15
Linux & macOs installation instructions via curl

adiian:
Added instructions to install latest aws-vault release from github, using curl in command line.


weitzman:
These directions worked for me both OSx and Linux. Thanks.

mtibben:
Thanks, but I think we'll leave the installation instructions targeting common package managers

PR #349, opened 2019-03-19
improve documentation on --no-session

gbataille:


gbataille:
Despite my previous PR to document `--no-session`, I realized that my understanding was still partial and that my documentation was lacking. Here is an attempt to improve it.

USAGE.md
```diff
@@ -125,76 +145,132 @@ For the `aws-vault login` subcommand:
...
+   process that will immitate the [metadata
+endpoint](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) that you
+would have on an EC2 instance. When your application will want to connect to AWS and fail to find
+credentials (typically in env variables), it will instead contact this server that will issue a new
+set of temporary credentials (using the same profile as the one the server was started with). This
+server will work only for the duration of the session (`--session-ttl`).  
+Note that this approach has the **major drawback** that while this `aws-vault` server runs, any
+application wanting to **connect** to AWS will be able to do so **implicitely**, with the profile the
+server was started with. Thanks to `aws-vault`, the credentials are not exposed, but the ability to
+use them to connect to AWS is!
```
gbataille: I'll let you confirm this. I think indeed aws-vault run in server mode will work for the entire duration of the session, thus solving the issue but I'm not 100% sure.
I have tried though to use it after it had been up for more than 1h and it seemed to work, so I assume this is ok

mtibben: Yes this is correct AFAIK

mtibben:
Thank you @gbataille !!

PR #316, opened 2018-11-29
Feature/add yubikey

j0hnsmith:
Adds Yubikey integration, #230.

PR #373, opened 2019-05-22
Make the MFA prompt configurable and add ykman support

viraptor:
When using a hardware key capable of totp, it's useful to configure it permanently. It's annoying to specify the command in a parameter and having different parameters for login and exec just adds to that.

This PR adds a configurable MFA prompt and passes the `mfa_serial` to the handler. This allows to use (for example) `ykman` to automatically retrieve the correct code.

### DIFF
--- a/prompt/ykman.go
+++ b/prompt/ykman.go
@@ -2,15 +2,13 @@ package prompt
 
 import (
 	"os/exec"
-	"strings"
 )
 
 func YkmanPrompt(_prompt string, mfa_serial string) (string, error) {
-	cmd := exec.Command("ykman", "oath", "code", mfa_serial)
+	cmd := exec.Command("ykman", "oath", "code", "-s", mfa_serial)
 	out, err := cmd.Output()
 	if err != nil {
 		return "", err
 	}
-	parts := strings.Split(string(out), " ")
-	return parts[len(parts)-1], nil
+	return string(out[:len(out)-1]), nil
 }
--- a/prompt/ykman.go
+++ b/prompt/ykman.go
@@ -1,0 +1,16 @@
+package prompt
+
+import (
+	"os/exec"
+	"strings"
+)
+
+func YkmanPrompt(_prompt string, profile string) (string, error) {
+	cmd := exec.Command("ykman", "oath", "code", profile)
+	out, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	parts := strings.Split(string(out), " ")
+	return parts[len(parts)-1], nil
+}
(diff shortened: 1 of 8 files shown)


prompt/ykman.go
```diff
@@ -0,0 +1,16 @@
+package prompt
+
+import (
+	"os/exec"
+	"strings"
+)
+
+func YkmanPrompt(_prompt string, mfa_serial string) (string, error) {
```
lox: `_prompt`?

lox:
This seems pretty reasonable, could you show an example of what output would look like for the new ykman prompt in action?

prompt/ykman.go
viraptor: I believe that's the standard for "unused, but I wanted to say what it is anyway" naming? Happy to change it to just 
`_` if you prefer.

viraptor:
I fixed up a potential issue with multiple mfa-s in the last commit.
As for the output... there's nothing to show really :( There's no output from `ykman` when it succeeds, so we just get the standard exec command output.

When it fails (unplugged key), I get:
```
2019/05/26 22:18:22 Looking up keyring for vault-test
2019/05/26 22:18:22 [keyring] Expanded file dir to /home/viraptor/.awsvault/keys/
2019/05/26 22:18:22 [keyring] Expanded file dir to /home/viraptor/.awsvault/keys/
Enter passphrase to unlock /home/viraptor/.awsvault/keys/: 
aws-vault: error: Failed to get credentials for vault-test: exit status 2
```
Because of the raw forwarded error... this could be improved a bit.

viraptor:
I tried to get the output and unfortunately due to `ykman` always returning help it looks like this:
```
aws-vault: error: Failed to get credentials for vault-test: output: Usage: ykman [OPTIONS] COMMAND [ARGS]...
Try "ykman -h" for help.

Error: No YubiKey detected!
; exit code: 2
```
I could potentially parse the `Error` line out, but that's getting a bit fiddly.

lox:
How does this related to https://github.com/99designs/aws-vault/pull/316?

viraptor:
Those 2 offer completely different solutions. The PR you referenced interacts with yubikey mfa that AWS knows about in its configuration. This PR deals with an interface for automating the generic oath mfa which happens to be implemented here by querying a yubikey app.

These solutions exist independently.

lox:
Ok cool, I will try and find some time to get my head across the differences and review accordingly.

mtibben:
I think I prefer this approach to #316 (now #392), though as I don't have a Yubikey maybe I'm missing something

I'm wondering if we should make this even more generic - have a configurable command for any particular MFA rather than even hardcoding the command here? That may also address the use-case in #386

mtibben:
Closed in favour of #514

PR #385, opened 2019-06-19
Fix login flag behavior to match that of identical flags on exec

rossmckelvie:
- `--assume-role-ttl` on exec honored the environment variable `AWS_ASSUME_ROLE_TTL`. This has been fixed for login command
- `--profile` on exec has a hint if not passed in, added this to login command


mtibben:
Thanks @rossmckelvie

PR #390, opened 2019-07-02
Document that aws-vault is available via Homebrew on Linux

issyl0:
- There's now a formula for `aws-vault` in Homebrew on Linux
  (https://github.com/Linuxbrew/homebrew-extra/blob/master/Formula/aws-vault.rb),
  because Casks don't work on Linux. This change shows that users can
  `brew tap linuxbrew/extra` and `brew install aws-vault`, or `brew
  install linuxbrew/extra/aws-vault` (the documentation I've gone with).


mtibben:
Thanks @issyl0

Issue #230, opened 2018-02-28
Yubikey Integration


eriksw:
It'd be nice if aws-vault could automatically generate TOTP codes using a Yubikey.

I don't expect this to be an easy undertaking, but I think it'd be a very nice quality of life improvement if prompts for MFA codes were replaced by "Touch your Yubikey to generate an MFA code." if the TOTP secret was stored with the option to require touch, or no prompt at all if not. This would be activated by an additional setting to configure alongside `mfa_serial`, to designate which TOTP secret in the Yubikey to use for a given profile (and maybe also which Yubikey).

What I would like to know is: If I (or anyone else—I'm _not_ promising I'll be able to put time into this right away) succeeded in implementing this, is this a feature that you'd be willing to merge? (Assuming adequate code quality and lack of license issues, of course.)

Yubico's own "Yubico Authenticator" app and the relevant (C) dependencies seems to be under 2-clause BSD in various Yubico repos on Github, although my inclination would be to instead use it as a roadmap to reimplement the necessary bits in Go.

lox:
Interesting idea, it's certainly something I'd consider. I personally use 1password for TOTP codes, and I've considered a similar integration for it.

Looks like https://github.com/jessfraz/libyubikey would make it viable.

lox:
I'd want to to verify first that it was possible to request an MFA code from a Yubikey via the API.

ngerstle:
If you add a otp seed to the yubikey and have `yubioath` installed, it's possible to run and get a token.
``` $ yubioath show aws | awk '{print $NF}'
Touch your YubiKey...
123456
```
I've started using this with aws-vault as follows:
```
$ aws-vault exec profilename -m `yubioath show aws | awk '{print $NF}'` --  command
$ aws-vault login profilename -t `yubioath show aws | awk '{print $NF}'`
```
It's a little irritating as the flags are different (-m for exec and -t for login), and aws-vault login has only been giving me the aws sign on page in the browser of late, but it's a temporary measure that works.

yawn:
Recently released https://github.com/yawn/ykoath - might fit your use case, if you can live with the `libpcsclite` dependency.

blimmer:
It looks like [YubiKey is now officially supported by AWS](https://aws.amazon.com/blogs/security/use-yubikey-security-key-sign-into-aws-management-console/).

tomelliff:
[...]

PR #392, opened 2019-07-05
Add support for Yubikey

mtibben:
Adds Yubikey integration, #230.

Yubikey support was originally merged in #316 but was rolled back as the introduction of CGO prevents cross-compilation. This PR will attempt to address that issue

/cc @j0hnsmith


mtibben:
@j0hnsmith the use of the github.com/ebfe/scard cgo dependency is problematic. Can we reimplement using a native go client?

mtibben:
Seem to be a few native clients here https://godoc.org/?q=pcsc

mtibben:
See here for the ykoath PR https://github.com/yawn/ykoath/pull/9

mtibben:
Could someone with a Yubikey please test this branch?

j0hnsmith:
All the pcsc libraries are wrappers around the C api. 

I think it's going to come down to either figuring out how to build with `CGO=1` (maybe windows and linux on travis, osx on whatever machine you use now as you need the cert right?) or don't merge this feature. Maybe it could be kept in a branch that people have to build themselves if they want it?

jonhadfield:
I've tried with https://github.com/99designs/aws-vault/commit/d1662449236ba155cbaca40cbc3640bb9bef37fc but enabled CGO in the Makefile as it otherwise drops support for the keychain backend. 
After successfully adding the profile using aws-vault, this was my attempt at adding a yubikey:
```
$ ./aws-vault add-yubikey test-aws-vault-yubikey test-aws-vault-yubikey --touch
Adding yubikey to user test-aws-vault-yubikey using profile test-aws-vault-yubikey
aws-vault: error: error registering yubikey
%!(EXTRA *errors.withStack=failed to initialize Yubikey: failed to establish context: dial unix: missing address)
```

j0hnsmith:
@jonhadfield this branch doesn't work, this should work though https://github.com/j0hnsmith/aws-vault/tree/feature/add_yubikey

xocasdashdash:
I've created #426 as a workaround this issue using a completely different way of working with a yubikey. I'd love if people could check it out and see if there was something to improve

Klohto:
Is there any way I could help with this? :angel: 
The workaround is nice, but I would welcome a fully implemented solution

j0hnsmith:
I'm going to fork this repo, build with this functionality and provide a mac only release, I don't have a certificate from apple so there will be a warning upon install. I'll also provide instructions how to build yourself so you don't have to trust that I haven't added anything dodgy.

mtibben:
If the functionality is mac-only then a PR will be welcome. The blocker here was that I don't have a Yubikey and couldn't get cross compilation working

j0hnsmith:
My fork won't solve the cross compilation problem, it will use CGO as before, I'm going to release for mac only initially as that's the machine I use. I might try a linux build at a later stage. 

I've rebased against master so shouldn't be too long.

j0hnsmith:
I was hoping to provide a binary but can't since I don't have an apple developer account therefore keychain will generate too many prompts for it to be practical. If anyone has a way, do let me know.

I've rebased against master and added a little documentation, I'll try to keep the branch up to date. This seems to be the best I can do for now https://github.com/j0hnsmith/aws-vault#aws-vault.

j0hnsmith:
@mtibben I've got it working with build tags 🎉 , Yubikey commands are included only in `darwin` builds (still requires `CGO` though). Is that something that could work with the github workflow build and the release build process?

mtibben:
Closed in favour of #514

PR #400, opened 2019-07-24
Add ecs-server command

xose:
This PR adds a new command to aws-vault: `ecs-server`

The purpose of this command is to start a local credential server daemon that follows the ECS task metadata service format, so other applications can request STS credentials by setting the proper environment variables. Unlike the `--server` option of the `exec` command, this is not meant to start another program with given AWS credentials but to run as a daemon in the background.

The goal is to allow developers to easily assume multiple roles at the same time, by simply adding a new environment variable to their program runs. This is especially useful for micro-service development, as there is a need to assume multiple roles for different services at the same time, while the EC2 metadata service can only provide one role at a time.

## Features:

The command starts an HTTP server on port 9999 (this default can be changed with the `--port` argument). This server provides an endpoint that assumes the requested role(s) and returns the credentials using the ECS task metadata format.

The enpoint is `/creds/{aws-vault profile}[/{optional role ARN}]`

The `{aws-vault profile}` is one of the user's configured profiles in `.aws/config`. It can be a profile with stored credentials or an IAM Role to assume with a source profile, and can optionally have MFA configured. The `{optional role ARN}` is a second role to assume by chaining on the profile's credentials.

### Examples:

`/creds/sandbox`
Gets credentials for the "sandbox" profile. This is equivalent to running "aws-vault exec sandbox"

`/creds/sandbox/arn:aws:iam::1234567890:role/myrole`
Gets credentials for the "sandbox" profile, and then chains to the given role "myrole". This is equivalent to running "aws-vault exec sandbox -- aws sts assume-role --role-arn arn:aws:iam::1234567890:role/myrole"

## User experience:

The following is an example of how this works from the perspective of a user. Let's assume the user has already configured their static IAM user credentials, as well as a "staging" and "sandbox" profiles with MFA that assume a role in two different accounts. The Role in Staging is also allowed to assume roles "ServiceA" for a service written in Java and "ServiceB" for another service written in Python.

1. User runs `aws-vault ecs-server` on a Terminal (or configures it to run as a daemon)
2. User opens his favourite Java and Python IDEs and loads both ServiceA and ServiceB to start working on.
3. User adds the following environment variables to the project settings of both services:
- For ServiceA: `AWS_CONTAINER_CREDENTIALS_FULL_URI=http://localhost:9999/creds/staging/arn:aws:iam::1234567890:role/ServiceA`
- For ServiceB: `AWS_CONTAINER_CREDENTIALS_FULL_URI=http://localhost:9999/creds/staging/arn:aws:iam::1234567890:role/ServiceB`
4. User starts ServiceA for debugging. A prompt comes up asking for his AWS MFA code, that he types. ServiceA is now running with its intended IAM Role permissions.
5. User starts ServiceB for debugging. No MFA prompt because the session has not expired. ServiceB is now running with its intended IAM Role permissions.
6. When the credentials expire after 15 minutes, both services will automatically request new ones and both services will receive their own set of credentials from their Roles.
7. User wants to run some scripts on the Sandbox account. He types `export AWS_CONTAINER_CREDENTIALS_FULL_URI=http://localhost:9999/creds/sandbox` (this can be easily turned into an alias) in the Terminal and can now run any script or aws-cli command directly with his Sandbox account permissions.

## Benefits:

- No root required. It listens on a non-privileged port of localhost, so there is no need to use sudo or change network route settings.
- Runs as a daemon. It can be set up to start on the background during login and will always be available. MFAs are requested as needed, not on startup.
- Assume multiple profiles/roles at the same time. No need to switch profiles or use complicated commands.
- Simple workflow. Just set an environment variable on your project and you're ready to go.
- Least privilege permissions. Allow developers to assume the exact roles their services will be running, and nothing more.
- Role chaining. No need to configure a large amount of profiles for every role to be assumed.

## Security:

- The service will always bind to localhost so it has no access over the network. This is also a limitation enforced by the AWS SDK: this credentials method only works on localhost!
- Any requests from non-loopback addresses will be rejected. Any CORS requests from browsers will be rejected as well.
- There is an optional setting to configure an authentication token. When this option is enabled, the configured token must be sent on every request to the credentials service. The AWS SDKs will send this token when setting the `AWS_CONTAINER_AUTHORIZATION_TOKEN` environment variable.

## Limitations:

When using the option to chain multiple roles there are two limitations to take into account:

- The chained role can have a maximum duration of 1 hour. The `--assume-role-ttl` argument is only used for the first profile, the chained role is always assumed with the default of 15 minutes.
- The chained role must have a Trust Policy that allows access from the source role (or account), even if both roles are already on the same account.

## Usage tips:

- Configure a graphical input method for the MFA token (like `osascript` on the Mac) so you get a popup instead of a prompt on a Terminal you might have lost track of.
- Configure a long session TTL, to avoid continuous MFA prompts.

Here is an example macOS LaunchAgent configuration to run this command as a daemon: https://gist.github.com/xose/03375fd98732259f5992657b297a07b7

### DIFF
--- a/cli/server_ecs.go
+++ b/cli/server_ecs.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/99designs/aws-vault/prompt"
+	"github.com/99designs/aws-vault/server"
 	"github.com/99designs/aws-vault/vault"
 	"github.com/99designs/keyring"
 	"github.com/aws/aws-sdk-go/aws"
@@ -22,9 +23,13 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
-const kEcsCredsPrefix = "/creds/"
+const (
+	kEcsPrivilegedIP = "169.254.170.2"
+	kEcsCredsPrefix  = "/creds/"
+)
 
 type EcsServerCommandInput struct {
+	Privileged   bool
 	ServerPort   int
 	AuthToken    string
 	Duration     time.Duration
@@ -38,7 +43,10 @@ func ConfigureEcsServerCommand(app *kingpin.Application) {
 
 	cmd := app.Command("ecs-server", "Run a local ECS credentials server")
 
-	cmd.Flag("port", "Port to listen on for the credentials server").
+	cmd.Flag("privileged", "Start the server on the privileged IP+port used by ECS (requires superuser)").
+		BoolVar(&input.Privileged)
+
+	cmd.Flag("port", "Port to listen on for the credentials server (ignored in privileged mode)").
 		Default("9999").
 		Short('p').
 		IntVar(&input.ServerPort)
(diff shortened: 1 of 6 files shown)


FernandoMiguel:
> • The service will always bind to localhost
This pr looked awesome until you mentioned it binds localhost only, making it useless when running Docker

xose:
> > • The service will always bind to localhost
>
> This pr looked awesome until you mentioned it binds localhost only, making it useless when running Docker

If you make it available on the `169.254.170.2` IP address for containers ([hint](https://github.com/awslabs/amazon-ecs-local-container-endpoints/blob/master/docs/setup-networking.md#option-1-use-a-user-defined-docker-bridge-network-recommended)) you can use the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` environment variable instead, dropping the host+port, and it will work!

xose:
@FernandoMiguel I have updated the PR to add a `--privileged` argument. When this argument is set the service will listen on the actual IP+port used by ECS, so it works fine from Docker containers. This requires starting aws-vault with sudo and the environment variable has to be changed to the relative one.

Important: macOS requires applications running as root to be codesigned to access the user's Keychain.

ryan-gerstenkorn-sp:
Any blockers on this right now? Let me know if I can help with anything to get this merged.

mtibben:
I can understand the use-case, however I am concerned that starting a general service that gives out AWS credentials is quite a different proposition to the function aws-vault currently fills.

Currently `aws-vault exec --server CMD` will run the command as a child process, and will terminate once the child process exits. It is fairly difficult to misuse this or accidentally leave aws-vault in an accessible state with this design.

The proposed `aws-vault ecs-server` here seems like something that could easily be left in the background accidentally, which is at odds with the aws-vault approach

PR #402, opened 2019-07-31
Bump keyring version

kjenneycloudhealth:


mtibben:
fixed in https://github.com/99designs/aws-vault/releases/tag/v4.6.3

PR #403, opened 2019-07-31
no-role and no-session flags for rotate

pjdufour-truss:
Adds `--no-session` and `--no-role` flags to the `aws-vault rotate` command.

`--no-session` is for overriding the `needsSessionToRotate` heuristic.

`--no-role` overrides the default behavior of assuming the role given in the profile, so a user can can `aws-vault rotate --no-role ...` instead of having to comment out the `role_arn` in their `~/.aws/config` every time they need to rotate.

### DIFF
--- a/cli/rotate.go
+++ b/cli/rotate.go
@@ -14,6 +14,7 @@ type RotateCommandInput struct {
 	Keyring   keyring.Keyring
 	MfaToken  string
 	MfaPrompt prompt.PromptFunc
+	NoRole    bool
 	NoSession bool
 }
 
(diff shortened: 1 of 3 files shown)


FernandoMiguel:
i never needed to comment out the role

pjdufour-truss:
Sometimes you only want users to change their credentials and not allow anyone assuming "engineer" to change anyone else's credentials.  After you assume the role, there is no way for the IAM policy to know the original user.  Therefore, `no-role` is needed so that you rotate your credentials using your user permissions rather than the permissions of an assumed role.

cblkwell:
> i never needed to comment out the role

Could this be because the role you assumed has permissions that allow you to rotate *any* credentials? When I use `aws-vault rotate` this works for me, because the role I'm assuming has AdministratorAccess; an engineer without this permission level runs into a problem.

chrisgilmerproj:
I am also experiencing this problem. Folks without administrator access can't rotate their keys directly using the tool via the assumed role.

rdhariwal:
JFYI I too am impacted by this.

mtibben:
This has been addressed in #450. `--no-session` will rotate without a session. If you don't require a role, ensure your profile config doesn't have a role defined. You can test this out with [v5.0.0-alpha1](https://github.com/99designs/aws-vault/releases/tag/v5.0.0-alpha1)

PR #409, opened 2019-08-17
Added documentation of credential-process with prompt

jychen7:
When I use aws-vault + terraform with multiple profiles, the MFA doesn't prompt and terraform plan fail. It takes times to find out the solution, so I would like to add detail to the document.


jychen7:
When I use aws-vault + terraform with multiple profiles, the MFA doesn't prompt and terraform plan fail. It takes times to find out the solution, so I would like to add detail to the document.

mtibben:
Thanks @jychen7

PR #416, opened 2019-09-04
Add default profile for login command

florisvdg:
Tiny feature to save some extra typing:

`aws-vault exec myprofile` already sets `AWS_VAULT` to `myprofile`, so why not use it when doing `aws-vault login` afterwards in a subshell?

Before PR:
```
$ aws-vault exec myprofile
(in subshell)
$ aws-vault login myprofile
```

After PR:

```
$ aws-vault exec myprofile
(in subshell)
$ aws-vault login
```

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -45,6 +45,7 @@ func ConfigureLoginCommand(app *kingpin.Application) {
 		BoolVar(&input.NoSession)
 
 	cmd.Arg("profile", "Name of the profile").
+		Envar("AWS_VAULT").
 		Required().
 		HintAction(ProfileNames).
 		StringVar(&input.Profile)


mtibben:
hmm should this be `AWS_PROFILE` instead as per [docs](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)?

florisvdg:
Makes sense, but `aws-vault` sets `AWS_VAULT`: 
https://github.com/99designs/aws-vault/blob/843ff8e556cac3bec1d2899a003b0fde3ba82e6b/cli/exec.go#L159

And unsets `AWS_PROFILE`:
https://github.com/99designs/aws-vault/blob/843ff8e556cac3bec1d2899a003b0fde3ba82e6b/cli/exec.go#L165

mtibben:
This "nesting" approach isn't something we want to support

PR #417, opened 2019-09-13
Make password prompts go to stderr

### LINKED ISSUE #405: Passphrase prompt is hidden when piping or redirecting output
README gives the example `aws-vault exec home -- env | grep AWS`, but `aws-vault` outputs a passphrase prompt which is hidden when a pipe or redirect is used.

```Shell
# Shows prompt:
aws-vault exec home -- env
# Hidden prompt:
aws-vault exec home -- env | grep AWS
# Hidden prompt:
aws-vault exec home -- env > env.txt
```

I would suggest a note in README for less experienced shell users, or is there a better work-around?

jonscheiding:
Fixes #405

### DIFF
--- a/cli/global.go
+++ b/cli/global.go
@@ -110,7 +110,7 @@ func fileKeyringPassphrasePrompt(prompt string) (string, error) {
 		return password, nil
 	}
 
-	fmt.Printf("%s: ", prompt)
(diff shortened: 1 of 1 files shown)


mtibben:
Thanks @jonscheiding

jonscheiding:
@mtibben Thanks for merging!  Do you have any estimate of when your next release will be?

PR #421, opened 2019-09-23
Add --no-save-session option for exec/login

### LINKED ISSUE #418: Add option to not store session data in backend
I've been using aws-vault with the pass backend. I recently looked at my listing of all entries in my pass repo and was surprised to see hundreds of aws-vault session entries. I would be nice to be able to use the aws-vault in session mode but not keep the session data in the backend and instead just make it available in the environment. I don't really want every single aws authentication record to be permanently stored in my pass git history.  

It also might be useful to add documentation around why the session data needs to be stored in the backend. I assumed it was to allow it to be shared between different terminals but that doesn't seem to be the case (every time I aws-vault exec I get a new session in my pass backend).

I also noticed that `aws-vault remove -s <profile>` says there aren't any sessions to cleanup. I'm not sure if this an issue with the pass backend or something else weird on my side.

psanford:
This flag allows you to use short lived sessions without saving them
to your keyring.

This is useful (at least) for the password store backend to not save
the history of all sessions in the password store's git repository.

This fixes #418

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -32,6 +32,7 @@ type ExecCommandInput struct {
 	CredentialHelper bool
 	Signals          chan os.Signal
 	NoSession        bool
+	NoSaveSession    bool
 }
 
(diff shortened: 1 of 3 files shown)


mtibben:
Fair enough

PR #422, opened 2019-09-26
Add fish completion

zubin:
## Context

Currently there is fish completion for bash and zsh.

## Change

Add fish completion, which uses `aws-vault --help` to complete the first argument; and then looks at AWS config to complete subsequent ones with IAM roles.

I've been using this for a while successfully.


mtibben:
Thanks @zubin !

PR #426, opened 2019-10-03
Added how to configure aws-vault as a virtual MFA device

### LINKED ISSUE #230: Yubikey Integration

xocasdashdash:
This PR adds some documentation on how to configure a yubikey as a virtual MFA device. 

I believe an integration like #392 is preferable for #230 but this is another way to solve this issue.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -17,6 +17,7 @@
      * [Being able to perform certain sts operations](#being-able-to-perform-certain-sts-operations)
 * [Rotating credentials](#rotating-credentials)
 * [Overriding the aws cli to use aws-vault](#overriding-the-aws-cli-to-use-aws-vault)
+* [Using a yubikey as a virtual MFA](#using-a-yubikey-as-a-virtual-mfa)
 
 ## Getting Help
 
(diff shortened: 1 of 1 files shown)


mtibben:
Thanks @xocasdashdash

PR #429, opened 2019-10-08
added dependencies

shantanuo:
https://stackoverflow.com/questions/58282220/can-not-create-an-image-because-of-failing-go-command


mtibben:
Hey @shantanuo I think this should be resolved by https://github.com/99designs/aws-vault/commit/091d2dc54aeb5d57e855c4df229f2057f5902a66

PR #433, opened 2019-10-25
Report errors via stderr

pbitty:
This fixes an issue where piping the standard output from `aws-vault exec` to another program hides error messages when `aws-vault` itself encounters an error.  For example, when an MFA token is invalid.

Thanks for making such a useful tool!

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ func run(args []string, exit func(int)) {
 		`A vault for securely storing and accessing AWS credentials in development environments.`,
 	)
 
+	app.ErrorWriter(os.Stderr)
 	app.Writer(os.Stdout)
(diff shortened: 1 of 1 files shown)


mtibben:
Thanks @pbitty

arkadiybutermanov:
@pbitty it seems like this doesn't actually fix the issue, because the very next line simply overrides the `errorWriter` back to `os.Stdout`, right?:

https://github.com/99designs/aws-vault/pull/433/files#diff-7ddfb3e035b42cd70649cc33393fe32cR24

https://github.com/alecthomas/kingpin/blob/b6657d9477a694e178dcba813954013bfd2f0346/app.go#L134

PR #434, opened 2019-10-25
Upgrade AWS SDK to latest version

reegnz:
There have been several session-based bugfixes since 0.20.17 that should
be part of aws-sdk now.

Some notable examples include:
https://github.com/aws/aws-sdk-go/releases/tag/v1.21.0
https://github.com/aws/aws-sdk-go/releases/tag/v1.22.0
https://github.com/aws/aws-sdk-go/releases/tag/v1.23.6

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/99designs/keyring v1.1.2
 	github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc // indirect
 	github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf // indirect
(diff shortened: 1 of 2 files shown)


reegnz:
I had an issue with credential_process + mfa_serial + role_arn being set in the config and the AWS_SDK_LOAD_CONFIG=1 environment variable being set. It just couldn't assume the role and didn't even get the session, prompting for MFA each time.

I updated the SDK to the latest version and tried with that -> works like a charm now.

Keeping the SDK up to date is anyway a good thing to do, since AWS constantly fix bugs and add new features to the aws config, and having that supported is important IMHO.

PR #435, opened 2019-10-26
Create notorized dmg for macOS releases

### LINKED ISSUE #430: Not working in Mac OS Catalina
When upgrading to Mac OS Catalina the following error appears as a pop up when using aws-vault

“aws-vault-darwin-amd64” can’t be opened because Apple cannot check it for malicious software.
This software needs to be updated. Contact the developer for more information.

```
➜  ~ uname -a
Darwin MacBook-Pro.local 19.0.0 Darwin Kernel Version 19.0.0: Wed Sep 25 20:18:50 PDT 2019; root:xnu-6153.11.26~2/RELEASE_X86_64 x86_64
```

```
➜  ~ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.15
BuildVersion:   19A583
```

mtibben:
Adds a macOS packaging script to sign the binary, create a dmg and notorize the dmg.

Fixes #430

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,10 +9,9 @@ SRC=$(shell find . -name '*.go')
 all: aws-vault-linux-amd64 aws-vault-darwin-amd64 aws-vault-windows-386.exe aws-vault-freebsd-amd64
 
 clean:
-	rm -f aws-vault-linux-amd64 aws-vault-darwin-amd64 aws-vault-windows-386.exe aws-vault-freebsd-amd64
+	rm -f aws-vault-linux-amd64 aws-vault-darwin-amd64 aws-vault-darwin-amd64.dmg aws-vault-windows-386.exe aws-vault-freebsd-amd64
 
-release: all
-	codesign -s $(CERT) aws-vault-darwin-amd64
+release: all aws-vault-darwin-amd64.dmg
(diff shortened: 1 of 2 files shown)


mtibben:
OK it would be great to get feedback on the notorized dmg, on both Catalina and checking this works on previous macOS releases. Here's what you need to do to test

1. go to https://github.com/99designs/aws-vault/releases
2.  download the dmg
3. open the dmg (there should be no gatekeeper prompts)
4. copy the binary to your system
5. run the binary (there should be no gatekeeper prompts)

reegnz:
This is with v4.7.0-beta1:

The first step (open dmg + copy) seems to work fine without issues.

Running the binary doesn't seem to work without gatekeeper prompt though :(

Screenshot 2019-10-28 at 10 11 02
Screenshot 2019-10-28 at 10 11 35

mtibben:
Well that is very strange, as it seems to be working from the command line. It's just if you try launch it through the GUI that I'm getting the gatekeeper prompt now. @reegnz can you check if you're getting the prompt if you run the binary in your terminal?

reegnz:
That is indeed very strange behaviour from MacOS, not getting the prompt when launching from the cli.
I guess that would be OK then, since it's a cli tool. :) But it's still a very curious behaviour from the OS.

pda:
Nice work @mtibben… I'm just back from Linux on the desktop, and this worked nicely on a fresh macOS Catalina 10.15.5.
(I used Finder to open the DMG and copy to `~/bin`, and of course run aws-vault from command line)

PR #443, opened 2019-11-05
Add support for AWS_MFA_SERIAL env var for rotate

reegnz:
Rotate didn't respect the AWS_MFA_SERIAL environment variable that is
used by the AWS SDK. when mfa_serial is defined in the profile directly,
or even 1 level of source_profile contains mfa_serial it made use of
that.
If however the mfa serial was provided as an environment variable with
AWS_MFA_SERIAL the rotate command failed.

This change allows rotate to pick up AWS_MFA_SERIAL env var and use a session
when the env var is defined.

### DIFF
--- a/cli/rotate.go
+++ b/cli/rotate.go
@@ -13,6 +13,7 @@ type RotateCommandInput struct {
 	Profile   string
 	Keyring   keyring.Keyring
 	MfaToken  string
+	MfaSerial string
 	MfaPrompt prompt.PromptFunc
 }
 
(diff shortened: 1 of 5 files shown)


cli/rotate.go
```diff
@@ -28,6 +29,10 @@ func ConfigureRotateCommand(app *kingpin.Application) {
 		Short('t').
 		StringVar(&input.MfaToken)
 
+	cmd.Flag("mfa-serial-override", "Override the MFA Serial defined in AWS Profile").
```
mtibben: The "override" is redundant, as CLI flags always override other config. I think `mfa-serial` for the flag and something like `The identification number of the MFA device to use`

mtibben:
Should this be added to `exec` and `login` as well for consistency?

mtibben:
Oh right, this is already in `exec`

rossmckelvie:
There is a section "Using MFA" on the README, any reason we do not want to have this documentation there as well? Or point README to USAGE?

PR #445, opened 2019-11-06
Remove profile chaining for MFA to restore aws-cli compatibility

### LINKED ISSUE #438: Restore compatibility with aws-cli
#369 allowed aws-vault to re-use credentials between profiles, which came at the expense of aws-cli compatibility. This went against the approach in #164 and #301 where compatibility with aws-cli was the first priority.

So aws-vault now recognises `mfa_serial` in a `source_profile` while aws-cli does not.

This has been the cause of a number of subtle bugs I've experienced, originating from a mis-understanding of how aws-cli uses credentials and config. Many of our profiles use the same credentials but individual roles may specify a MFA device. This means that if a non-MFA role is used first the session is cached and trying to use roles with MFA will subsequently fail.

While I could now work around this by creating different source profiles, this is not an issue with aws-cli and I feel this divergence is more confusing than it's worth, and we should restore compatibility with aws-cli. 

I'm pretty sure we can have both compatibility with the aws-cli and keep the session chaining. This can be achieved by changing our session caching strategy to use the mfa metadata, and look up sessions based on the mfa device being used.

Thoughts @lox @jstewmon @rossmckelvie @FernandoMiguel

mtibben:
We now have a way to [globally set an MFA device with `AWS_MFA_SERIAL`](https://github.com/99designs/aws-vault/pull/443), so I believe we can now remove the `mfa_serial` chaining behaviour, restoring compatibility with aws-cli config

Fixes #438

### DIFF
--- a/vault/provider.go
+++ b/vault/provider.go
@@ -421,13 +421,10 @@ type VaultCredentials struct {
 }
 
 func NewVaultCredentials(k keyring.Keyring, profile string, opts VaultOptions) (*VaultCredentials, error) {
-	// always get the list of profiles for cycle detection
-	profiles, err := profileChain(profile, opts.Config)
-	if err != nil {
-		return nil, err
-	}
-	if len(opts.MfaSerial) == 0 {
-		opts.MfaSerial = findMfaSerial(profiles)
+	if opts.MfaSerial == "" {
+		if configProfile, exists := opts.Config.Profile(profile); exists {
+			opts.MfaSerial = configProfile.MFASerial
+		}
 	}
 	provider, err := NewVaultProvider(k, profile, opts)
 	if err != nil {
@@ -439,26 +436,4 @@ func NewVaultCredentials(k keyring.Keyring, profile string, opts VaultOptions) (
 
 func (v *VaultCredentials) Expires() time.Time {
 	return v.provider.expires
-}
-
-func findMfaSerial(profiles []Profile) string {
-	for _, profile := range profiles {
-		if len(profile.MFASerial) > 0 {
-			return profile.MFASerial
-		}
-	}
-	return ""
-}
-
-func profileChain(profile string, config *Config) ([]Profile, error) {
-	visited := map[string]bool{}
-	var profiles []Profile
-	for configProfile, exists := config.Profile(profile); exists; configProfile, exists = config.Profile(configProfile.SourceProfile) {
-		if _, ok := visited[configProfile.Name]; ok {
-			return nil, fmt.Errorf("source profile cycle detected for profile %v", profile)
-		}
-		visited[configProfile.Name] = true
-		profiles = append(profiles, configProfile)
-	}
-	return profiles, nil
 }


jstewmon:
I can see that you're very eager to change this 😉 

But, using the `AWS_MFA_SERIAL` environment variable doesn't result in compatibility with the aws cli - the aws cli doesn't recognize that environment variable. Further, the current capability supports multiple, discrete master credentials and mfa serial combinations, which the single environment variable does not.

I haven't had a chance to test, but I also think patch breaks the ergonomic benefits of only having to enter an mfa token once per master credentials and mfa serial combination for users who specify the mfa serial on each profile with a role_arn.

I think we can figure out the best approach together by articulating the desired behaviors in the discussion on #438.

mtibben:
Yes, I am eager - one of our devs ran into this issue again this morning trying to rotate an MFA device. It meant we needed to modify the config file for a particular operation to succeed

jstewmon:
What workflow do you use to rotate the user’s mfa device? I’m not sure what the actual vs expected behavior is for your use case.

mtibben:
We use sudo-like behaviour - read operations always succeed, write operations require MFA

```ini
[profile myprofile]
region=us-east-1

[profile prod]
region=us-east-1
role_arn=arn:aws:iam::11111111111:role/ReadOnly
source_profile=myprofile

[profile prod-admin]
region=us-east-1
role_arn=arn:aws:iam::11111111111:role/Administrator
mfa_serial=arn:aws:iam::2222222222:mfa/user.name
source_profile=myprofile
```

mtibben:
> the current capability supports multiple, discrete master credentials and mfa serial combinations, which the single environment variable does not

True, but you can still achieve this by specifying `mfa_serial` on individual profiles. It doesn't break it

> I haven't had a chance to test, but I also think patch breaks the ergonomic benefits of only having to enter an mfa token once per master credentials and mfa serial combination for users who specify the mfa serial on each profile with a role_arn.

It doesn't break this behavioiur - please do test this yourself to confirm.

jstewmon:
I see - the mfa serial is still passed up the chain, but the discovery is lazy instead of eager.

I was more curious about the specific problem you mentioned earlier:
> one of our devs ran into this issue again this morning trying to rotate an MFA device. It meant we needed to modify the config file for a particular operation to succeed

It isn't clear to me what happened and how it relates to this issue.

I'm also still trying to understand exactly what you expect the behavior to be in all the cases this affects.

It sounds to me like you have some users who are setting mfa_serial on their master credentials, and you don't want them to do that because it changes or breaks some of your workflows, and your solution to that problem is to remove support for that specific config setup. Am I understanding that part correctly?

The part I'm struggling with is that if you don't want to use mfa_serial on the master credentials, you don't have to. Is there some unexpected behavior that's happening when you're not using mfa_serial on the master credentials?

Considering the case I mentioned on #438 regarding rotating master credentials using a session with mfa, we would probably be fine handling that using an environment variable, since most of our users only routinely use a single master credential and mfa combination.

But, I think using the environment variable could be problematic for users who have more than one master credentials and mfa combination. The environment variable is an override for the config file, so in a given shell, you can only work with one set of master credentials if you have the environment variable set.

I wish we had more tests to document what the defined expectations are for these scenarios b/c there are so many that I find it very difficult to understand all the effects these changes have, and whether those changes break expected behaviors.

I don't want to oppose this just for hypothetical / academic reasons, so I'll build this branch and try it locally tomorrow to verify my expectations.

mtibben:
Hey @jstewmon, as a concrete example with the [above config](https://github.com/99designs/aws-vault/pull/445#issuecomment-550121622)

```
aws-vault exec myprofile -- aws iam delete-login-profile --user-name=user.name

An error occurred (InvalidClientTokenId) when calling the DeleteLoginProfile operation: The security token included in the request is invalid
```

Adding `mfa_serial=arn:aws:iam::2222222222:mfa/user.name` to `myprofile` works, but then any call to profile `prod` requires an MFA device, which is not the desired behaviour.

This sudo-like behaviour was always the intent of aws-vault (as well as [similar tools](https://github.com/99designs/aws-vault#references-and-inspiration))

FernandoMiguel:
Reverting this will break the current configuration for all devs at my previous job, and everyone I've been telling about aws-vault since this came out.
It's a massive breaking change, that is going out as a regular release

Although I do like the environment variable solution, I haven't even had the time to test it.

And as I said before, sudo like behaviour of having a read only role, to me feels bad practice!
All roles should have MFA

mtibben:
> Reverting this will break the current configuration

Hey @FernandoMiguel, I totally agree we should avoid breaking changes, but #369 was the breaking change - and now there is no configuration that can put in place that can provide the desired and expected behaviour. The [v4.6.0](https://github.com/99designs/aws-vault/releases/tag/v4.6.0) release at the time didn't explain that or flag it as a breaking change, because it wasn't understood.

Reverting the cascading config restores the original behaviour (of just a few months ago) and it restores compatibility with aws-cli config. We also now have more config options for mfa - `mfa_serial` in profiles, a `AWS_MFA_SERIAL` env var or `--mfa-serial` command line flags. And if we do merge this, we should ensure it's explained in the release notes.

If a cascading `mfa_serial` config really is a desired behaviour but is a breaking change, we should target it for v5 of aws-vault. Or alternatively we should find a way to get the desired behaviour without breaking existing configs. However I do think we already have these options now with `AWS_MFA_CONFIG`.

> Although I do like the environment variable solution, I haven't even had the time to test it.

I'm happy to give everyone time to test and be satisfied that their config can be achieved with these options

> And as I said before, sudo like behaviour of having a read only role, to me feels bad practice!
All roles should have MFA

Whether that's true or not, aws-vault shouldn't dictate how roles and mfa are set up

rossmckelvie:
I am in the same boat as @FernandoMiguel and @jstewmon. I have multiple master profiles and using the env variable doesn't allow me to use aws-vault in the same session. Obviously I could change the ENV var, but for reference here is my config:

```
[profile master1]
mfa_serial=arn:aws:iam::111111111111:mfa/ross.mckelvie
region=us-east-1

[profile assume1]
role_arn=arn:aws:iam::121212121212:role/Staff
source_profile=master1
region=us-east-1

[profile master2]
region=us-east-1
mfa_serial=arn:aws:iam::222222222222:mfa/ross.mckelvie

[profile assume2]
region=us-east-1
source_profile=master2
role_arn=arn:aws:iam::232323232323:role/OrganizationAccountAccessRole
```

I built this branch, in my shell environment exported `AWS_MFA_SERIAL` set to `arn:aws:iam::111111111111:mfa/ross.mckelvie`.

Any profile that references a source_profile with mfa device matching the ENV variable works as expected:

```
./aws-vault-darwin-amd64 exec assume1 -- aws sts get-caller-identity
Enter token for arn:aws:iam::111111111111:mfa/ross.mckelvie: 123456
{
    "UserId": "AROAWY4WODJ8CUGEC6FIQ:1573049545886142000",
    "Account": "121212121212",
    "Arn": "arn:aws:sts::121212121212:assumed-role/Staff/1573049545886147000"
}
```

However, any profiles referencing master2 do not work with the ENV var set:
```
./aws-vault-darwin-amd64 exec assume2 -- aws sts get-caller-identity
Enter token for arn:aws:iam::111111111111:mfa/ross.mckelvie: 123456
aws-vault: error: Failed to get credentials for master2 (source profile for assume2): AccessDenied: MultiFactorAuthentication failed, unable to validate MFA code.  Please verify your MFA serial number is valid and associated with this user.
	status code: 403, request id: fa653cb1-009f-11ea-bb90-6ba8d5d7ab02
```

I _could_ change the ENV variable each time I need to switch into org2, but by reviewing the contents of the `config` file, it's reasonable to assume that profile `assume2` _should_ be using the correct MFA device serial defined in `master2`.

rossmckelvie:
Not using the ENV variable, and using code from this branch, does not break my workflow. The only effect that I can see from this PR is that it removes the cycle protection that results in a crash when a cycle is configured.

Could you maybe provide more information such as example policy that's causing the issue for you @mtibben? I'm having a hard time understanding the exact problem/configuration that this PR is addressing.

jstewmon:
>  but #369 was the breaking change - and now there is no configuration that can put in place that can provide the desired and expected behaviour

#369 did not break anything by design, but a few bugs have been found by users and promptly fixed (e.g. #440 )

The design intent was simply to allow the mfa_serial to be placed on the master credentials profile, so that users could easily create a configuration that will always use mfa with the master credentials.

I understand that is not the behavior that everyone wants, and any affect the feature had on your workflow was unintentional, and considered a bug.

Since #440 was merged, I haven't been able to decipher any other unexpected behaviors from our discussions.

IIUC, the issue you mentioned in [above](https://github.com/99designs/aws-vault/pull/445#issuecomment-550148315) is that the user needs to use the master credentials principal to perform an operation with mfa. You can achieve that in the following ways:
1. add mfa_serial to the master credentials profile and opt into the always-use-mfa behavior
1. run the command with the `mfa-serial-override` option
    1. `aws-vault exec --mfa-serial-override=<serial> myprofile -- aws iam delete-login-profile --user-name=user.name`
    1. `AWS_MFA_SERIAL=<serial> aws-vault exec myprofile -- aws iam delete-login-profile --user-name=user.name`

If we didn't allow option 1, you'd have to use option 2.i or 2.ii the same as you can with option 1 being allowed.

What @rossmckelvie has observed aligns with my understanding of the change in this PR simply changing the mfa serial detection from being eager with cycle detection to lazy without cycle detection.

I think we are all just trying to understand what the remaining problem is that you're trying to fix.

mtibben:
> I have multiple master profiles and using the env variable doesn't allow me to use aws-vault in the same session

Hey @rossmckelvie for your case it looks like you'll need to specify `mfa_serial` in each profile to get your desired behaviour

> Could you maybe provide more information such as example policy that's causing the issue for you

See the examples at https://github.com/99designs/aws-vault/pull/445#issuecomment-550148315 and  https://github.com/99designs/aws-vault/issues/438#issuecomment-547222251

mtibben:
> #369 did not break anything by design

Yes I understand that, that's why no breaking changes were flagged in v4.6.0. However it does change behaviour and made certain configurations impossible to achieve with config files alone, so it is a breaking change.

> IIUC, the issue you mentioned in above is that the user needs to use the master credentials principal to perform an operation with mfa. You can achieve that in the following ways

Hey @jstewmon Option 1 is not the desired behaviour, as we have a read-only role without MFA and an admin role with MFA. Option 2 requires special config for particular commands, and so that is also not desired behaviour

jstewmon:
@mtibben I think we are talking past each other...

I have tested your changes, and the only discernible difference I can find is that it removes cycle detection, which results in a hung process if a cycle is configured.

I can't figure out why you expect this command to work with this patch - I have tested it and it still doesn't work:

```
✗ aws-vault remove -s default
Deleted 1 sessions.
✗aws-vault exec default -- aws iam delete-login-profile --user-name=jstewmon

An error occurred (InvalidClientTokenId) when calling the DeleteLoginProfile operation: The security token included in the request is invalid
```

If a temporary session with mfa present is required to perform that action, how do you expect aws-vault to know to prompt for an mfa token if there is neither an mfa_serial property on the profile nor an AWS_MFA_SERIAL environment variable set?

Can you please provide an example of a config + commands that don't work with the master build but do work with this build, so we can reach a common understanding of what problem you're trying to solve?

mtibben:
Hey folks, there seems to be a lot of strong feelings about this, which I understand.

Given that
 1. #369 was a breaking change which resulted in some configuration not be able to be expressed via config (when an MFA is provided for some roles but not others as given in examples https://github.com/99designs/aws-vault/pull/445#issuecomment-550148315 and  https://github.com/99designs/aws-vault/issues/438#issuecomment-547222251)
 2. All desired behaviour given in the threads in #445 and #438 can still be achieved with a single config file by specifying the `mfa_serial` per profile. This could be optimised further by templating out the config file or using `AWS_MFA_SERIAL`

I'm going to move this forward and merge. Please test this out and let me know if any configuration is not able to be achieved - if it is not I'm happy to continue discussing

rossmckelvie:
I strongly disagree with the rash merge, considering we had an open PR for quite some time before it was merged and released without as much engagement in the discussions. Changes that folks active in this repo would like to maintain. This breaks the configuration in use by over 200 developers and requires your “find and replace” workaround to assign MFA for each role that is specific to the user.

I urge you to try out the configuration/workflow that we and others suggest, as it is the direction AWS themselves are moving and recommending, to see if you can first adopt the workflow, and also notice any improvements.

This is your software so do as you please, but it does seem like you’ve been dismissful of the improvements ever since your rotate stopped working, and not leaving things open long enough for extensive discussion and testing, important for a project without tests.

jstewmon:
@mtibben you seem to be the only participant with “feelings” about this - everyone else tried to take a rational approach to understanding if there was an actual problem you sought to address, but you declined to engage in a productive discussion.

I still don’t have the slightest clue what problem you were encountering or how this patch resolved it.

mtibben:
Hey @rossmckelvie, I'm sorry you feel that way. We really do want to support everyone's workflow in the best way possible. If you have alternate ways to approach this we're open to any suggestions.

> we had an open PR for quite some time before it was merged and released without as much engagement in the discussions

@rossmckelvie there was no mention in that PR that this would break some configurations. If there was, there certainly would have been discussion on it.

However, we do want to keep config working that has been working previously to continue working.

> ever since your rotate stopped working

I'm not sure what you're referring to here - is there an open issue?

mtibben:
> I still don’t have the slightest clue what problem you were encountering or how this patch resolved it.

Hey @jstewmon I explained two scenarios where the config file can no longer express the desired config in https://github.com/99designs/aws-vault/pull/445#issuecomment-550148315 and  https://github.com/99designs/aws-vault/issues/438#issuecomment-547222251. What are you still unclear about?

jstewmon:
I tested the cases you mentioned with and without this patch and got the same results, which I explained above. I asked what you expected to be different, and I asked you to provide the config and commands you used to validate the change in behavior, but you ignored that request.

So, I’ll repeat - the only difference I see from static analysis and practical testing is that cycle detection has been lost.

I have all but begged you to explain what material differences you think this patch has made or to demonstrate any material changes with a test case we can all understand.

mtibben:
@jstewmon I'll try to explain again.

Here is a config scenario that can no longer be expressed in aws config as of v4.6.0

2 roles defined:
 - ReadOnly with `aws:MultiFactorAuthPresent: "false"`
 - Administrator with `aws:MultiFactorAuthPresent: "true"`
- .aws/config looks like this
```ini
[profile jonsmith]
region = us-east-1
mfa_serial = arn:aws:iam::2222222222:mfa/jonsmith

[profile prod-readonly]
region=us-east-1
role_arn = arn:aws:iam::111111111111:role/ReadOnly
source_profile = jonsmith

[profile prod-admin]
region=us-east-1
role_arn = arn:aws:iam::111111111111:role/Administrator
mfa_serial = arn:aws:iam::2222222222:mfa/jonsmith
source_profile = jonsmith
```

- Credentials are stored against `jonsmith`
- However `aws-vault exec prod-readonly` prompts for an MFA, which is not wanted.
- Removing `mfa_serial` from `jonsmith` restores the wanted behaviour, however now commands on `jonsmith` do not succeed

      aws-vault exec jonsmith -- aws iam delete-login-profile --user-name=jonsmith

      An error occurred (InvalidClientTokenId) when calling the DeleteLoginProfile operation: The security token included in the request is invalid


So the desired config cannot be expressed as of v4.6.0. This is not an issue with aws-cli which does not have a cascading `mfa_serial` config

mtibben:
> what material differences you think this patch has made

This patch allows the above config to be expressed again. Currently this config cannot be expressed because the `mfa_serial` is sourced from `source_profile`. It's either all roles need MFA or none, you can't specify roles that require MFA

jstewmon:
I have to assume you did not test that scenario with this patch because I did, and it still failed. I mentioned this in https://github.com/99designs/aws-vault/pull/445#issuecomment-550523540 just before you unilaterally decided to merge this.

It seems like either `delete-login-profile` requires mfa with a session token or is cannot be performed with a session token. Either way, this patch doesn't appear to fix the error.

mtibben:
Yes exactly, `delete-login-profile` requires MFA. `mfa_serial` needs to be set against `jonsmith`.

However with `mfa_serial` set against `jonsmith` I then get.

```
aws-vault exec prod-readonly
Enter token for arn:aws:iam:: 2222222222:mfa/jonsmith:
```

`mfa_serial` is not set on `prod-readonly`, or against the `ReadOnly` role. Prompting for MFA in this scenario is incorrect

jstewmon:
With this patch, what command do you run to provide mfa when you want to call `delete-login-profile`?

rossmckelvie:
> However aws-vault exec prod-readonly prompts for an MFA, which is not wanted.

This isn't native aws behavior by default for a trust policy. From my [comments in the issue thread](https://github.com/99designs/aws-vault/issues/438#issuecomment-547237135), the only way that I'm able to fail to assume a role is when it's been explicitly denied, or the condition requires the mfa present value to be false. The response was that _"the whole point of the read-only role is that MFA is not required. So this is not behaving the way I'd expect."_ 

By default, assume-role operations originating from a session with MFA present will work, whether or not MFA is explicity required by my user policy or the role's trust policy. My understanding of your configuration is that you have a role `ReadOnly` that requires MFA _not_ be used.

Is that correct, and my question from a security perspective if you're not going to require MFA for ReadOnly, why go the additional step to require MFA not be used?

---

> Currently your config is restricted to all roles requiring MFA or none requiring MFA, you can't currently specify individual roles that require MFA

I'm not sure that this is correct, I'm not sure where this would be getting blocked in a default scenario (only way i'm able to reproduce this is with the example linked above).

Here is an example config:

```
[profile corp-iam]
region=us-east-1

[profile sandbox-admin]
role_arn=arn:aws:iam::1111111111111:role/Administrator
source_profile=corp-iam
region=us-east-1
mfa_serial=arn:aws:iam::000000000000:mfa/ross.mckelvie

[profile sandbox-billing]
role_arn=arn:aws:iam::1111111111111:role/Billing
source_profile=corp-iam
region=us-east-1
```

I get the same behavior on the build of this branch as I do in production (v4.7.1). Because I don't have any policies _against_ using MFA, I only tested the configuration where I put the mfa serial on one of my roles:
```
$ aws-vault exec sandbox-admin -- aws sts get-caller-identity
Enter token for arn:aws:iam::000000000000:mfa/ross.mckelvie: 050456
{
    "UserId": "XXXXXXX:YYYYYYYYY",
    "Account": "1111111111111",
    "Arn": "arn:aws:sts::1111111111111:assumed-role/Administrator/1573091144636718000"
}

$aws-vault exec sandbox-billing -- aws sts get-caller-identity
{
    "UserId": "XXXXXXX:YYYYYYYYY",
    "Account": "1111111111111",
    "Arn": "arn:aws:sts::1111111111111:assumed-role/Billing/1573091517783517000"
}

$ aws-vault list
Profile                        Credentials              Sessions
=======                        ===========              ========
default                        -
corp-iam                  corp-iam            1573105545 (mfa), 1573105697
```

mtibben:
> With this patch, what command do you run to provide mfa when you want to call delete-login-profile?

```
aws-vault exec jonsmith -- aws iam delete-login-profile --user-name=jonsmith
```

rossmckelvie:
I have also tested the changes in this PR and verified that cycle detection is indeed removed.

Here is a sample configuration that has different behavior in production from a build of this branch:
```ini
[profile test-admin]
role_arn=arn:aws:iam:: 1111111111111:role/Administrator
source_profile=test-developer
region=us-east-1

[profile test-developer]
role_arn=arn:aws:iam::434270711857:role/Developer
source_profile=test-admin
region=us-east-1
```

In production (v4.7.1):
```
aws-vault exec test-developer -- aws sts get-caller-identity
aws-vault: error: source profile cycle detected for profile test-developer
```

In a build of this branch, my fans spins up and nothing occurs:
```
time ./aws-vault-darwin-amd64 exec test-developer -- aws sts get-caller-identity
^C
./aws-vault-darwin-amd64 exec test-developer -- aws sts get-caller-identity  35.90s user 9.75s system 102% cpu 44.396 total
```

rossmckelvie:
I have also tested the changes in this PR and verified that cycle detection is indeed removed.

Here is a sample configuration that has different behavior in production from a build of this branch:
```ini
[profile test-admin]
role_arn=arn:aws:iam:: 1111111111111:role/Administrator
source_profile=test-developer
region=us-east-1

[profile test-developer]
role_arn=arn:aws:iam::1111111111111:role/Developer
source_profile=test-admin
region=us-east-1
```

In production (v4.7.1):
```
aws-vault exec test-developer -- aws sts get-caller-identity
aws-vault: error: source profile cycle detected for profile test-developer
```

In a build of this branch, my fans spins up and nothing occurs:
```
time ./aws-vault-darwin-amd64 exec test-developer -- aws sts get-caller-identity
^C
./aws-vault-darwin-amd64 exec test-developer -- aws sts get-caller-identity  35.90s user 9.75s system 102% cpu 44.396 total
```

jstewmon:
> > With this patch, what command do you run to provide mfa when you want to call delete-login-profile?
> 
> ```
> aws-vault exec jonsmith -- aws iam delete-login-profile --user-name=jonsmith
> ```

Here's what happens when I run the equivalent. Notice there is no prompt for mfa b/c there is nothing to indicate to aws-vault that mfa is needed when using my `default` profile.

```
➜  aws-vault git:(rm-mfa-serial-chaining) ✗ git log -1
commit 665377840d12097934e2ab20e8cf9a6513c9c504 (HEAD -> rm-mfa-serial-chaining, 99designs/rm-mfa-serial-chaining)
Author: Michael Tibben <michael.tibben@99designs.com>
Date:   Wed Nov 6 12:21:28 2019 +1100

    Remove profile chaining for MFA to restore aws-cli compatibility
➜  aws-vault git:(rm-mfa-serial-chaining) ✗ go install
➜  aws-vault git:(rm-mfa-serial-chaining) ✗ which aws-vault
/Users/jstewmon/go/bin/aws-vault
➜  aws-vault git:(rm-mfa-serial-chaining) ✗ go env GOPATH
/Users/jstewmon/go
➜  aws-vault git:(rm-mfa-serial-chaining) ✗ aws-vault remove -s default
Deleted 1 sessions.
➜  aws-vault git:(rm-mfa-serial-chaining) ✗ aws-vault exec default -- aws iam delete-login-profile --user-name=JonathanStewmon

An error occurred (InvalidClientTokenId) when calling the DeleteLoginProfile operation: The security token included in the request is invalid
➜  aws-vault git:(rm-mfa-serial-chaining) ✗ head -3 ~/.aws/config
[profile default]
region = us-east-1

```

You have to either have an mfa_serial on your master credentials profile or be setting the AWS_MFA_SERIAL environment variable to get a prompt for mfa when using the master credentials profile.

And, if you have either of those, you are going to get a prompt for mfa when you use any profile that refers to the master credentials.

rossmckelvie:
@jstewmon he said that it works in this case, however it's incorrect that MFA is being prompted for read-only.

His acceptance criteria is that when assuming-role for readonly, the user _must not_ be prompted for MFA, or use a session created with MFA due to conditions in the policies requiring MFA-less assume role. In order to meet this criteria, he must remove the mfa_serial from his source profile and place it on all of the other roles.
 
To me this is where the problem lies, he can't create a config that doesn't prompt/allow for MFA on ReadOnly, while always requiring MFA on source profile (to delete login-profile, rotate creds, etc). 

Keeping this configuration in mind that meets the AC for ReadOnly, he would have to adjust workflow to require AWS_MFA_SERIAL env var to be set.

Now that I think about it, why can't you set the env var only in circumstances where you will require MFA to execute commands on your source profile? From the examples given, it seems like MFA is only required for certain operations like assume Admin and modify my own user credentials, but explicitly NOT required in the read only assume role, so there are likely other cases where MFA should not be provided.

mtibben:
> You have to either have an mfa_serial on your master credentials profile or be setting the AWS_MFA_SERIAL environment variable to get a prompt for mfa when using the master credentials profile.
> 
> And, if you have either of those, you are going to get a prompt for mfa when you use any profile that refers to the master credentials.

Yes exactly - this is the problem. Either

1. You specify `mfa_serial` on your master credentials and _all_ profiles using it as `source_profile` now need MFA (not wanted for ReadOnly role)
2. You don't specify `mfa_serial` on your master credentials and a number of commands no longer work, for example the delete-login-profile` example

rossmckelvie:
@mtibben why is MFA not wanted for ReadOnly? The only problem it actually causes that I have been able to find is when the policies dictate that MFA not be used at all.

mtibben:
> why is MFA not wanted for ReadOnly?

To replicate a sudo-like behaviour. This has in fact been the example config in the README since the inception of aws-vault

rossmckelvie:
Perhaps we can introduce an environment variable to disable the MFA, as well as a flag `--ignore-mfa-requirements` to meet this AC.

```grep '\[profile' ~/.aws/config | wc -l
      74```
I've got 74 profiles in my config, all of them require MFA, as does any action against my own iam user such as the operations used in the examples above. aws-vault with @jstewmon additions seemed like the perfect compliment to the security policies put in place with our architecture and we made some changes that improved the ergonomics. Today I find myself here, where we are rolling back *chain detection* because of a problem that only exists when a policy is configured to not allow MFA at all, and otherwise there are additional prompts for users to enter MFA (which maybe they don't have? still trying to figure out the true use case here outside of being in readme since inception).

**Comment by rossmckelvie** [comment]
Perhaps we can introduce an environment variable to disable the MFA, as well as a flag `--ignore-mfa-requirements` to meet this AC.

```
grep '\[profile' ~/.aws/config | wc -l
      74
```
I've got 74 profiles in my config, all of them require MFA, as does any action against my own iam user such as the operations used in the examples above. aws-vault with @jstewmon additions seemed like the perfect compliment to the security policies put in place with our architecture and we made some changes that improved the ergonomics. Today I find myself here, where we are rolling back **chain detection** because of a problem that only exists when a policy is configured to not allow MFA at all, and otherwise there are additional prompts for users to enter MFA (which maybe they don't have? still trying to figure out the true use case here outside of being in readme since inception).

**Comment by mtibben** [comment]
> still trying to figure out the true use case here

To replicate a sudo-like behaviour. @rossmckelvie we need to keep configs working that have worked forever in aws-vault

> we are rolling back chain detection

Yep let's not do that, we should restore if that's an issue

**Comment by rossmckelvie** [comment]
Have you seen my comments in this thread proving this restored a crash when configs are chained?

**Comment by mtibben** [comment]
Yes I did - we should fix that behaviour

**Comment by jstewmon** [comment]
> 1. You specify `mfa_serial` on your master credentials and _all_ profiles using it as `source_profile` now need MFA (not wanted for ReadOnly role)
> 2. You don't specify `mfa_serial` on your master credentials and a number of commands no longer work, for example the `delete-login-profile` example

Is it your impression that this patched changed one of these? It sounds like you think it changed the first point, so that given this config:

```
[default]
mfa_serial=arn:aws:iam::1111111111:mfa/JonathanStewmon

[profile ViewOnly]
source_profile=default
role_arn=arn:aws:iam::1111111111:role/ViewOnly
```

I should not be prompted for mfa when I run:

```
✗ aws-vault exec ViewOnly -- aws sts get-caller-identity
Enter token for arn:aws:iam::1111111111:mfa/JonathanStewmon: 861644
{
    "UserId": "XXXXXXXXXXXX:1573097311323409000",
    "Account": "1111111111",
    "Arn": "arn:aws:sts::1111111111:assumed-role/ViewOnly/1573097311323409000"
}
```

The only way I've found to have mfa_serial on the master credentials profile without getting prompted for mfa is to call a profile section `[profile default]`, which breaks something else because I will never get prompted for mfa when using that specific section label - I think this may be caused by some tertiary bug in `config.go`, but who knows.

I would love it if you could provide some of the config samples and commands you're running to demonstrate the difference you see with this patch because I just can't seem to reproduce what you're I think you're stating as the intended effects.

jstewmon:
@mtibben I would be happy to chat with you over zoom or similar at your convenience to try to straighten this out, since we’ve had so much trouble reaching an understanding in the comments.

FernandoMiguel:
Having to enter the mfa serial for each profile is a nightmare
Not only that, having dozens of roles, each time we need to assume a different one, we need to enter the mfa again, something that chaining from the profile fixed and has been Godsend.

If this behaviour breaks it for you, why can't your config only have MFA on your prod role? Should cause no impact on the read only.
And if your policies state that read only should not use mfa, maybe aws-vault could have a parameter per role to not use mfa too?

AWS_MFA_SERIAL although useful, won't work for me since I have 5 accounts setup, all with MFA (for all roles obviously)
So I would need to store the mfa serials somewhere else and set them up every time I need to use that account.
For single account users, it's easy to put that on bashrc

Please restore the functionality of chaining mfa, it's incredibly powerful

mtibben:
> Not only that, having dozens of roles, each time we need to assume a different one, we need to enter the mfa again, something that chaining from the profile fixed and has been Godsend.

@FernandoMiguel the session caching does NOT change in this PR. You will not have to enter MFA tokens again

> If this behaviour breaks it for you, why can't your config only have MFA on your prod role? Should cause no impact on the read only.

Because doing this means that operations on the master credential profile fail. This is a regression, a break in backwards compatibility.

> AWS_MFA_SERIAL although useful, won't work for me since I have 5 accounts setup, all with MFA (for all roles obviously). So I would need to store the mfa serials somewhere else and set them up every time I need to use that account.

I'm open to suggestions on how to solve this. I'm thinking either via templating the config, or via an expanded env var scheme e.g. `AWS_MFA_SERIAL_<PROFILE>`

mtibben:
I've been thinking about this, and I think I have a reasonable way forward - https://github.com/99designs/aws-vault/issues/448

PR #450, opened 2019-11-12
Add parent_profile config variable

### LINKED ISSUE #448: RFC: introduce `parent_profile` config parameter
#445 removed the profile chaining behaviour for `mfa_device`, as it broke compatibility with older versions of aws-vault, existing configs, aws-cli and the AWS SDK.

I propose that we introduce a new config parameter to the config, `parent_profile`, to reintroduce and extend this chaining behaviour.

Using `parent_profile` would allow a profile to import all of the config and credentials from the specified parent profile. This would include `mfa_serial` but would also extend to other config options like `region` and `duration_seconds `.

This would be a useful addition to aws-vault, as it would give users the ability to build up more complex configurations without the need for templating.

And this means that `source_profile` would retain it's intended semantics of importing credentials only, which means if desired you could retain compatibility with aws-cli config and the SDK. Plus aws-cli config docs would still be applicable to aws-vault

Here's example config to get you thinking about how it might work

```ini
[profile org2]
region=us-east-1
mfa_serial = arn:aws:iam::3333333333:mfa/jonsmith

[profile org2-roleA]
parent_profile=org2
role_arn = arn:aws:iam::44444444444:role/roleA

[profile org2-roleB-short]
parent_profile=org2
role_arn = arn:aws:iam::44444444444:role/roleB
duration_seconds=900

[profile org2-roleB-long]
parent_profile=org2-roleB
duration_seconds=28800

[profile org1]
region=us-east-1

[profile org1-mfa]
parent_profile=org1
mfa_serial = arn:aws:iam::111111111111:mfa/jonsmith

[profile org1-roleA]
parent_profile=org1
role_arn = arn:aws:iam::22222222222:role/ReadOnly

[profile org1-roleB]
parent_profile=org1-mfa
role_arn = arn:aws:iam::22222222222:role/Administrator

[profile org1-roleC]
parent_profile=org1-mfa
role_arn = arn:aws:iam::22222222222:role/Another-Role-Needing-MFA
```

Thoughts @lox @jstewmon @FernandoMiguel @rossmckelvie ?

mtibben:
Fixes #448.

#445 removed the profile chaining behaviour for `mfa_device`, as it broke compatibility.

This PR introduces a `parent_profile` parameter to specify wanted inheritance behaviour. This gives users the ability to build up more complex configurations without the need for templating.

Using `parent_profile` allows a profile to import all of the config and credentials from the specified parent profile. This includes `mfa_serial` but also extends to other config options. `source_profile` retains its original semantics of importing credentials only, which means compatibility with aws cli and the SDK.

It should be noted that aws-cli and the AWS SDK does not recognise `parent_profile` as a variable and will be ignored

In order to achieve this, I needed to refactor parts of the codebase 
- All config handling has been moved to `ConfigLoader`
- `VaultCredentials` has been removed, it is unnecessary
- `VaultProvider` has been simplified
- rotation has been simplified, this also adds `--no-session` to rotate

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -1,22 +1,24 @@
 # Usage
 
-* [Help](#help)
-* [Managing profiles](#managing-profiles)
-    * [Using multiple profiles](#using-multiple-profiles)
-    * [Example ~/.aws/config](#example-~/.aws/config)
-    * [Listing profiles](#listing-profiles)
-    * [Removing profiles](#removing-profiles)
+* [Getting Help](#getting-help)
+* [Config](#config)
 * [Environment variables](#environment-variables)
+* [Managing Profiles](#managing-profiles)
+  * [Using multiple profiles](#using-multiple-profiles)
+  * [Example ~/.aws/config](#example---aws-config)
+  * [Listing profiles](#listing-profiles)
+  * [Removing profiles](#removing-profiles)
 * [Backends](#backends)
+* [MFA](#mfa)
 * [Removing stored sessions](#removing-stored-sessions)
-* [Logging into aws console](#logging-into-aws-console)
+* [Logging into AWS console](#logging-into-aws-console)
 * [Using credential helper](#using-credential-helper)
 * [Not using session credentials](#not-using-session-credentials)
-     * [Considerations](#considerations)
-     * [Assuming a role for more than 1h](#assuming-a-role-for-more-than-1h)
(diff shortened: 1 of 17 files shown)


mtibben:
See [v5.0.0-alpha1](https://github.com/99designs/aws-vault/releases/tag/v5.0.0-alpha1) built from this branch

PR #452, opened 2019-11-15
Add more --server instructions

gatsbys:
We have faced an issue with that change explains and we think is a good point to share it.

We had a machine that for some reason had the environment variables previously set and after run the `aws-vault`in server mode we cannot run any aws command without the `aws-vault`first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,13 @@ The credentials are exposed to the subprocess in one of two ways:
 
 The default is to use environment variables, but you can opt-in to the local instance metadata server with the `--server` flag on the `exec` command.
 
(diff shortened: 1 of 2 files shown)


mtibben:
Perhaps we should display a warning when running `aws-vault exec --server` if AWS env vars have been set

gatsbys:
> Perhaps we should display a warning when running `aws-vault exec --server` if AWS env vars have been set

That feedback should be super helpful yes! But I think that it's compatible with the little tip on the docs, what do you think @mtibben?

mtibben:
I think this may be better in USAGE.md? The README.md is targeted at the high-level concepts without specific usage tips

gatsbys:
I've moved the text to the USAGE file. just after the --server command.

PR #461, opened 2019-12-05
Ignore case of section and key names when parsing config.

jonhadfield:
This PR simply allows for the user to specify upper or lower (or mixed case) section and key names in their config. I raise it as I wasted 3 hours of my life wondering why my aws-vault generated session was invalid before realising mfa_**S**erial=arn:aws:iam::... was not being read.
The aws cli also ignores case. I've proven this by running 'aws configure' then 'aws_access_key_id=AK...' to '**A**ws_access_key_id=AK...' and it still reads the credentials. 

I've modified the test for config to handle this use-case.

There is another related issue to address: a user specifying a token value, but not an MFA serial, is returned an invalid session rather than being warned that there's no point specifying token value if serial isn't supplied. I'll try and get a PR out to handle that.

### DIFF
--- a/vault/config_test.go
+++ b/vault/config_test.go
@@ -28,7 +28,7 @@ region=us-east-1
 [profile withmfa]
 source_profile=user2
 role_arn=arn:aws:iam::4451234513441615400570:role/aws_admin
-mfa_serial=arn:aws:iam::1234513441:mfa/blah
+mfa_Serial=arn:aws:iam::1234513441:mfa/blah
 region=us-east-1
 duration_seconds=1200
 
--- a/vault/config_test.go
+++ b/vault/config_test.go
@@ -25,7 +25,7 @@ output=text
 source_profile=user2
 region=us-east-1
 
-[profile withmfa]
+[profile withMFA]
 source_profile=user2
 role_arn=arn:aws:iam::4451234513441615400570:role/aws_admin
 mfa_Serial=arn:aws:iam::1234513441:mfa/blah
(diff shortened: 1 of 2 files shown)


FernandoMiguel:
Given that mixed case won't work with aws sdk anyway, having aws-vault support that would break backwards compatibility, right?

FernandoMiguel:
And.... I see you edited to include a test with aws sdk... OK

vault/config_test.go
```diff
@@ -28,7 +28,7 @@ region=us-east-1
 [profile withmfa]
 source_profile=user2
 role_arn=arn:aws:iam::4451234513441615400570:role/aws_admin
-mfa_serial=arn:aws:iam::1234513441:mfa/blah
+mfa_Serial=arn:aws:iam::1234513441:mfa/blah
```
mtibben: This turns into an incidental test, could you create a more explicit `TestCaseSensitivity` test or similar?

jonhadfield: Sure. Will do.

jonhadfield: I've added an explicit test and additional check for key case insensitivity (previously just section) also.
Let me know if ok and if you want me to rebase.

jimbocoder:
> Given that mixed case won't work with aws sdk anyway, having aws-vault support that would break backwards compatibility, right?

@FernandoMiguel are you sure about this? I made two different profiles that only differ in case, and with the `aws` cli tool (`v1.16.300`), they are treated distinctly.

I am asking, because I consider re-raising this issue, as the new behavior breaks configs with mixed-case _profile_ names, which is valid. What is invalid is mixed-case _key_ names.

PR #462, opened 2019-12-06
exec --help describe default command

tekumara:
makes it obvious to new users that exec defaults to running the shell as a subprocess.


mtibben:
Thanks @tekumara

PR #464, opened 2019-12-07
Update session duration defaults and the CLI flags

mtibben:
1. Replaces the `--session-ttl`, `--assume-role-ttl` and `--federation-token-ttl` flags with a single `--duration` flag.

   Given that we know which credential call will be made, having 3 flags is more complex than is necessary. This is especially apparent on the login command, where all 3 flags could potentially be used in creating the session, and it's difficult to know which flag to use.
   The `AWS_SESSION_TOKEN_TTL`, `AWS_ASSUME_ROLE_TTL`, `AWS_FEDERATION_TOKEN_TTL` env vars remain available for users who require more fine-grained control over the session.

2. Changes the default durations 
   - `GetSessionToken`: 1h (or 8h when using AssumeRole and the session is primarily to cache MFA) (previously 4h)
   - `AssumeRole`: 1h (previously 15m)
   - `GetFederationToken`: 1h (previously 12h)
   
   Rationale: 
      - 1h is the AWS default for GetSessionToken, AssumeRole and GetFederationToken
      - Anecdotally, 15m very often feels too short and is exceeded when using login
      - When using GetSessionToken without a role, credentials are supposed to be temporary so 4h seems too long
      - 1h is the AWS max for IAM users calling GetFederationToken
      - Having the same 1h default for all three calls is simpler

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -56,31 +56,27 @@ parent_profile = work
 
 ## Environment variables
 
-The following environment variables can be set to override the default flag
-values of `aws-vault` and its subcommands.
-
-For the `aws-vault` command:
+To configure the default flag values of `aws-vault` and its subcommands:
 * `AWS_VAULT_BACKEND`: Secret backend to use (see the flag `--backend`)
 * `AWS_VAULT_KEYCHAIN_NAME`: Name of macOS keychain to use (see the flag `--keychain`)
 * `AWS_VAULT_PROMPT`: Prompt driver to use (see the flag `--prompt`)
 * `AWS_VAULT_PASS_PASSWORD_STORE_DIR`: Pass password store directory (see the flag `--pass-dir`)
 * `AWS_VAULT_PASS_CMD`: Name of the pass executable (see the flag `--pass-cmd`)
 * `AWS_VAULT_PASS_PREFIX`: Prefix to prepend to the item path stored in pass (see the flag `--pass-prefix`)
+* `AWS_VAULT_FILE_PASSPHRASE`: Password for the "file" password store
+* `AWS_CONFIG_FILE`: The location of the AWS config file
 
-For the `exec` subcommand:
-* `AWS_SESSION_TTL`:  Expiration time for aws session (see the flag `--session-ttl`)
-* `AWS_ASSUME_ROLE_TTL`: Expiration time for aws assumed role (see the flag `--assume-role-ttl`)
-
-For the `aws-vault login` subcommand:
-* `AWS_FEDERATION_TOKEN_TTL`: Expiration time for aws console session (see the flag `--federation-token-ttl`)
-* `AWS_ASSUME_ROLE_TTL`: Expiration time for aws assumed role (see the flag `--assume-role-ttl`)
-
-For the `exec`, `login` and `rotate` subcommands:
+To override the AWS config file (used in the `exec`, `login` and `rotate` subcommands):
 * `AWS_REGION`: The AWS region
 * `AWS_DEFAULT_REGION`: The AWS region, applied only if `AWS_REGION` isn't set
 * `AWS_ROLE_ARN`: Specifies the ARN of an IAM role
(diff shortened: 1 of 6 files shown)


mtibben:
Thoughts @FernandoMiguel @lox @jstewmon ?

lox:
Yeah, that sounds super sensible. No complaints here.

FernandoMiguel:
I do like that all values are the same, 1h.

Just not sure going from 15 min to 1h is a good thing, security wise. 

But heck, using aws-vault is already a big gain compared to static keys.. 

I'm OK with the change

mtibben:
> Just not sure going from 15 min to 1h is a good thing, security wise.

Yeah I share this concern, and want to make sure we get the balance right. Initially I was thinking I'd keep `AssumeRole` at 15m and `GetFederationToken` at 1h. However login actually uses `AssumeRole`, so then you'd have to vary the default between the subcommands. This could end up being more confusing.

jinroh:
The short flag for `--mfa-token` has changed with this commit from `-m` to `-t` (previously short for `--session-ttl`). Not sure if this change was on purpose ?

See here:  https://github.com/99designs/aws-vault/pull/464/commits/7aa76dcd9a4ce1fde5438241d9a074443302557d#diff-13522587639c6e93f594555cd0c4c29cR51

mtibben:
Yeah for some reason `login` and `exec` were using different short flags different 😬 . I'll make a note of it in the release notes

jinroh:
@mtibben i see ;)

This part of the documentation should be updated to reflect the change: https://github.com/99designs/aws-vault/blob/master/USAGE.md#using-a-yubikey-as-a-virtual-mfa

zakswindow:
Hi,
I have assume role session duration set to 8 hrs in aws but when i use aws-vault exec <profile-name> i only get 1h session. is it possible to extend it?

zakswindow:
before i was able to use flag --assume-role-ttl but with 5.1.1 it is deprecated. What is the new option? @mtibben

mtibben:
see the release notes https://github.com/99designs/aws-vault/releases/tag/v5.0.0

sellers:
Thanks for the updates on this.  Could this pose a problem for folks who were using different policies for different aspects of the three attributes that are rolled into one.    In order to have the 12h policy for federated token, now all three aspects need to allow the 12h policy (or whatever number you want)?

mtibben:
@sellers [You can use env vars to control these values ](https://github.com/99designs/aws-vault/blob/master/USAGE.md#environment-variables)

ask0n:
@mtibben i have same issue as @zakswindow now. aws-vault definitely reads value for `-d` flag:
```
$ aws-vault login -s -d1m aws-profile --debug
2020/02/20 11:48:38 [keyring] Considering backends: [keychain pass file]
2020/02/20 11:48:38 Loading config file /Users/user/.aws/config
2020/02/20 11:48:38 Parsing config file /Users/user/.aws/config
aws-vault: error: Login failed: Minimum GetSessionToken duration is 15m0s
``` 
but when I set desired value it is just ignored:
```
$ aws-vault login -s -d6h aws-profile --debug
2020/02/20 11:50:40 [keyring] Considering backends: [keychain pass file]
2020/02/20 11:50:40 Loading config file /Users/user/.aws/config
2020/02/20 11:50:40 Parsing config file /Users/user/.aws/config
2020/02/20 11:50:40 Looking up keyring for aws-profile
2020/02/20 11:50:40 [keyring] Querying keychain for service="aws-vault", account="aws-profile", keychain="aws-vault.keychain"
2020/02/20 11:50:40 [keyring] Found item "aws-vault (aws-profile)"
2020/02/20 11:50:41 Using GetFederationToken for credentials
2020/02/20 11:50:42 Generated credentials ****************VDMT using GetFederationToken, expires in 1h0m1.989163s
```

mtibben:
@ask0n can you try using the latest aws-vault release?

ask0n:
@mtibben works with v5.3.2, thanks!

Issue #337, opened 2019-02-14
boto3 script and profile switching


ismailyenigul:
Hi,
I need a help to use aws-vault in boto3 codes.

when I create boto3 script I can use *.aws/credentials* file passing the profile and also I can switch from one to another profile easly:
```python session = boto3.Session(profile_name=my_aws_profile)```

If I want to use aws-vault, Should I  use env vars and call aws-vault everytime because credentials are saved in other place?

**Comment by FernandoMiguel** [comment]
just run aws-vault exec PROFILE -- python3 /PATH/YOUR_SCRIPT.py

**Comment by KursLabIgor** [comment]
no it doesnt work as mentioned above. OR need to perform another configuration of boto
```python 
botocore.exceptions.InvalidConfigError: The source_profile "terraform_dev" must specify either static credentials or an assume role configuration
```

ryan-gerstenkorn-sp:
For anyone coming across this ticket because of the error below it seems this is caused by an old version of awscli (in my case ubuntu 18.04's default version). I believe this would apply to the SDK as well.

> The source_profile "terraform_dev" must specify either static credentials or an assume role configuration

rossigee:
A dirty hack, in case it helps others finding this bug report to use `aws-vault` with boto...

```
  envvars = subprocess.check_output(['aws-vault', 'exec', aws_profile, '--', 'env'])
  for envline in envvars.split(b'\n'):
      line = envline.decode('utf8')
      eqpos = line.find('=')
      if eqpos < 4:
          continue
      k = line[0:eqpos]
      v = line[eqpos+1:]
      if k == 'AWS_ACCESS_KEY_ID':
        aws_access_key_id = v
      if k == 'AWS_SECRET_ACCESS_KEY':
        aws_secret_access_key = v
      if k == 'AWS_SESSION_TOKEN':
        aws_session_token = v

  session = boto3.Session(
    aws_access_key_id, aws_secret_access_key, aws_session_token
  )
```

mjstevens777:
And in case it helps anyone else, the `export` command is a little cleaner

```python
import subprocess
import json
import boto3

def get_vault_session(aws_profile):
    aws_env_json = subprocess.check_output(['aws-vault', 'export', '--format=json', aws_profile])
    aws_env = json.loads(aws_env_json)
    return boto3.Session(
        aws_access_key_id=aws_env['AccessKeyId'],
        aws_secret_access_key=aws_env['SecretAccessKey'],
[...]

PR #465, opened 2019-12-08
add shell command.

jonhadfield:
This PR is an **early** attempt to add a new feature: to be able fork to a shell with assumed role credentials set in the environment.
The use-case is that I often need to run scripts and applications at the command line that depend on temporary credentials (obtained with MFA to satisfy Trust Relationship), for example: ansible, terraform, and self-written. I think this old issue expresses the same need: https://github.com/99designs/aws-vault/issues/337
This is currently working for me with:
`aws-vault shell <profile>`
The result being that I end up in a shell with the relevant _AWS__ envvars set for ansible, etc. to make use of.
I realise I could achieve the same result with:
`aws-vault exec <profile> -- ansible-playbook ...`
But that adds delay in having to both prefix every command and in retrieving new credentials each time.
This should work for macOS, Linux/Unix variants, and Windows.
Any feedback appreciated.

### DIFF
--- a/cli/shell.go
+++ b/cli/shell.go
@@ -1,0 +1,154 @@
+package cli
+
+import (
+	"errors"
+	"fmt"
+	"github.com/99designs/aws-vault/vault"
+	"github.com/99designs/keyring"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"gopkg.in/alecthomas/kingpin.v2"
+	"os"
+	osexec "os/exec"
+	"runtime"
+	"syscall"
+)
+
+type ShellCommandInput struct {
+	ProfileName string
+	Command     string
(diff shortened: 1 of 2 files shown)


FernandoMiguel:
Isn't it the all point of having aws proxy command exactly so that you don't need to prefix every command?

Plus shell history makes this so fast

mtibben:
Hey @jonhadfield, isn't this exactly what `aws-vault exec <profile>` achieves? By default `aws-vault exec` will fork a shell

jonhadfield:
Sorry, I missed this, and that is what I'm after. Thanks!
IMHO the mention of exec on the README should also include this (just more clearly perhaps?) feature as it's easily overlooked.

PR #470, opened 2019-12-18
Adding instructions for using credential_process with aws-vault

mhumeSF:


mtibben:
This looks like an opinionated workflow. Perhaps add it to USAGE.md under a Recipes header or similar

mtibben:
Added in 51909f167a6532408c7e0f28f69b4c403e2f179a

mhumeSF:
@mtibben Totally agree belongs in `USAGE.md`. Thanks for the addition!

Issue #122, opened 2017-05-19
Force new session option


coen-hyde:
It would be nice to have a force new session option. We need a guaranteed period of time where the session is valid. Currently we set the session length larger than desired and set the --debug flag so we can watch for session endings.

Or put another way. No session caching.

lox:
Sounds pretty reasonable. Would be interested to know a bit more about the
use case!

coen-hyde:
We use Terraform for infrastructure provisioning. Some deployments / modifications are known to take a long time. For example Elastic Search or RDS modifications. These can take up to 20 minutes to complete. Expiration of AWS credentials half way through can be pretty bad as the Terraform state file will become out of sync with reality.

lox:
Have you checked out the --server mode? It will generate new credentials
via a fake IAM Roles server

coen-hyde:
Interesting, I missed that in the documentation. If I understand correctly the fake IAM server runs on localhost and reissues temporary credentials when requested? If so that would be a problem for our setup. We use MFA. Also i'd prefer temporary credentials not be available to whoever has access to the laptop without re-authenticating.

chbiel:
i would like to have this feature. if someone has an idea where this should go into the code i would like to try to implement it

FernandoMiguel:
@chbiel what are you trying to achieve?

chbiel:
we have the same problem as @coen-hyde .
we are enforced to use mfa with a session timout of 1h. it happens regularly that the session ends during a aws-vault exec run with terraform and we have to clean up the account after such "crash".
i would like to have e.g. a commandline option like "--renew-mfa" that deletes the session and forces me to reenter my mfa credentials.

another cool thing we be to have something like:
"your sessions end in xx minutes! do you want to refresh? y/n"
and to make it configurable when this should be displayed, e.g. when the session ends in less than 30 minutes.

i hope the use case got clear :)

FernandoMiguel:
@chbiel you can now increase the session up to 12h.
that should help.
running with `--server` would also request MFA (at least does on Mac)

adding an extra command to aws-vault will not help in anyway here.
it is as simple as re-running the command and it will reauth

chbiel:
badly i am not allowed to change the session timeout (big company, central it, etc...)

The problem is that there is no easy way to find out, how long my session is open and when the timeout will be reached.
i already found a workaround by using: `aws-vault remove -s` to delete my session and force a revreationg of the session

[...]

PR #474, opened 2019-12-21
add ability to force a new session rather than fetch from cache

ZoidBB:
Fixes issues #122.

There are numerous use-cases where the credentials server won't work, particular when building tooling /around/ aws-vault itself.

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -49,6 +49,10 @@ func ConfigureExecCommand(app *kingpin.Application) {
 		Short('n').
 		BoolVar(&input.NoSession)
 
+	cmd.Flag("force-new-session", "Force a new session to be created").
+		Short('f').
+		BoolVar(&input.Config.ForceNewSession)
+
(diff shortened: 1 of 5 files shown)


mtibben:
There is a global that does this @ZoidBB, perhaps this could be controlled with an env var? https://github.com/99designs/aws-vault/blob/0112406b2cc4f63ea55baaa51e592a8d0aa08bba/vault/vault.go#L18

ZoidBB:
My thought here was that UseSessionCache turns off the cache entirely, whereas "force-new-session" forces a new session /and/ caches it. What are your thoughts?

mtibben:
`--duration` was added in v5 which I think addresses #122. Closing for now

jimbrowne:
I also need this functionality so a babysitting process can prompt/force a refresh of credentials when expiration is soon to ensure no period of time in which credentials are stale.

Issue #475, opened 2019-12-22
aws-vault v5.0.0 removes default profile block header


mbacchi:
When adding a new credential to `aws-vault` on v5.0.0, it removes the `[default]` block header.

```
$ aws-vault --version
v5.0.0
$ aws-vault list
Profile                  Credentials              Sessions                 
=======                  ===========              ========                 
default                  default (missing)        -                        
user1                    user1                    1577058481 (mfa) 
$ cat ~/.aws/config 
[default]
output=json
region=us-east-1

[profile user1]
mfa_serial=arn:aws:iam::XXX:mfa/user1
region=us-east-2

$ aws-vault add experiment_user1
Enter Access Key ID: XXX
Enter Secret Access Key: XXX
Added credentials to profile "experiment_user1" in vault

$ cat ~/.aws/config 
output=json
region=us-east-1

[profile user1]
mfa_serial=arn:aws:iam::XXX:mfa/user1
region=us-east-2

[profile experiment_user1]


```

ZoidBB:
I had this issue a couple days ago, can confirm. I'm going to look into what's causing this.

ZoidBB:
The INI lib being used doesn't put the header over default values by default, so this ended up being a one-line change to configure it to output the header.

I've created PR #476 to address this.

mtibben:
Fixed in #476

PR #476, opened 2019-12-23
475 5.0 removes default block header

ZoidBB:
Fixes issue #475 by configuring the INI lib to output the default header. Some other tools that read the AWS config file bork when this header is missing, despite it being valid for the INI specification. To be most compatible with other tools, any tool that edits/rewrites these files should use the same format as the stock AWS CLI tools -- and thus have the default header.

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -36,6 +36,7 @@ const (
 )
 
 func init() {
+	ini.DefaultHeader = true
 	ini.PrettyFormat = false
 }
 
--- a/vault/config.go
+++ b/vault/config.go
@@ -202,6 +202,10 @@ func (c *ConfigFile) ProfileSection(name string) (ProfileSection, bool) {
 	return profile, true
 }
 
+func (c *ConfigFile) Save() error {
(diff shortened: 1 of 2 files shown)


mtibben:
Thanks @ZoidBB, could you include a specific test for this?

ZoidBB:
@mtibben done. i added a Save() method to ConfigFile as well, so it can be saved from memory to file manually without having to necessarily modify the in-memory config (right now the only way to right to disk is via ConfigFile.Add, which doesnt make sense for this test).

mtibben:
Hmm what do we do in the case where the ini file does not have a default header?

ZoidBB:
If the INI file as-read doesn't have one, it'll get one -- which will bring it back into consistency w/what the AWS CLI does during its own rewrites of the file. I think erring on the side of following AWS CLI is the best idea here, rather than trying to follow what the existing file has at read time.

ZoidBB:
I can add a test for that scenario if you'd like

mtibben:
Great looks good, thanks @ZoidBB

PR #480, opened 2019-12-31
Adds support for output

esacteksab:
New to Go and new to this library. Trying to capture support for this https://docs.aws.amazon.com/cli/latest/userguide//cli-chap-configure.html#cli-quick-configuration-format. 

Won't be offended if you don't want this. Happy to take any feedback or direction should I have missed something or did something incorrectly. 

Thanks!

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -144,6 +144,7 @@ type ProfileSection struct {
 	DurationSeconds uint   `ini:"duration_seconds,omitempty"`
 	SourceProfile   string `ini:"source_profile,omitempty"`
 	ParentProfile   string `ini:"parent_profile,omitempty"`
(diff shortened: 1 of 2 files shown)


mtibben:
Hey @esacteksab, what's the purpose of supporting `output`? I'd be happy to merge if was used by aws-vault for some purpose, but if it's not going to be used it should be fine to be ignored

esacteksab:
Apologies. Not sure how I missed the reply. We use `aws-vault` as a library for a script we wrote leveraging `aws-vault`'s mechanism to generate the aws config file, but `aws-vault` doesn't support all the supported parameters for the aws config file (like output), so wanted to add that.

PR #514, opened 2020-01-30
add-mfa-device, remove-mfa-device & Yubikey

j0hnsmith:
So upon reflection #312 was a hack.

This PR adds support for reading a OTP from a Yubikey natively on a Mac (behind filename build flag, requires `CGO=1` which is implicit when building on a Mac) and via `ykman`
on other systems (essentially includes #373, thanks @viraptor).

Two new commands are added `add-mfa-device` and `remove-mfa-device`.

The `--prompt` flag becomes `--mfa-token-provider` and `AWS_VAULT_PROMPT` is changed to `AWS_VAULT_MFA_TOKEN_PROVIDER` to reflect the refactor of prompt functions into `mfa.TokenProviders`, feels like it fits in better with the rest of the codebase.

@mtibben I _really_ hope that this fits in with the build/release process, or can be accommodated somehow. Please try to build locally and let me know any issues you encounter.

Supersedes #392.


main.go
```diff
@@ -33,6 +34,8 @@ func run(args []string, exit func(int)) {
 	cli.ConfigureRemoveCommand(app)
 	cli.ConfigureLoginCommand(app)
 	cli.ConfigureServerCommand(app)
+	cli.ConfigureAddMfaCommand(app)
+	cli.ConfigureRemoveMfaCommand(app)
```
mtibben: would list make sense here?

mfa/mfatokenprovider.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+)
+
+// TokenProvider is an interface to provide an mfa token. It's intended that provided do whatever is necessary to get
+// a token, eg prompt the use via the terminal or fetch it from a yubikey.
+type TokenProvider interface {
+	Retrieve(mfaSerial string) (string, error)
+}
```
mtibben: Perhaps `type TokenProvider func() (string, error)` would be simpler (and would match the AWS SDK)

vault/vault.go
```diff
@@ -62,14 +52,20 @@ func NewSessionTokenProvider(creds *credentials.Credentials, k *CredentialKeyrin
 		return nil, err
 	}
 
+	var tokenProvider mfa.TokenProvider
+	if config.MfaToken != "" {
+		tokenProvider = mfa.KnownToken{Token: config.MfaToken}
+	} else {
+		tokenProvider = mfa.GetTokenProvider(config.MfaTokenProvider)
+	}
```
mtibben: Feels like this logic might better sit in Config? e.g. `config.GetTokenProvider()`

vault/vault.go
```diff
@@ -27,24 +27,14 @@ func FormatKeyForDisplay(k string) string {
 
 // Mfa contains options for an MFA device
 type Mfa struct {
-	MfaToken        string
-	MfaPromptMethod string
-	MfaSerial       string
+	TokenProvider mfa.TokenProvider
+	Serial        string
 }
```
mtibben: this struct may now be unnecessary

mfa/yubikey_darwin_amd64.go
```diff
@@ -0,0 +1,102 @@
+package mfa
```
mtibben: Needs a build tag?

vault/iammfa.go
```diff
@@ -0,0 +1,198 @@
+package vault
```
mtibben: this no longer has anything to do with "vault"ing, so might fit better in a different package?

mtibben:
> Please try to build locally and let me know any issues you encounter 

Builds OK on my machine, haven't tested on other platforms yet

mtibben:
The QR code generated in my terminal is huge. can we reduce the size at all?

cli/addmfadevice.go
```diff
@@ -0,0 +1,83 @@
...
+	cmd := app.Command("add-mfa-device", "Adds a MFA device to the IAM user")
+	cmd.Arg("username", "Name of the user to add the MFA as device for").
+		Required().
+		StringVar(&input.Username)
+
+	cmd.Arg("profile", "Name of the profile").
+		Required().
+		HintAction(awsConfigFile.ProfileNames).
+		StringVar(&input.ProfileName)
+
```
mtibben: Should we flip this to `add-mfa-device <profile> <username>`, as all the other commands have profile as the first argument

vault/vault.go
j0hnsmith: I had tried unsuccessfully to get rid of that, had another go and it's gone but I'm not entirely happy with the result.

mfa/mfatokenprovider.go
j0hnsmith: At some level the serial is needed as any kind of tool that can generate mfa tokens for multiple things needs the serial to know which token to generate. In `SessionTokenProvider` and `AssumeRoleProvider` the mfa token is useless on its own as you also have to pass the mfa serial when authing. So the serial has be stored somewhere, I've considered expanding the interface to include the serial but haven't come up with anything I like (suggestions welcome). Possibly clunky but the obvious choice is something along these lines, could  simplify code elsewhere though.

```
type TokenProvider interface {
    GetToken() string, error
    GetSerial() string, error
    SetSerial(serial string)
}
```

Where do you see that signature in the aws SDK, can't find it, would like to see that usage?

j0hnsmith:
> Builds OK on my machine, haven't tested on other platforms yet

Great

> The QR code generated in my terminal is huge and doesn't fit in the window. Can we reduce the size at all?

I did try the smaller 'half blocks' size but it's barely recognisable as a QR code, got errors when scanning it. I'll try other libs but the limitation is the characters available to a terminal and the font size.

vault/iammfa.go
j0hnsmith: will change

mfa/yubikey_darwin_amd64.go
mtibben: 👍 Ah yep cool. So we can drop the amd64? `yubikey_darwin.go`

main.go
mtibben: Oh I mean we have "add-mfa-device", "rm-mfa-device" but no way to list the mfa devices

mfa/yubikey_darwin_amd64.go
j0hnsmith: It's `_$GOOS_$GOARCH.go`, I don't think are there any non amd64 darwin machines around but won't hurt to keep it as it is.

main.go
j0hnsmith: Reverted my change. You can only have one mfa device per iam user and you probably know if you've added one or not, but happy to add for completeness. Actually the cli has [`list-mfa-devices`](https://docs.aws.amazon.com/cli/latest/reference/iam/list-mfa-devices.html) so will add, the output will be similar

j0hnsmith: added

mfa/mfatokenprovider.go
j0hnsmith: I've refactored `mfa.TokenProvider` to include serial, it has simplified other code too.

j0hnsmith:
@mtibben Ready for another look. 

I reduced the border around the qr code, can't get it any smaller, even with other libs but did add a message.

ilpianista:
@j0hnsmith please port [kdialog](https://github.com/99designs/aws-vault/commit/b5a6d8f745ee972e2fb7cb22a1946887729e1137) also to TokenProvider. Thanks!

j0hnsmith:
@ilpianista done

cli/removemfadevice.go
```diff
@@ -0,0 +1,83 @@
...
+
+	sess, err := session.NewSessionWithOptions(session.Options{
+		Config: aws.Config{
+			Region:      aws.String(p.Region),
+			Credentials: credentials.NewCredentials(&credentials.StaticProvider{Value: creds}),
+		},
+	})
+	if err != nil {
+		app.Fatalf("error creating session: %s", err.Error())
+	}
```
mtibben: You can use vault.NewSession for this
https://github.com/99designs/aws-vault/blob/f26b718f741a3f5c51809982539e6d7a1b6f4280/vault/vault.go#L20

cli/listmfadevices.go
```diff
@@ -0,0 +1,88 @@
...
+			Keyring:         &vault.CredentialKeyring{Keyring: input.Keyring},
+			CredentialsName: p.Name,
+		}
+
+		creds, err := provider.Retrieve()
+		if err != nil {
+			app.Fatalf("unable to retrieve creds for profile %s", p.Name)
+		}
+
+		sess, err = session.NewSessionWithOptions(session.Options{
```
mtibben: You can use vault.NewSession for this
https://github.com/99designs/aws-vault/blob/f26b718f741a3f5c51809982539e6d7a1b6f4280/vault/vault.go#L20

cli/addmfadevice.go
```diff
@@ -0,0 +1,83 @@
...
+		Keyring:         &vault.CredentialKeyring{Keyring: input.Keyring},
+		CredentialsName: p.Name,
+	}
+
+	masterCreds, err := provider.Retrieve()
+	if err != nil {
+		app.Fatalf("unable to retrieve master creds")
+	}
+
+	sess, err := session.NewSessionWithOptions(session.Options{
```
mtibben: You can use vault.NewSession for this
https://github.com/99designs/aws-vault/blob/f26b718f741a3f5c51809982539e6d7a1b6f4280/vault/vault.go#L20

cli/global.go
```diff
@@ -47,10 +47,10 @@ func ConfigureGlobals(app *kingpin.Application) {
 		Envar("AWS_VAULT_BACKEND").
 		EnumVar(&GlobalFlags.Backend, backendsAvailable...)
 
-	app.Flag("prompt", fmt.Sprintf("Prompt driver to use %v", promptsAvailable)).
+	app.Flag("mfa-token-provider", fmt.Sprintf("Mfa token provider to use %v", mfaTokenProvidersAvailable)).
 		Default("terminal").
-		Envar("AWS_VAULT_PROMPT").
-		EnumVar(&GlobalFlags.PromptDriver, promptsAvailable...)
+		Envar("AWS_VAULT_MFA_TOKEN_PROVIDER").
+		EnumVar(&GlobalFlags.MfaTokenProvider, mfaTokenProvidersAvailable...)
```
mtibben: Can we offer backwards compatibility for `--prompt` here? Kingpin gives you `Hidden` to achieve that

mtibben:
There are conflicting files, can you rebase?

cli/global.go
j0hnsmith: done

cli/removemfadevice.go
j0hnsmith: done (I think it's as you expect but not totally sure, please check)

j0hnsmith:
@mtibben rebased

mtibben:
I've only just grasped that this PR is not using U2F at all as AWS API doesn't support it - I didn't realise that. So this is a wrapper around yubikey TOTP, but falling back to using ykman?

So firstly, if ykman is good enough for other platforms, why not macOS? It seems like a simpler approach to me.

Secondly, the commands for adding/removing mfa devices overlap with aws-cli. Are there issues with the aws-cli that motivated you to write that? I can see the ergonomics of the commands are easier to use, but can't see any different functionality.  The add-mfa-device command doesn't do anything specifically for the yubikey, so do you still need to use `ykman`  to register the MFA virtual device with yubikey. And the mfa commands seem a little out of place in aws-vault - they don't directly relate to vaulting or generating creds

An alternative suggestion: I'm thinking that we should:
 1. Spin your code here out as it's own CLI tool - `aws-mfa`? The commands here are useful in their own right to be able to easily manipulate MFA devices
 2. Create an aws-vault MFA token provider to take inputs from an external command
 3. Add token providers in aws-vault for `ykman` and `aws-mfa` i.e. revive #373

Thoughts?

j0hnsmith:
> So this is a wrapper around yubikey TOTP, but falling back to using ykman?

Yes

> So firstly, if ykman is good enough for other platforms, why not macOS? It seems like a simpler approach to me.

`ykman` is a python library, it needs to be installed and has dependencies. Adding the functionality directly to `aws-vault` means there aren't any external dependencies.

> Secondly, the commands for adding/removing mfa devices overlap with aws-cli. Are there issues with the aws-cli that motivated you to write that? 

The main reason is to set the id in the token provider to ensure the correct TOTP can be generated when needed. If it's up to the user, some will modify the string/config then be unable to get a token.

> The add-mfa-device command doesn't do anything specifically for the yubikey, so do you still need to use ykman to register the MFA virtual device with yubikey.

The QR code can be used by app called Yubikey Authenticator (that sets it up with the correct id) or any other app that uses a QR code, alternatively you could use the uri with `ykman` or anything else that uses uris. There's nothing Yubikey specific however the details in the uri/QR code are important.

> And the mfa commands seem a little out of place in aws-vault

I was going to mention this when you asked for the list command. The refactor of prompts to mfa token providers creates an interface for `aws-vault` to get tokens from external places. The key to be able to get a token is to set it up correctly, the add command is essentially a way to ensure interoperability (via the [key uri format](https://github.com/google/google-authenticator/wiki/Key-Uri-Format)). I totally get that the add functionality isn't strictly necessary however pragmatically/ergonomically it is necessary (well without it the user experience wouldn't be very good, wouldn't work as expected in many cases).

> Spin your code here out as it's own CLI tool - aws-mfa? 

I don't think I'd do that, an extra dependency feels weird and by the time users realise they want it they'll probably have already setup their mfa device, removing and readding leaves room for errors/mistakes which could mean people can't access their accounts.

My preference is for the add comment to be retained, happy to remove the list and remove commands. If you're really not happy with that we could just keep the mfa token provider bit that opens up interoperability with other commands that are able to provide tokens.

mtibben:
> ykman is a python library, it needs to be installed and has dependencies. Adding the functionality directly to aws-vault means there aren't any external dependencies.

Right so the argument for having this in aws-vault is for portability? However we're only building this for macos - python is already available on macos and `brew install ykman` is one command away. If we were able to build cross platform there would be a good case for portability

> The main reason is to set the id in the token provider to ensure the correct TOTP can be generated when needed. If it's up to the user, some will modify the string/config then be unable to get a token.

Right but this is a very specific workflow for using yubikey TOTP. Should we consider other hardware or virtual TOTP devices? Feels like this is almost a contrib script, rather than part of general purpose cli tool. If this was a generalised "u2f" or "virtual totp" integration, I think inclusion would make more sense. This functionality seems more aligned with ykman than aws-vault

> I don't think I'd do that, an extra dependency feels weird and by the time users realise they want it they'll probably have already setup their mfa device, removing and readding leaves room for errors/mistakes which could mean people can't access their accounts.

I'm not sure about that - if we have a section in the docs explaining how to use yubikey, it's not unreasonable that they need to use other tools to set it up - they'll already need ykman. If we pull out this code, maybe it's a binary that can be downloaded to do it all, or alternatively this could be a bash script running aws-cli and ykman to set up the device correctly - it wouldn't even need aws-vault right?

j0hnsmith:
I have counter arguments however this PR is too big, it contains 3 distinct things, for more targeted discussions I'm going to close this and turn it into 3 separate PRs so that resistance to one thing doesn't detract from others. The first is https://github.com/99designs/aws-vault/pull/533

Issue #495, opened 2020-01-14
Should AWS_SECURITY_TOKEN be removed or have a flag for it ?


mateimicu:
The`AWS_SECURITY_TOKEN` is set and also `AWS_SESSION_TOKEN` but looking at this [article](https://aws.amazon.com/blogs/security/a-new-and-standardized-way-to-manage-credentials-in-the-aws-sdks/) it seems like `AWS_SESSION_TOKEN` is the standard.

Having both variables can cause weird problems like https://github.com/gruntwork-io/terragrunt/issues/998, I was wondering if it makes sense to add a flag for this variable or to remove it completely.

mtibben:
Yes it probably makes sense to put these legacy env vars behind some flag

mateimicu:
I'm new to golang but I'll try to propose a PR.

mateimicu:
Created https://github.com/99designs/aws-vault/pull/516

mtibben:
Closing for now, haven't heard of anyone else with issues and it's easy to unset in the shell if required

PR #516, opened 2020-02-06
Add a flag for enabling AWS_SECURITY_TOKEN

mateimicu:
AWS_SECURITY_TOKEN was dropped in favor of AWS_SESSION_TOKEN, currently
aws-vault sets both variables in a new shell that can cause problems.

I preserved the default behavior but put it under a flag

I looked at `exec_test.go` I was wondering if it makes sense to add a test there and how would you do it?

Addresses #495

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -19,15 +19,16 @@ import (
 )
 
 type ExecCommandInput struct {
-	ProfileName      string
-	Command          string
-	Args             []string
-	Keyring          keyring.Keyring
-	StartServer      bool
-	CredentialHelper bool
-	Config           vault.Config
-	SessionDuration  time.Duration
-	NoSession        bool
+	ProfileName             string
+	Command                 string
+	Args                    []string
+	Keyring                 keyring.Keyring
+	StartServer             bool
+	CredentialHelper        bool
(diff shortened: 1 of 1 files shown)


j0hnsmith:
Can you document it https://github.com/99designs/aws-vault/blob/master/USAGE.md#environment-variables

mateimicu:
@j0hnsmith I am not sure what to document in that section, I don't see `AWS_SECURITY_TOKEN` or `AWS_SESSION_TOKEN` being documented there.

cli/exec.go
```diff
@@ -45,6 +46,9 @@ func ConfigureExecCommand(app *kingpin.Application) {
 
 	cmd := app.Command("exec", "Executes a command with AWS credentials in the environment")
 
+	cmd.Flag("use-legacy-aws-secure-token", "Export also the deprecated AWS_SECURE_TOKEN alongside the standard, AWS_SESSION_TOKEN").
```
mtibben: Lets flip it for now, maybe `--disable-legacy-env-vars`?

mtibben:
Closing for now, haven't heard of anyone else with issues and it's easy to unset in the shell if required

PR #522, opened 2020-02-11
Improve completion script

### LINKED ISSUE #415: Improved auto-completion for zsh
I made a ZSH completion script for aws-vault, as I was frustrated that I didn't find a decent one. Here it is: https://github.com/reegnz/aws-vault-zsh-plugin

What is missing from the existing completion script is being able to complete profile names, and that was more important to me than completing flags.

Do you think it would be worth it to contribute to your repository?

reegnz:
For the commands `rotate` and `remove` only provide completion hints of
profiles that aws-vault is managing credentials for.

Fixes #415

### DIFF
--- a/cli/global.go
+++ b/cli/global.go
@@ -19,6 +19,7 @@ const (
 
 var (
 	keyringImpl      keyring.Keyring
+	credKeyring      *vault.CredentialKeyring
 	awsConfigFile    *vault.ConfigFile
 	configLoader     *vault.ConfigLoader
 	promptsAvailable = prompt.Available()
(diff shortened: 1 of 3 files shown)


reegnz:
I am a golang noob, so feel free to give me some hints on what to change.
I managed to try out the changes on my local system with `make install` and the completion for `rotate` and `remove` properly reduces the profiles list to the ones where aws-vault is managing the key for.

mtibben:
The problem here is that it requires access to the keychain/keyring backend in order for shell autocomplete. Some backends will work fine as this is all metadata, and some may put up the password prompt. I'd prefer that the autocomplete script not need access to the actual keyring backend.

reegnz:
Could it be possible to store metadata not in the keychain then to improve usability?

reegnz:
Could it be possible to store metadata not in the keychain then to improve usability?

PR #523, opened 2020-02-12
Add support for kdialog

ilpianista:
This allows to use [KDialog](https://kde.org/applications/utilities/org.kde.kdialog) as prompt.


mtibben:
Thanks @ilpianista

Issue #529, opened 2020-02-18
The pass keyring backend doesn't follow symlinks


stefancocora:
- [x] I am using the latest release of AWS Vault
- ~~I have provided my `.aws/config` (redacted if necessary)~~ not necessary
- [x] I have provided the debug output using `aws-vault --debug` (redacted if necessary)

My setup is such that the `pass` storage location is behind a symlink
```
ls -lh ~/.password-store

/home/user/.password-store -> /the/real/path/to/.password-store
```

Example aws-vault with keyring v1.1.4 failing to retrieve secrets from the `pass` backend

```
aws-vault --version
v5.2.1

aws-vault --backend=pass --debug  exec pers/aws/profiles/testground -- aws sts get-caller-identity
2020/02/18 19:20:19 [keyring] Considering backends: [pass]
2020/02/18 19:20:19 Loading config file /home/user/.aws/config
2020/02/18 19:20:19 Parsing config file /home/user/.aws/config
aws-vault: error: exec: Error getting temporary credentials: profile pers/aws/profiles/testground: credentials missing

```

This is a follow up fix dependent on the fix in the keyring library https://github.com/99designs/keyring/pull/63

PR to follow.

stefancocora:
Waiting on (in order):
- https://github.com/99designs/keyring/pull/63
- https://github.com/99designs/aws-vault/pull/530

stefancocora:
https://github.com/99designs/keyring/pull/63

PR #530, opened 2020-02-18
Fix/follow simlink to passwordstore

stefancocora:
[fix] Better way to deal with path names returned by following symlink

A better way of dealing with []keyName(s) retrieved from a symlink to
passwordstore.

A keyring fix will complement this work, waiting for upstream fixes in keyring:master
which has introduced what appears to be broken unit tests. See context in comments in https://github.com/99designs/keyring/issues/62

Related
https://github.com/99designs/keyring/issues/62
https://github.com/99designs/keyring/pull/63

Fixes
https://github.com/99designs/aws-vault/issues/529

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ BUILD_FLAGS=-ldflags="-X main.Version=$(VERSION) -s -w" -trimpath
 CERT_ID ?= Developer ID Application: 99designs Inc (NRM9HVJ62Z)
 SRC=$(shell find . -name '*.go')
 
+export CGO_ENABLED=0
+
 .PHONY: binaries clean release install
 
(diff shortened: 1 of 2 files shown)


mtibben:
I would think that if there's any issues with symlinks, they should probably be handled exclusively in keyring, not sure why it would affect any aws-vault code. So I don't really understand what this PR is doing, or why you are doing it. Are you able to explain the use-case with full examples and/or tests?

stefancocora:
This PR was meant to be a follow up PR that would use the keyring "follow symlinks" fix once that has been merged and re-build aws-vault with that fix.
- https://github.com/99designs/keyring/pull/63

PR #539, opened 2020-03-09
Support specifying custom libsecret collection to use as backend storage

Neverous:
Adds `--secret-service-collection=COLLECTION_NAME` flag allowing to specify different secret-service (libsecret) collection at runtime instead of currently static default `awsvault`. 
Similar to macOS keychain name.


cli/global.go
```diff
@@ -57,6 +58,11 @@ func ConfigureGlobals(app *kingpin.Application) {
 		Envar("AWS_VAULT_KEYCHAIN_NAME").
 		StringVar(&GlobalFlags.KeychainName)
 
+	app.Flag("libsecret-collection", "Name of libsecret collection to use, if it doesn't exist it will be created").
```
mtibben, line 61: The backend is called `secret-service` So might make sense for this to be `--secret-service-collection`?

Neverous: Sure, I guess. I was going by preexisting variable names, but backend name is probably better for user-provided parameters as libsecret is not mentioned anywhere.

Issue #534, opened 2020-02-26
Constantly have to re-enter keychain password, part 3


Gowiem:
# TODOs
- [x] I am using the latest release of AWS Vault'
- I'm on v5.3.2
- [x] I have provided my `.aws/config` (redacted if necessary)
  - Can provide if needed.
- [x] I have provided the debug output using `aws-vault --debug` (redacted if necessary)
  - Can provide if needed. 

# Info
Following up on #219. I believe myself and others are still experiencing this issue. It sounds as if there is an ability to work around it, but it is not documented? I'm creating this issue so it is continued to be surfaced that a workaround or fix is still desired. 

I've tried the one workaround mentioned in #219 with using `export AWS_VAULT_KEYCHAIN_NAME=login`, but that just causes `aws-vault exec *` to spit out the following:

```
aws-vault: error: exec: Error getting temporary credentials: profile OMITTED: credentials missing
```

Any suggestions here? Happy to make a documentation update once I understand what need to be done so I don't need to enter my long password constantly 😅

mtibben:
> Can provide if needed

please do provide!

Gowiem:
@mtibben Gotcha, no problem. 

*Issue*: My personal issue is that I need to type in my aws-vault keyvault password every 5 (maybe 10?) or so minutes to run `aws-vault` commands. I just recently installed yesterday, got all my profiles setup, and started using the tool. On first use, I was prompted for my password twice, once with a "always allow" button which I selected, and then again a second time. On each subsequent use after that I am again prompted every 5 or minutes for my password to use the command. This would be fine if it was a longer session that was being created, but it's just painful with how short it is. 

`~/.aws/config`:
```
[default]
output=json
region=us-east-1

# ... 5 other, empty profiles

[profile made-up-name]
region=us-west-2
```

Here is some `aws-vault --debug` output where the first command I'm trying to use the workaround mentioned in #211 and the second command I'm using the default setup. I was prompted for my password on the second `aws-vault` command since I hadn't used `aws-vault` within the last 10 minutes. 
```
» export AWS_VAULT_KEYCHAIN_NAME=login                                                                                                      [py: v3.5.9] [node: v12.14.0] [ruby: v2.6.3] gowiem@GowieBookPro2
» av --debug exec made-up-name -- aws s3 ls                                                                                                 [py: v3.5.9] [node: v12.14.0] [ruby: v2.6.3] gowiem@GowieBookPro2
[...]

Issue #211, opened 2018-01-11
Constantly have to re-enter keychain password


jkugler:
OS: OS X 10.12.6
Using iTerm2

I use `aws-vault exec <profile> -- comand goes here` extensively.  However, I have to enter the keychain password (sometimes twice), although I press "Always allow" every time I enter the password, and on every command I run.

Is there a way to be able to enter the keychain password only once every "configurable amount of time"?

FernandoMiguel:
After -- press enter 
Then you are in a subshell that will last about an hour if no mfa 
You can change the ttl up to eight hours.

lox:
We fixed some issues related to this recently. We'll get out a new release shortly.

lox:
Try https://github.com/99designs/aws-vault/releases/tag/v4.2.0 @jkugler (thanks for the quick response @mtibben)

jkugler:
Thank you. I'll try to give it a try tomorrow. I install it via brew, so will have to fiddle some to clear out that version and install it manually.

lox:
We'll land it in brew next!

endzyme:
This is still happening quite often - I get the "Allow Always" dialog for about 75% of my aws-vault exec commands.

seanscottking:
I'm still seeing this problem as well. OSX 11 El Capitan (yes I know I need to upgrade). Installed 4.2.0 via brew.

```
$ brew cask info aws-vault
aws-vault: 4.2.0
https://github.com/99designs/aws-vault
/usr/local/Caskroom/aws-vault/4.2.0 (9.5MB)
From: https://github.com/caskroom/homebrew-cask/blob/master/Casks/aws-vault.rb
==> Name
aws-vault
==> Artifacts
aws-vault-darwin-amd64 -> aws-vault (Binary)
```

PR #540, opened 2020-03-09
Adds info on changing Keychain password timeout

Gowiem:
This is a common issue for folks -- See issues: #534, #219, and #211. Hopefully this documentation helps others find this simple way to configure how often you need to supply your password.


Gowiem:
@z0ph Updated with your recent instructions from the issue! 😄

mtibben:
Thanks @Gowiem and @z0ph

PR #542, opened 2020-03-17
Allow AWS region retrieval from metadata server

Raphx:
AWS region is returned via dynamic data, on instance identity document
endpoint of the metadata server. This region data is used by AWS SDK in
determining which region to consume the service on.

Allowing region retrieval on aws-vault server mode allows the server to
behave slightly more closely to the actual metadata server, and aids in
local application development, by allowing the code to be more agnostic
of its operating environment.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -120,7 +120,7 @@ func StartIdentityDocumentServer(config *vault.Config) error {
 func identityDocHandler(config *vault.Config) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		err := json.NewEncoder(w).Encode(map[string]interface{}{
-			"region":          config.Region,
+			"region": config.Region,
 		})
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
--- a/server/server.go
+++ b/server/server.go
@@ -9,14 +9,17 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/99designs/aws-vault/vault"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 )
 
 const (
-	metadataBind    = "169.254.169.254:80"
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -9,14 +9,17 @@ import (
...
 )
 
 const (
-	metadataBind    = "169.254.169.254:80"
-	awsTimeFormat   = "2006-01-02T15:04:05Z"
-	localServerURL  = "http://127.0.0.1:9099"
-	localServerBind = "127.0.0.1:9099"
+	metadataBind               = "169.254.169.254:80"
+	awsTimeFormat              = "2006-01-02T15:04:05Z"
+	identityDocumentServerURL  = "http://127.0.0.1:9098"
```
mtibben, line 19: Any reason we create a second server? Can we use the existing one?

Raphx: Hi @mtibben, thanks for the review!

The existing local server has one global handler registered to it, i.e.:

https://github.com/99designs/aws-vault/blob/9682762130ad5fbabeff5a1a39a445c21a7089c1/server/server.go#L90

If we were to use the existing local server, it would have to be adjusted to handle different routes. This means there would be two route handling logics, one on the metadata server listening on 169.254.169.254:80, and another on the local server 127.0.0.1:9099. Illustrated:

```
[metadata server]
(route logic)
     |
     |
     v
[local server]
(route logic) --> credential
     |
     |
     v
identity document
```

The approach in this PR concentrates route handling logic only on the metadata server listening on 169.254.169.254:80, and for the local server, it just acts as a catch all for any request coming in, i.e:

```
[metadata server]
(route logic) --> [local server] --> credential
     |
     |
     v
[local server]
     |
     |
     v
identity document
```

I'm leaning towards the latter approach, therefore creating a second server. Definitely there is a refactoring opportunity here.

PR #547, opened 2020-03-26
Update README

sillyfellow:
add subsection to describe how to install using `go get`


mtibben:
This won't get you signed binaries on macOS, and there are subtle issues depending on which version of go is installed. Might wait for https://github.com/golang/go/issues/30515 to be resolved

PR #549, opened 2020-03-31
Add support for AWS SSO profiles

### LINKED ISSUE #449: Support AWS SSO
The AWS CLI v2 (beta) now has support for using AWS SSO. It would be great if aws-vault could support this too. The blog post is [here](https://aws.amazon.com/blogs/developer/aws-cli-v2-now-supports-aws-single-sign-on/). I will provide a more in-depth description of how this works in this issue.

## Expected behavior
If you use SSO with aws2, your config file willl look like this:

```
[profile AWSAdministratorAccess-123456789012]
sso_start_url = https://d-90xxxxxxxx.awsapps.com/start
sso_region = us-east-1
sso_account_id = 123456789023
sso_role_name = AWSAdministratorAccess
region = eu-west-1
output = json
```
The AWS CLI has a wizard to generate this, but I think we can assume that the user either will follow this wizard, or will get the config from someone else.

The flow is as follows:
1. The aws-vault will read this config and redirect the user to a browser if he does not have a cached and valid access token (see below for how to generate the redirect urls)
2. The user logs in to SSO and allows aws-vault to login (there is  no redirect back to aws-vault)
3. aws-vault queries sso-oidc to get an access token after the approval (this can be done in a loop)
4. the access token can be exchanged for access key / secret key

## Testing this out
I tried to do the same steps as the aws2 cli, but step by step. This is how that goes, and what I found:

### 1. Register a client
This will generate a client_id and a client_secret for aws-vault to use. AWS expects that this will be cached and long-lived.
The request is unauthenticated and the region most be the `sso_region`.

```bash
aws sso-oidc register-client --client-name aws-vault-test --client-type public --region $sso_region --no-sign-request
{
    "clientId": "REDACTED",
    "clientSecret": "REDACTED",
    "clientIdIssuedAt": 1573304968,
    "clientSecretExpiresAt": 1581080968
}
```
Some notes:
- clientSecretExpiresAt is 90 days in the future
- clientSecret is a JWT token with payload that looks like `{"serialized": "...a-serialized-json-dictionary..."}`. If you deserialize the dictionary, it looks like this:
```json
{
  "expired": false,
  "clientId": {
      "value": "REDACTED"
  },
  "clientName": "aws-vault-test",
  "clientType": "PUBLIC",
  "templateArn": null,
  "templateContext":null,
  "expirationTimestamp": 1581080968.984000000,
  "createdTimestamp":1573304968.984000000,"
  updatedTimestamp":1573304968.984000000,
  "createdBy": null,
  "updatedBy": null
}
```

### Step 2: Redirect the user to a verification URI

(note: If you use the CLI, dissable the resolving of URLS with `aws configure set cli_follow_urlparam false`)

The user has to enter a verifiaction code on a verification uri. We can also combine these two steps by adding the verification code in a query parameter. AWS will return us both options if we call `start-device-authorization`

```
aws sso-oidc start-device-authorization --client-id $client_id --client-secret $client_secret --start-url $sso_start_url --region $sso_region --no-sign-request
{
    "deviceCode": "REDACTED",
    "userCode": "HZBR-VSPV",
    "verificationUri": "https://device.sso.us-east-1.amazonaws.com/",
    "verificationUriComplete": "https://device.sso.us-east-1.amazonaws.com/?user_code=HZBR-VSPV",
    "expiresIn": 600,
    "interval": 1
}
```
The user has to be redirected to `verificationUriComplete` and aws-vault needs to safe `deviceCode`in memory.

### Step 3: User signs in:
login
approve
approved

I assume the "AWS CLI" text at some point will be changed to use the ClientName from the first api call.

### Step 4: Getting Credentials
Once the User has pushed approve, the deviceCode can be exchanged for an access token. The CLI retries this until it does, so no user interaction is needed (after expiresIn seconds the loop can break, as it will never succeed later).

```bash
aws sso-oidc create-token --client-id $client_id --client-secret $client_secret --grant-type urn:ietf:params:oauth:grant-type:device_code --device-code $device_code --region us-east-1 --no-sign-request
{
    "accessToken": "REDACTED",
    "tokenType": "Bearer",
    "expiresIn": 28800
}
```
I found the grant type in botocore. There is also a `--scope` parameter and botocore has a default scope, but it seems not used. If it fails without, use `sso-portal:*`.

The accessToken is again a JWT. The payload has been encrypted by KMS, so it's  not direct usable.

```bash
echo $payload | base64 --decode | strings
H"'~7"
DataPlaneSession
Peregrine
aws-kms
Karn:aws:kms:us-east-1:276787839696:key/4603c12d-5f0b-4eb4-b7c9-ccb048912c24
�;&;'
0=QOC
]Hi`
```

The accessToken can be cached and reused to get credentials for all roles with the same sso_start_url and sso_region configuration.

```
aws sso get-role-credentials --role-name $sso_role_name --account-id $sso_account_id --access-token $access_token --region $sso_region --no-sign-request
{
    "roleCredentials": {
        "accessKeyId": "REDACTED",
        "secretAccessKey": "REDACTED",
        "sessionToken": "REDACTED",
        "expiration": 1573311707000
    }
}
```

## References:

- [change in botocore](https://github.com/boto/botocore/commit/06d6a4464cf3950bd27f1d5077781913602fdff9)
- [change in the aws-cli](https://github.com/aws/aws-cli/pull/4627/files)
- [blog post](https://aws.amazon.com/blogs/developer/aws-cli-v2-now-supports-aws-single-sign-on/)
- [aws sso-oidc cli documentation](https://docs.aws.amazon.com/cli/latest/reference/sso-oidc/)

itsdalmo:
This PR hopefully closes #449 by adding support for AWS SSO.

Given a [AWS SSO Profile configuration](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html) in `~/.aws/confg`:
```ini
[profile Administrator-123456789012]
sso_start_url=https://aws-sso-portal.awsapps.com/start
sso_region=eu-west-1
sso_account_id=123456789012
sso_role_name=Administrator
```

This PR allows `aws-vault` to obtain short lived credentials for an SSO profile, which can be used by e.g. `exec` (and `login`), and produces output as seen below:

```shell
$ aws-vault exec Administrator-123456789012 -- aws sts get-caller-identity

Attempting to automatically open the SSO authorization page in your default
browser. If the browser does not open or you wish to use a different device to
authorize this request, open the following URL:

https://device.sso.eu-west-1.amazonaws.com/?user_code=HZZB-FPRL
(Use Ctrl-C to abort)

{
    "UserId": "AROAWH...",
    "Account": "123456789012",
    "Arn": "arn:aws:sts::123456789012:assumed-role/AWSReservedSSO_ReadOnly_f4296a2ab1f54217/..."
}
```

In the above example, `aws-vault` would take care of:
1. Creating and registering a new AWS SSO client (one per unique `sso_start_url` in your config).
2. Initiating the client authorization flow in AWS SSO (and opening the default browser for the user).
3. Creating a new access token after the client has been authorised by the user.
4. Creating a new temporary session for the `exec` command.

The client credentials (ID and secret), access token and session credentials will all be cached in the Keyring, and automatically refreshed (by repeating the relevant steps above) when they expire. The AWS SSO Client is valid for 3 months by default, while the access token and session expiration depends on the settings for your AWS SSO Organization.

Please take a look and let us know what you think 😅

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -222,6 +222,25 @@ role_arn = arn:aws:iam::123456789012:role/target
 
 You can also set the `mfa_serial` with the environment variable `AWS_MFA_SERIAL`.
 
+## AWS Single Sign-On (AWS SSO)
+
+The AWS CLI can [generate the SSO profile configuration](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html), but it's also possible to directly input this information in your `~/.aws/config` file. The configuration options are as follows:
+* `sso_start_url` The URL that points to the organization's AWS SSO user portal.
+* `sso_region` The AWS Region that contains the AWS SSO portal host. This is separate from, and can be a different region than the default CLI region parameter.
+* `sso_account_id` The AWS account ID that contains the IAM role that you want to use with this profile.
+* `sso_role_name` The name of the IAM role that defines the user's permissions when using this profile.
+
+### Example ~/.aws/config
+
+Here is an example `~/.aws/config` file, to help show the configuration for use with AWS SSO.
+
+```ini
+[profile Administrator-123456789012]
+sso_start_url=https://aws-sso-portal.awsapps.com/start
+sso_region=eu-west-1
+sso_account_id=123456789012
+sso_role_name=Administrator
+```
 
 ## Removing stored sessions
 
--- a/vault/ssorolecredentialsprovider.go
+++ b/vault/ssorolecredentialsprovider.go
@@ -1,0 +1,257 @@
+package vault
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/99designs/keyring"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/service/sso"
+	"github.com/aws/aws-sdk-go/service/ssooidc"
+	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/skratchdot/open-golang/open"
+)
+
+const (
+	ssoClientName         = "aws-vault"
+	ssoClientType         = "public"
+	oAuthTokenGrantType   = "urn:ietf:params:oauth:grant-type:device_code"
+	authorizationTemplate = `
+Attempting to automatically open the SSO authorization page in your default
+browser. If the browser does not open or you wish to use a different device to
+authorize this request, open the following URL:
+
+%s
+(Use Ctrl-C to abort)
+
+`
+)
+
+// CachedSSORoleCredentialsProvider uses the keyring to cache SSO Role sessions.
+type CachedSSORoleCredentialsProvider struct {
+	CredentialsName string
+	Keyring         *CredentialKeyring
+	Provider        *SSORoleCredentialsProvider
+	ExpiryWindow    time.Duration
+	credentials.Expiry
+}
+
+// Retrieve the cached credentials or generate new ones.
+func (p *CachedSSORoleCredentialsProvider) Retrieve() (credentials.Value, error) {
+	sessions := p.Keyring.Sessions()
+
+	session, err := sessions.Retrieve(p.CredentialsName, "")
+	if err != nil {
+		// session lookup missed, we need to create a new one.
+		session, err = p.Provider.GetRoleCredentials()
+		if err != nil {
+			return credentials.Value{}, err
+		}
+
+		err = sessions.Store(p.CredentialsName, "", session)
+		if err != nil {
+			return credentials.Value{}, err
+		}
+	} else {
+		log.Printf("Re-using cached credentials %s generated from GetRoleCredentials, expires in %s", FormatKeyForDisplay(*session.AccessKeyId), time.Until(*session.Expiration).String())
+	}
+
+	p.SetExpiration(*session.Expiration, p.ExpiryWindow)
+
+	return credentials.Value{
+		AccessKeyID:     *session.AccessKeyId,
+		SecretAccessKey: *session.SecretAccessKey,
+		SessionToken:    *session.SessionToken,
+	}, nil
+}
+
+// SSORoleCredentialsProvider creates temporary credentials for an SSO Role.
+type SSORoleCredentialsProvider struct {
+	OIDCProvider *SSOOIDCProvider
+	SSOClient    *sso.SSO
+	AccountID    string
+	RoleName     string
+	ExpiryWindow time.Duration
+	credentials.Expiry
+}
+
+// Retrieve generates a new set of temporary credentials using SSO GetRoleCredentials.
+func (p *SSORoleCredentialsProvider) Retrieve() (credentials.Value, error) {
+	creds, err := p.GetRoleCredentials()
+	if err != nil {
+		return credentials.Value{}, err
+	}
+
+	p.SetExpiration(*creds.Expiration, p.ExpiryWindow)
+	return credentials.Value{
+		AccessKeyID:     *creds.AccessKeyId,
+		SecretAccessKey: *creds.SecretAccessKey,
+		SessionToken:    *creds.SessionToken,
+	}, nil
+}
+
+func (p *SSORoleCredentialsProvider) GetRoleCredentials() (*sts.Credentials, error) {
+	token, err := p.OIDCProvider.GetAccessToken()
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := p.SSOClient.GetRoleCredentials(&sso.GetRoleCredentialsInput{
+		AccessToken: aws.String(token.Token),
+		AccountId:   aws.String(p.AccountID),
+		RoleName:    aws.String(p.RoleName),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	expiration := aws.MillisecondsTimeValue(resp.RoleCredentials.Expiration)
+
+	// This is needed because sessions.Store expects a sts.Credentials object.
+	creds := &sts.Credentials{
+		AccessKeyId:     resp.RoleCredentials.AccessKeyId,
+		SecretAccessKey: resp.RoleCredentials.SecretAccessKey,
+		SessionToken:    resp.RoleCredentials.SessionToken,
+		Expiration:      aws.Time(expiration),
+	}
+
+	log.Printf("Got credentials %s for SSO role %s (account: %s), expires in %s", FormatKeyForDisplay(*resp.RoleCredentials.AccessKeyId), p.RoleName, p.AccountID, time.Until(expiration).String())
+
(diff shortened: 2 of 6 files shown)


vault/ssorolecredentialsprovider.go
```diff
@@ -0,0 +1,257 @@
...
+
+	fmt.Printf(authorizationTemplate, aws.StringValue(auth.VerificationUriComplete))
+
+	if err := open.Run(aws.StringValue(auth.VerificationUriComplete)); err != nil {
+		log.Printf("failed to open browser: %s", err)
+	}
+
+	for {
+		// Sleep to allow the user to complete the login flow
+		time.Sleep(3 * time.Second)
```
statik, line 237: I wonder about this delay, should it be configurable? I am not very fast on the keyboard, it could definitely take me longer than 3 seconds to enter my credentials in AWS SSO via my 3rd party IdP, including finding my MFA device and completing the authentication. Or is it that the user should already be logged into the browser ahead of time and this time is for clicking a single approve button?

statik:
I've tested this branch with my AWS SSO configured to use a 3rd Party IdP (Google Apps SAML), and it works well so far 😍

vault/ssorolecredentialsprovider.go
rickardl: Just to clarify, retries every 3 seconds. If you have a SSO session against your IDP, it doesn’t require you to re-authenticate with the MFA, so waiting longer might be annoying as it’s almost instant after the first initial authentication.

itsdalmo: I did some digging around to see how the AWS CLI does this, and ended up here: https://github.com/boto/botocore/blob/v2/botocore/utils.py#L1663

And from their comments around the default retry interval and slow down delay:
https://tools.ietf.org/html/draft-ietf-oauth-device-flow-15#section-3.5

I.e., we (like the AWS CLI) should just be using the interval that is specified in the response from the API when calling `StartDeviceAuthorization`, and fall back to the 5s defaults defined by the RFC. https://github.com/99designs/aws-vault/pull/549/commits/d7ce500b1aed44ed9d4d2d3dbe649c035ddaa747 implements this, and also adds handling for the `SlowDownException` error code.

From my testing, this has effectively reduced the retry interval to 1s, which makes AWS Vault more snappy to respond after I had completed the browser login.

itsdalmo:
Hi @mtibben! Sorry to ping you like this; I was wondering when you might have time to look at this PR (since you seem to be the one maintaining `aws-vault` these days) - do you need anything else from us before reviewing and/or merging the PR?

everops-kevin:
@itsdalmo you'll end up in an infinite loop with `aws-vault login SSOprofile`

vault/ssorolecredentialsprovider.go
```diff
@@ -0,0 +1,272 @@
...
+	auth, err := p.OIDCClient.StartDeviceAuthorization(&ssooidc.StartDeviceAuthorizationInput{
+		ClientId:     aws.String(creds.ID),
+		ClientSecret: aws.String(creds.Secret),
+		StartUrl:     aws.String(p.StartURL),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf(authorizationTemplate, aws.StringValue(auth.VerificationUriComplete))
```
mtibben, line 229: If we dump this to stdout with `fmt.Printf` it may interfere with the output of whatever command you're executing... can we get this output to stderr, preferably through kingpin somehow? Alternatively should we be using the `prompt` driver to achieve this?

vault/ssorolecredentialsprovider.go
```diff
@@ -0,0 +1,272 @@
...
+		AccountId:   aws.String(p.AccountID),
+		RoleName:    aws.String(p.RoleName),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	expiration := aws.MillisecondsTimeValue(resp.RoleCredentials.Expiration)
+
+	// This is needed because sessions.Store expects a sts.Credentials object.
```
mtibben, line 114: yes I'm hoping to refactor sessions.Store completely and turn it into an ordinary `Provider`

vault/ssorolecredentialsprovider.go
```diff
@@ -0,0 +1,272 @@
...
+			case ssooidc.ErrCodeAuthorizationPendingException:
+				time.Sleep(retryInterval)
+				continue
+			default:
+				return nil, err
+			}
+		}
+		return &SSOAccessToken{
+			Token:      aws.StringValue(t.AccessToken),
+			Expiration: time.Now().Add(time.Duration(aws.Int64Value(t.ExpiresIn)) * time.Second),
```
mtibben, line 269: Can we break this up to make it more readable?

USAGE.md
```diff
@@ -222,6 +222,25 @@ role_arn = arn:aws:iam::123456789012:role/target
 
 You can also set the `mfa_serial` with the environment variable `AWS_MFA_SERIAL`.
 
+## AWS Single Sign-On (AWS SSO)
```
mtibben, line 225: Add this to the TOC at the top

vault/ssorolecredentialsprovider.go
```diff
@@ -0,0 +1,272 @@
...
+authorize this request, open the following URL:
+
+%s
+(Use Ctrl-C to abort)
+
+`
+)
+
+// CachedSSORoleCredentialsProvider uses the keyring to cache SSO Role sessions.
+type CachedSSORoleCredentialsProvider struct {
```
mtibben, line 35: Hmm I was planning a sessions refactor, but it may need to encompass this. Mind if I ping you if I get to this?

vault/ssorolecredentialsprovider.go
```diff
@@ -0,0 +1,272 @@
...
+		// These are the default values defined in the following RFC:
+		// https://tools.ietf.org/html/draft-ietf-oauth-device-flow-15#section-3.5
+		slowDownDelay = 5 * time.Second
+		retryInterval = 5 * time.Second
+	)
+	if i := aws.Int64Value(auth.Interval); i > 0 {
+		retryInterval = time.Duration(i) * time.Second
+	}
+
+	for {
```
mtibben, line 244: Would it be possible to capture the retry logic in its own separate function? There's an example of this in https://github.com/99designs/aws-vault/blob/877981b40a4614bf5cd921ed63eac96a436b4c97/cli/rotate.go#L149

mtibben:
This is fantastic, thank you @itsdalmo!

I know from experience that testing this stuff is really really hard as you basically have to mock out everything. But if there are any simple tests you can write to prevent me breaking this going forward that would be ideal

vault/ssorolecredentialsprovider.go
itsdalmo: Great catch! Since `prompt` is only used for requesting user input (which we are not doing here), and `kingpin` is not currently used anywhere outside the `cli/` page right now, I think it makes most sense to use `Fprintf(os.Stderr, ...)` here.

glenjamin:
Does anyone have this working with `credentials_process`?

PR #556, opened 2020-04-18
Add ECS server

mtibben:
Rebases #375

AWS SDKs universally support the ECS Credential Service, which supports
reading the service URI from environment variables.

Providing an http interface allows the subprocess to refresh credentials
as long as the master credentials session is valid.

Supporting the ECS Credential provider offers the following advantages
over the EC2 Metadata provider:

- Binding to a random, ephimeral port
  - Does not require adminstrator privileges
  - Allows multiple providers simultaneously for discrete processes
  - Partially mitigates the security issues that accompany the EC2
    Metadata Service because the address is not well-known
- Requiring an Authorization token further mitigates the potential for
  another process to access the credentials, since the Authorization
  token is only exposed to the subprocess via environment variables

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -26,6 +26,7 @@ type ExecCommandInput struct {
 	Args             []string
 	Keyring          keyring.Keyring
 	StartEc2Server   bool
+	StartEcsServer   bool
 	CredentialHelper bool
 	Config           vault.Config
 	SessionDuration  time.Duration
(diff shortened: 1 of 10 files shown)


mtibben:
Thank you @jstewmon, this is way better than the current ec2 server. Next question, should we make `--ecs-server` the default, and how should we role that out?

PR #558, opened 2020-04-20
Yubikey support

mtibben:
Support for Yubikey OATH-TOTP via ykman.

Thanks to @j0hnsmith for the initial PRs at #316, #514, #533

Could someone with a Yubikey please test this? @j0hnsmith @jtopper or anyone else? 
[Docs](https://github.com/99designs/aws-vault/blob/ykman-support/USAGE.md#using-a-yubikey) and [binaries](https://github.com/99designs/aws-vault/releases/tag/v5.5.0-alpha2)

PR #559, opened 2020-04-21
Add contrib script to enable a mfa device via ykman

asiragusa:
This script helps to add a yubikey using ykman. See https://github.com/99designs/aws-vault/pull/558

### DIFF
--- a/contrib/enable-mfa-device-ykman.sh
+++ b/contrib/enable-mfa-device-ykman.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+
+set -e
+
+if [ -n "$AWS_SESSION_TOKEN" ]; then
+  echo "aws-vault must be run without a STS session, please run it with the --no-session flag" >&2
+  exit 1
+fi
+
+cleanup()
+{
+  if [ -z "$OUTFILE" ]; then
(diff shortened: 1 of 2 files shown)


contrib/enable-mfa-device-ykman.sh
```diff
@@ -0,0 +1,43 @@
...
+  --outfile "$OUTFILE" \
+  --query VirtualMFADevice.SerialNumber \
+  --output text)
+
+ykman oath add -t "$SERIAL_NUMBER" < "$OUTFILE" 2> /dev/null
+rm "$OUTFILE"
+
+CODE1=$(ykman oath code -s "$SERIAL_NUMBER")
+
+echo "Waiting 30s before generating a second code" >&2
```
christophgysin, line 34: How about just running the command until we get a new code? 30s is the worst case.

asiragusa: It is another option, but having to touch the Yubikey multiple times can be tedious too. 

The low-level driver allows to query the key with a timestamp, but ykman does not have that option...

christophgysin: Ah, of course. Thanks for the explanation.

PR #563, opened 2020-04-21
Add unit tests for AWS SSO provider

rickardl:
Hi @mtibben,

Sorry for the late response on your feedback, this PR adds the missing unit-tests as highlighted in your review of #549 . 

Kind regards
Rickard


mtibben:
I am wary that these kind of mocks/fakes can be quite brittle, but as it's nicely contained to SSORoleCredentialsProvider that shouldn't be an issue

PR #565, opened 2020-04-22
Stream errors to stderr

arkadiybutermanov:
This is a follow-up for my comment here: https://github.com/99designs/aws-vault/pull/433/files#issuecomment-617656861

/cc @mtibben / @pbitty

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,8 +20,7 @@ func run(args []string, exit func(int)) {
 		`A vault for securely storing and accessing AWS credentials in development environments.`,
 	)
 
-	app.ErrorWriter(os.Stderr)
-	app.Writer(os.Stdout)
+	app.UsageWriter(os.Stdout)
 	app.Version(Version)
 	app.Terminate(exit)
 


main.go
```diff
@@ -20,8 +20,7 @@ func run(args []string, exit func(int)) {
 		`A vault for securely storing and accessing AWS credentials in development environments.`,
 	)
 
-	app.ErrorWriter(os.Stderr)
```
arkadiybutermanov, line 23: This is already set to `stderr` by default: https://github.com/alecthomas/kingpin/blob/b6657d9477a694e178dcba813954013bfd2f0346/app.go#L56

main.go
```diff
@@ -20,8 +20,7 @@ func run(args []string, exit func(int)) {
...
-	app.Writer(os.Stdout)
```
arkadiybutermanov, line 24: This was overriding the previously set value and is deprecated: https://github.com/alecthomas/kingpin/blob/b6657d9477a694e178dcba813954013bfd2f0346/app.go#L133

main.go
```diff
@@ -20,8 +20,7 @@ func run(args []string, exit func(int)) {
...
+	app.UsageWriter(os.Stdout)
```
arkadiybutermanov, line 23: We probably want `--help`  and `--version` to output to `stdout` (since we don't use `app.FatalUsage` anywhere). By default it outputs to `stderr`: https://github.com/alecthomas/kingpin/blob/b6657d9477a694e178dcba813954013bfd2f0346/app.go#L57

mtibben:
Makes sense. Thanks @arkadiybutermanov

pbitty:
Good catch @arkadiybutermanov .  Thanks for fixing it.

PR #570, opened 2020-04-30
Revert "Ignore case of section and key names when parsing config."

### LINKED ISSUE #528: Mixed-case profile names are valid, but unrecognized
- [x] I am using the latest release of AWS Vault
- [ ] I have provided my `.aws/config` (redacted if necessary)

This was a result of merging [PR#461](https://github.com/99designs/aws-vault/pull/461) which in my opinion was ill-advised.

The whole motivation for the change was to make it possible for a specific class of typographical error accepted by `aws-vault`. Considering this change breaks existing configurations for teams like mine, that doesn't seem like a strong enough justification.

Profile names are case-sensitive in AWS SDKs. It only ini *keys* which should be case-insensitive.

Thus, it would be fine to lowercase the keys, I guess. But not the ini section names (which are the profile names). Unfortunately [PR#461](https://github.com/99designs/aws-vault/pull/461) was a little too broad in that sense. I would propose either:

1. Don't do any insensitivity. People should expect that invalid keys won't work.
1. Or, improve this changeset by only treating keys insensitively. It's going to be more code, but more correct.

mtibben:
Reverts 99designs/aws-vault#461 to resolve #528

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -116,7 +116,6 @@ func (c *ConfigFile) parseFile() error {
 	log.Printf("Parsing config file %s", c.Path)
 	f, err := ini.LoadSources(ini.LoadOptions{
 		AllowNestedValues: true,
-		Insensitive:       true,
 	}, c.Path)
 	if err != nil {
 		return fmt.Errorf("Error parsing config file %q: %v", c.Path, err)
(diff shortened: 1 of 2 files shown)


statik:
Thank you! looking forward to this fix

Issue #545, opened 2020-03-25
simple bug and am close (hopefully 😁) to making PR: login command hangs if profile doesn't exist


matthewliuswims:
- [x] I am using the latest release of AWS Vault
- [x] I have provided my `.aws/config` (redacted if necessary)
- [x] I have provided the debug output using `aws-vault --debug` (redacted if necessary)

# bug/replication
`aws-vault login <non-existent-profile>` hangs forever if profile doesn't exist 😓

# fix attempt
I did some digging: the hang is caused by infinite recursion here https://github.com/99designs/aws-vault/blob/1ce365597489943c2894d3265280933af28de102/vault/vault.go#L220-L231 

```
hasMasterCreds, err := keyring.Has(profileName)  // the culprit
// hasMasterCreds, err end up being false and nil respectively forever and ever
```

func is called here 
https://github.com/99designs/aws-vault/blob/1ce365597489943c2894d3265280933af28de102/vault/vault.go#L196-L197

I'm happy to make a pr, but i'm not sure what kind of base case checking I should be doing for `MasterCredentialsFor` func without breaking functionality 

Thank you!

mtibben:
A PR would certainly be welcome! Get something going and we can discuss

mtibben:
How's that PR coming along @matthewliuswims? :)

mtibben:
Possible solution here? https://github.com/99designs/aws-vault/issues/507#issuecomment-577427034

matthewliuswims:
ah yes - that looks like a great fix. :) gonna do some testing and i'll hopefully put in a PR today!

matthewliuswims:
@mtibben PR here https://github.com/99designs/aws-vault/pull/575/files !

mtibben:
Fixed in 1496990

Issue #507, opened 2020-01-22
Infinite recursion checking macOS keychain


jameshiew:
Version: v5.2.0
on macOS 10.15.2

Running `aws-vault --debug login default` results in infinite logs like

```
2020/01/22 22:57:17 [keyring] Querying keychain for service="aws-vault", keychain="aws-vault.keychain"
2020/01/22 22:57:17 [keyring] Found 0 results
```

Without the `--debug` flag the program appears to be hanging.

I have a `default` profile in my `.aws/config` file, but the same issue can occur if a non-existent profile string is used.

jameshiew:
Adding an error if empty string for the `keyring.CredentialKeyring.Has(string)` method prevents the infinite loop.

```go
func (ck *CredentialKeyring) Has(credentialsName string) (bool, error) {
	if credentialsName == "" {
		return false, fmt.Errorf("empty credential name")
	}
```

`vault.MasterCredentialsFor` can end up calling that func over and over with an empty string. I don't have enough context though to know if handling it with an error like above is appropriate though.

mtibben:
dupe #545

PR #575, opened 2020-05-01
Fix infinite recursion issue for missing/invalid profile

matthewliuswims:
- happens when invalid profile is passed to login command
- did some light tweaks to README

## issues
https://github.com/99designs/aws-vault/issues/545
https://github.com/99designs/aws-vault/issues/507#issuecomment-577427034

## testing
manually from CLI

## credits 
fix is from https://github.com/99designs/aws-vault/issues/507#issuecomment-577427034 - just changed error message to be more generic

## notes
- i don't know this code base, so unintended consequences are totally possible so 👀 would be appreciated
- error message IMO shouldn't say "empty credential name" (even though it's more specific) because it'll then bubble up that as the error message for invalid profiles

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,17 +145,23 @@ sso_role_name=Administrator
 
 This profile should work expected with AWS Vault commands, e.g. `exec` and `login`. See [Basic Usage](#basic-usage) for more information.
 
-## Development
+## Local Development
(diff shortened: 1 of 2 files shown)


mtibben:
Thanks @matthewliuswims

Issue #408, opened 2019-08-17
Bash completion does not work correctly


hoegertn:
The command `aws-vault --completion-bash` works perfectly for the first argument, but it does not work for the second argument.

This is correct:

```bash
$ aws-vault --completion-bash "exec"
profile1
foobar
blubb
```

But this does not work:

```bash
$ aws-vault --completion-bash "exec pr"
help
add
list
rotate
exec
remove
login
```

It should autocomplete to `profile1`

mtibben:
Hmm perhaps this issue is in kingpin (the library providing this feature)?

tkalus:
Definitely an issue with kingpin.  It appears to have been fixed with alecthomas/kingpin#212, but it _also_ appears that there hasn't been a v2 release to `gopkg.in` since 2017.  Kingpin.v2 seems stuck in limbo with consumers pulling from the upstream `master` or `v3-unstable`.

@mtibben Do you want a new thread/issue to cover this discussion?

PR #576, opened 2020-05-02
Fix for 99designs/aws-vault#408

tkalus:
Issue in kingpin cmdline parser around parsing partial shell completions.
Seems to have been mostly-partially fixed with alecthomas/kingpin#212, but never
pushed to gopkg.in.

This change moves to github.com, pinning to `master` branch/commit
at time of writing/test.

Also, fixes issue in `zsh` contributed completion; words want to start
at the second param and not be quoted.

There's still an issue if one has a profile name that's a substring of
a longer name (E.G. `one` and `one-ro`); will consider the matched
substring a full and unique match.  Completing the example, if two
profiles are named `one-rw` and `one-ro`, it will continue to match
both.


tkalus:
Additional remarks showing both the fix and the remaining issue:

Demonstrated fix:
```
$ grep profile ~/.aws/config
[profile one-two]
[profile one-tre]
$ ./aws-vault-darwin-amd64 --completion-bash exec one-t
one-two
one-tre
$ ./aws-vault-darwin-amd64 --completion-bash exec one-tw
one-two
$ ./aws-vault-darwin-amd64 --completion-bash exec one-two
$
```

Still a problem:
```
$ grep profile ~/.aws/config
[profile one]
[profile one-two]
$ ./aws-vault-darwin-amd64 --completion-bash exec on
one
one-two
$ ./aws-vault-darwin-amd64 --completion-bash exec one
$ ./aws-vault-darwin-amd64 --completion-bash exec one-
one-two
$
```
(completion considered `one` to be a fully-unique match and had no output)

mtibben:
Thanks @tkalus commit is at 75a69719c073e7ed4a846d54039b4f10a3add6d0

PR #580, opened 2020-05-05
Check the host header to mitigate a DNS rebinding attack

### LINKED ISSUE #578: aws-vault server should require `Host: 169.254.169.254` header
tl;dr:  With the aws-vault EC2 metadata service running, it's possible for an attacker controlling any web page the victim visits with javascript enabled to mount a [DNS rebinding attack](https://en.wikipedia.org/wiki/DNS_rebinding) to grab short-lived credentials from the aws-vault server.

If it were to require an explicit `Host: 169.254.169.254` header, that attack would be mitigated.

## Setup

Run `aws-vault exec --server <profile>`.  It creates an alias IP 169.254.169.254 on the loopback interface, starts a proxy bound to 169.254.169.254:80 that just proxies to http://127.0.0.1:9099/, and starts an http server listening on localhost:9099 that exposes short-lived credentials, similar to how the EC2 metadata service does.

Put the following HTML somewhere where you can serve it with plain http on port 80:

```html
<html>
    <body>
	<script type="text/javascript">
	 var xhr = new XMLHttpRequest();
	 xhr.onload = function() {
	     console.log(xhr.responseText);
	 }
	 xhr.open("GET", "http://www.example.org/latest/meta-data/iam/security-credentials/local-credentials");
	</script>
	<input type="button" onclick="xhr.send(null);" value="click me">
    </body>
</html>
```

Let's say you're serving that on http://1.2.3.4:80/index.html.  Add a line to your /etc/hosts `1.2.3.4 www.example.org`, flush all your DNS caches (`sudo killall -HUP mDNSResponder` on mac, "Clear host cache" button on chrome://net-internals#dns, whatever).  Load that page in Chrome by going to http://www.example.org/index.html .

Edit your /etc/hosts line to be `169.254.169.254 www.example.org`, flush all your DNS caches, then click the "click me" button on index.html.

### Result

Chrome re-resolves the name www.example.org to 169.254.169.254, allows the XHR to that IP, and logs the short-lived credential to the console.  In a real attack, the attacker would exfiltrate it.

Although I used a button in the above PoC, nothing about this attack requires the victim to click or interact with anything;  as long as they load a malicious page and have javascript enabled it can work.  A typical DNS rebinding attack uses DNS records with short TTL (and a DNS server that actively tries to work around things like Chrome's in-browser DNS cache).

## Mitigation

In the above example, Chrome sends a header `Host: www.example.org`.  If aws-vault's server required a header `Host: 169.254.169.254`, that would mitigate this attack.  Chrome will block  XHR to http://169.254.169.254/... from www.example.org because aws-vault doesn't send any CORS headers that would allow that access.

### Possible downsides

While botocore explicitly uses the literal http://169.254.169.254:80 during the `iam-role` step of the credentials chain provider, I haven't looked at all the SDKs to make sure they use that IP; it's possible they some DNS name that happens to resolve to 169.254.169.254 inside AWS's split-horizon DNS.

If that turns out to be the case, we could control this behavior (of requiring the header) with a command line switch that is on by default, and let the user decide whether to disable it (`--no-require-host-header` or something).

mtibben:
Fixes #578

With the aws-vault EC2 metadata service running, it's possible for an attacker controlling any web page the victim visits with javascript enabled to mount a [DNS rebinding attack](https://en.wikipedia.org/wiki/DNS_rebinding) to grab short-lived credentials from the aws-vault server.

This requires the http request Host to be 169.254.169.254 to mitigate the attack.

### DIFF
--- a/server/ec2.go
+++ b/server/ec2.go
@@ -15,6 +15,7 @@ import (
 
 const (
 	awsTimeFormat            = "2006-01-02T15:04:05Z"
+	ec2MetadataEndpointIP    = "169.254.169.254"
 	ec2MetadataEndpointAddr  = "169.254.169.254:80"
 	ec2CredentialsServerAddr = "127.0.0.1:9099"
 )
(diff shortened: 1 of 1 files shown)


mtibben:
How does this fix look @philsnow, are you able to confirm that this mitigates the attack?

PR #582, opened 2020-05-07
Get aws vault params from ~/.aws/config file

### LINKED ISSUE #537: Add support for aws-vault config file
Now you have to specify aws-vault parameters as args or environment variables

It would be convenient to be able to load parameters from config file

For example, my ~/.aws/config
```
[profile A]
region = eu-central-1
mfa_serial = arn:aws:iam::111111111111:mfa/user
credential_process = aws-vault --backend=pass --pass-prefix=aws-vault --prompt=osascript exec A --json
```
I would like to have `--backend=pass --pass-prefix=aws-vault --prompt=osascript` stored in some config file as defaults

olfway:
Load aws-vault parameters from the ~/.aws/config file

```
[default]
aws_vault_backend = pass
aws_vault_pass_prefix = aws-vault
aws_vault_prompt = osascript

```
Fixes #537 

@mtibben could you look pls? should it be done this way?

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -24,6 +24,8 @@ const (
 
 	defaultSectionName          = "default"
 	roleChainingMaximumDuration = 1 * time.Hour
+
+	defaultPromptMethod = "terminal"
 )
 
 // UseSession will disable the use of GetSessionToken when set to false
@@ -153,6 +155,7 @@ type ProfileSection struct {
 	SSOAccountID    string `ini:"sso_account_id,omitempty"`
 	SSORoleName     string `ini:"sso_role_name,omitempty"`
 	AWSVaultBackend string `ini:"aws_vault_backend,omitempty"`
+	AWSVaultPrompt  string `ini:"aws_vault_prompt,omitempty"`
 }
 
 func (s ProfileSection) IsEmpty() bool {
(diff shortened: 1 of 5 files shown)


vault/config.go
```diff
@@ -138,20 +140,26 @@ func (c *ConfigFile) parseFile() error {
...
+	SSOStartURL          string `ini:"sso_start_url,omitempty"`
+	SSORegion            string `ini:"sso_region,omitempty"`
+	SSOAccountID         string `ini:"sso_account_id,omitempty"`
+	SSORoleName          string `ini:"sso_role_name,omitempty"`
+	AWSVaultBackend      string `ini:"aws_vault_backend,omitempty"`
+	AWSVaultPrompt       string `ini:"aws_vault_prompt,omitempty"`
+	AWSVaultKeychainName string `ini:"aws_vault_keychain_name,omitempty"`
+	AWSVaultPassCmd      string `ini:"aws_vault_pass_cmd,omitempty"`
+	AWSVaultPassDir      string `ini:"aws_vault_pass_dir,omitempty"`
+	AWSVaultPassPrefix   string `ini:"aws_vault_pass_prefix,omitempty"`
```
mtibben, line 162: I'm not keen to allow global config changes per profile, this would add a new layer of complexity with different profiles using different backends altogether 😱 . 

Should we consider splitting this config out, perhaps use the `ini` library directly to get this instead of `vault/config`. And maybe we should consider be using the `plugin` feature of aws cli to store the config in a different section name altogether (rather than under the default profile)?

What are your thoughts?

olfway: Actually I also prefer not to store keyring parameters in aws profile config
Could you look, please, on #590 ?

mtibben:
superseded by #590

Issue #571, opened 2020-04-30
Restore OIDC token caching 


mtibben:
#569 removed tests and caching for the OIDC token, we should restore

/cc @itsdalmo @rickardl

mtibben:
@itsdalmo 
> The two levels of caching was there because the start_url is the same for multiple profiles as you say. What was being cached was a device ID and an access token that could be used to exchange for role credentials, and by removing the caching users will need to go through login-flow (opening the browser and clicking sign-in etc) every time the role credentials expire (typically 1h) instead of just when the access token expires (e.g. every 8h). The device ID and access tokens are cached in the v2 AWS CLI as well, so I think AWS expects these values to be "long lived".

Sounds like the OIDC is the credential with friction? We should probably just cache that then instead of the SSO credentials, because the SSO credentials can be generated on-demand?

flyinprogrammer:
> SSO credentials, because the SSO credentials can be generated on-demand?

In my, humble and still learning, opinion, we might want to cache both?

caching OIDC will mean not having to open a browser across multiple profiles that share the same sso_start_url - i.e. login

caching SSO credentials will mean lower latency when using the AWS API, just like today, because we can use those credentials until they expire.

PR #584, opened 2020-05-09
Implement clear subcommand to remove sessions

### LINKED ISSUE #412: "clear" alias for "--sessions-only"
It would be nice to have a `clear` alias for `--sessions-only` because running this

```
aws-vault remove work --sessions-only
```

can be dangerous if we forget the `--sessions-only`. It would be easier to do this.

```
aws-vault clear work
```
 
I store my keys in a password manager if I accidentally remove them from aws-vault.

My workaround now is to use these aliases

```
alias av="aws-vault"
alias avcw="av remove work --sessions-only"
```

olfway:
Fixes #412

### DIFF
--- a/cli/clear.go
+++ b/cli/clear.go
@@ -1,0 +1,27 @@
+package cli
+
+import (
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+func ConfigureClearCommand(app *kingpin.Application, a *AwsVault) {
+	input := RemoveCommandInput{}
+
+	cmd := app.Command("clear", "Removes sessions for the profile")
+
+	cmd.Arg("profile", "Name of the profile").
+		Required().
+		HintAction(a.MustGetProfileNames).
+		StringVar(&input.ProfileName)
+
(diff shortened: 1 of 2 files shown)


cli/clear.go
```diff
@@ -0,0 +1,27 @@
...
+
+func ConfigureClearCommand(app *kingpin.Application, a *AwsVault) {
+	input := RemoveCommandInput{}
+
+	cmd := app.Command("clear", "Removes sessions for the profile")
+
+	cmd.Arg("profile", "Name of the profile").
+		Required().
+		HintAction(a.MustGetProfileNames).
+		StringVar(&input.ProfileName)
```
mtibben, line 15: I'm not sure we should require a profile here. Maybe something like `aws-vault clear-cache` should clear all cached sessions or credentials for all profiles. (Note that the recent SSO addition means that there may be other types of cached credentials to clear, still to be implemented in #571). Thoughts?

olfway:
Another pull request was created, close this one

Issue #544, opened 2020-03-24
duration_seconds not honored


kennedy:
- [x] I am using the latest release of AWS Vault
- [x] I have provided my `.aws/config` (redacted if necessary)
- [x] I have provided the debug output using `aws-vault --debug` (redacted if necessary)

I have set `duration_seconds` to be 43200 seconds (12 hours) in my configuration, but mfa/sts sessions are only 1 hour. 

I have currently tested this on:
- windows10 powershell.core 7
- windows10 powershell.core 6
- macOS 10.15.3 zsh

According to the USAGE.md page, I can override the `exec` command's sts session request by modifying the env variables, but that didnt work either. I even set all four possible `*_TTL` variable and it still gives me 1 hour. 

config (redacted):
```
[default]
duration_seconds = 43200

[profile testaccount]
mfa_serial = arn:aws:iam::xxxxx:mfa/kennedy
region = us-east-1
duration_seconds = 43200
```

debug (redacted):
```
PS C:\Users\kennedy\Documents\projects\test\terraform> echo $env:AWS_ASSUME_ROLE_TTL
43200
PS C:\Users\kennedy\Documents\projects\test\terraform> echo $env:AWS_CHAINED_SESSION_TOKEN_TTL
43200
PS C:\Users\kennedy\Documents\projects\test\terraform> echo $env:AWS_FEDERATION_TOKEN_TTL
43200
PS C:\Users\kennedy\Documents\projects\test\terraform> echo $env:AWS_SESSION_TOKEN_TTL
43200
PS C:\Users\kennedy\Documents\projects\test\terraform> aws-vault --debug exec testaccount -- terraform
2020/03/24 16:17:50 aws-vault v5.3.2
2020/03/24 16:17:50 [keyring] Considering backends: [wincred pass file]
2020/03/24 16:17:50 Loading config file C:\Users\kennedy\.aws\config
2020/03/24 16:17:50 Parsing config file C:\Users\kennedy\.aws\config
2020/03/24 16:17:50 profile testaccount: using stored credentials
2020/03/24 16:17:50 profile testaccount: using GetSessionToken (with MFA)
2020/03/24 16:17:50 Looking for sessions for testaccount
2020/03/24 16:17:50 Looking up all keys in keyring
Enter token for arn:aws:iam::XXXXX:mfa/kennedy: XXXXX
2020/03/24 16:17:59 Looking up keyring for 'testaccount'
[...]

PR #585, opened 2020-05-11
Make session duration format clear in usage docs

robinmitra:
The format of the session duration isn't quite clear, but https://github.com/99designs/aws-vault/issues/544#issuecomment-607581197 explains it well. Adding it to the docs to make it clear for others.


mtibben:
Thanks @robinmitra

PR #586, opened 2020-05-12
Improve fish shell completions

liamdawson:
I figure this should make it easier for the maintainers to keep `fish` supported (via `aws-vault --completion-bash`), and I've found the `aws-vault exec ... --` completions invaluable day-to-day:

Fish shell completions showing s3 subcommands for the command aws-vault exec a-role -- aws s3

### DIFF
--- a/contrib/completions/fish/aws-vault.fish
+++ b/contrib/completions/fish/aws-vault.fish
@@ -1,17 +1,24 @@
-complete -c aws-vault -x -a '(__fish_aws_vault_completion)'
+if status --is-interactive
+  complete -ec aws-vault
 
-function __fish_aws_vault_completion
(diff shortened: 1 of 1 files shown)


mtibben:
I don't use Fish but @zubin (as the original contributor in #422) do you want to review?

zubin:
> I don't use Fish but @zubin (as the original contributor in #422) do you want to review?

Sure, I'll give it a burl soon.

zubin:
The part after the `--` doesn’t seem to work for me. It’s just returning local filesystem, eg:

```
aws-vault exec <role> -- [tab shows `ls`]
```

I'm on v4.6.4. Do I need a newer version?

PR #587, opened 2020-05-12
Assume role with web identity provider

alindeman:
AWS allows users and systems to assume roles using OpenID Connect tokens via the [AssumeRoleWithWebIdentity API call]. This API allows role assumption via "Amazon Cognito, Login with Amazon, Facebook, Google, or any OpenID Connect-compatible identity provider."

This change proposes that aws-vault support AssumeRoleWithWebIdentity, either by fetching a token from a static file (via the standard [web_identity_token_file] config option) or by executing a process (via the web_identity_token_process config option, similar to [credential_process] but adapted for web identity). The latter would be a custom option for aws-vault, and would allow for interactive logins through OpenID Connect identity providers that can redirect back to localhost or [urn:ietf:wg:oauth:2.0:oob].

A static example:

```
[profile foo]
role_arn = arn:aws:iam::123457890:role/foo
web_identity_token_file = /path/to/file
```

A dynamic example, using a tool like [oidccli]:

```
[profile foo]
role_arn = arn:aws:iam::123457890:role/foo
web_identity_token_process = oidccli -issuer=https://example.com -client-id=aws -client-secret=localonly raw
```

[AssumeRoleWithWebIdentity API call]: https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html
[web_identity_token_file]: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-oidc
[credential_process]: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html
[urn:ietf:wg:oauth:2.0:oob]: https://cloud.google.com/iap/docs/authentication-howto#signing_in_to_the_application
[oidccli]: https://github.com/pardot/oidc/blob/master/cmd/oidccli


vault/assumerolewithwebidentityprovider.go
```diff
@@ -0,0 +1,161 @@
...
+		return string(b), nil
+	}
+
+	// Exec WebIdentityTokenProcess to retrieve OpenID Connect token
+	var cmdArgs []string
+	if runtime.GOOS == "windows" {
+		cmdArgs = []string{"cmd.exe", "/C", p.WebIdentityTokenProcess}
+	} else {
+		cmdArgs = []string{"/bin/sh", "-c", p.WebIdentityTokenProcess}
+	}
```
mtibben, line 104: Would using https://github.com/mattn/go-shellwords (instead of invoking the shell) be a better approach? Or are there specific shell features you're expecting to be used?

vault/vault.go
```diff
@@ -192,6 +225,8 @@ func (t *tempCredsCreator) provider(config *Config) (credentials.Provider, error
 		}
 	} else if config.HasSSOStartURL() {
 		return NewSSORoleCredentialsProvider(t.keyring.Keyring, config)
+	} else if config.HasRole() && (config.HasWebIdentityTokenFile() || config.HasWebIdentityTokenProcess()) {
+		return NewAssumeRoleWithWebIdentityProvider(credentials.NewCredentials(sourceCredProvider), t.keyring.Keyring, config)
```
mtibben, line 229: Referencing `sourceCredProvider` inside this if-else block doesn't seem right, as the sourceCredProvider is defined in the if-else. Does this condition need to move after the final else?

USAGE.md
```diff
@@ -391,6 +392,27 @@ sso_account_id=123456789012
 sso_role_name=Administrator
 ```
 
+## Assuming roles with web identities
+
+AWS supports assuming roles using [web identity federation and OpenID Connect](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-oidc), including login using Amazon, Google, Facebook or any other OpenID Connect server. The configuration options are as follows:
+* `web_identity_token_file` A file that contains an OpenID Connect identity token. The token is loaded and passed as the `WebIdentityToken` argument of the `AssumeRoleWithWebIdentity` operation.
+* `web_identity_token_process` A command that executes to generate an OpenID Connect identity token. The token written to the command's standard out is passed as the `WebIdentityToken` argument of the `AssumeRoleWithWebIdentity` operation.
```
mtibben, line 399: I can't find any reference to `web_identity_token_process` in the AWS CLI docs. Is this a new config parameter?

mtibben: Ah yes I see from the PR description that it is.

Although we have endeavoured to retain compatibility with the aws cli config, this seems pretty reasonable.

We should document this further perhaps in the config section of this doc, pointing out it is not standard

vault/assumerolewithwebidentityprovider.go
```diff
@@ -0,0 +1,161 @@
...
+	select {
+	case err := <-readCh: // process exited
+		if err != nil && err != io.EOF {
+			return "", fmt.Errorf("read output from %q failed: %v", p.WebIdentityTokenProcess, err)
+		}
+	case <-timer.C: // timeout
+		return "", fmt.Errorf("command %q timed out after %s", p.WebIdentityTokenProcess, defaultTimeout)
+	}
+
+	return b.String(), nil
```
mtibben, line 160: There is a lot of code here with channels, defers, etc, and I'm wondering if this is more complex than it needs to be.

Wouldn't a `out, err := exec.Command(cmdArgs).Output()` suffice here?

vault/assumerolewithwebidentityprovider.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Wait for read to finish (or timeout)
+	select {
+	case err := <-readCh: // process exited
+		if err != nil && err != io.EOF {
+			return "", fmt.Errorf("read output from %q failed: %v", p.WebIdentityTokenProcess, err)
+		}
+	case <-timer.C: // timeout
+		return "", fmt.Errorf("command %q timed out after %s", p.WebIdentityTokenProcess, defaultTimeout)
+	}
```
mtibben, line 158: There is a lot of code here with channels, defers, etc, and I'm wondering if this is more complex than it needs to be.

Would a `out, err := exec.Command(cmdArgs).Output()` suffice here?

vault/vault.go
alindeman: Good catch. Since we're using a web identity to assume a role, we don't actually need source credentials at all. This is a copy & paste error.

I've updated it in 6f25cc6 to use the same method for constructing a session as `NewSSORoleCredentialsProvider` which also does not require a `sourceCredProvider`.

vault/assumerolewithwebidentityprovider.go
alindeman: This function is admittedly modeled after the [aws-sdk-go's processcreds provider](https://github.com/aws/aws-sdk-go/blob/0961961790e6dc3be58a7fa17bfc294143320536/aws/credentials/processcreds/provider.go#L298), since I modeled `web_identity_token_process` after `credential_process`. That function invokes a shell, so I was inspired to do it here as well.

If you think it's better to avoid executing a shell, I'd be happy to implement it that way though.

USAGE.md
alindeman: > Although we have endeavoured to retain compatibility with the aws cli config, this seems pretty reasonable.

Awesome. It's designed to be really similar to the officially-supported `credential_process` option, but for retrieving a web identity token instead of an AWS access key/secret key/session key directly.

> We should document this further perhaps in the config section of this doc, pointing out it is not standard

:+1: What do you think of 4d712c1?

vault/assumerolewithwebidentityprovider.go
alindeman: It would suffice for the simple case, yes. This bit of code is again inspired by [aws-sdk-go's implementation of processcreds](https://github.com/aws/aws-sdk-go/blob/0961961790e6dc3be58a7fa17bfc294143320536/aws/credentials/processcreds/provider.go#L332).

This code is more defensive than `exec.Command(cmdArgs).Output()` would be in some cases. Specifically:
* If the process writes more than `defaultMaxBufSize` bytes to stdout, we will stop reading and close our end of the pipe. The process will die with EPIPE/SIGPIPE. The amount of memory we use as a buffer is fixed. On the other hand, `Output` is implemented with a `bytes.Buffer` which will expand 'infinitely' (until it runs out of memory).
* It will timeout and kill the process after `defaultTimeout`, but this could be also achieved using `exec.CommandContext()`.

Since `aws-vault` is configured by the user for their own use on their own machine, we might not need to be as defensive to processes that run away and write lots to stdout. I think in the worse case, `aws-vault` would panic with `ErrTooLarge` and die? If you think that's the case, I'm happy to simplify this to use `exec.Command().Output` or `exec.ContextCommand().Output()`. Let me know :)

mtibben: Yes I don't think we need to be as defensive as the SDK, just go with the simple code I reckon

alindeman: Sounds good. I pushed that up as 78eb6646be3b96119337dce80955585c817c7e12.

mtibben:
Thank you @alindeman !

alindeman:
Thanks for pulling this in @mtibben. I'm fairly motived to see v6 ship soon, so if there are any other issues I can help with around that, definitely let me know :smile:

PR #590, opened 2020-05-17
Get aws vault params from ~/.aws/config file (v2)

olfway:
As suggested in #582, read keyring params from the separate section

```
[aws-vault]
backend = pass
pass_prefix = aws-vault

[default]
output=table
region = eu-central-1
mfa_prompt = osascript
```

mfa prompt method is already passed in Config.MfaPromptMethod,
so I think it's better to read it from the profile config

### DIFF
--- a/cli/global.go
+++ b/cli/global.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


olfway:
@mtibben did you had a chance to look at it?

PR #591, opened 2020-05-17
Implement clear-cache command

### LINKED ISSUE #412: "clear" alias for "--sessions-only"

olfway:
Implements `aws-vault clear-cache [<profile>]` to remove a cached sessions
Fixes #412

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -276,9 +276,11 @@ $ aws-vault login work
 
 ### Removing stored sessions
 
-If you want to remove sessions managed by `aws-vault` before they expire, you can do this with the `--sessions-only` flag.
+If you want to remove sessions managed by `aws-vault` before they expire, you can do this with `aws-vault clear-cache` command.
+
+You can also specify a profile to remove sessions for this profile only.
 ```shell
-aws-vault remove <profile> --sessions-only
+aws-vault clear-cache [profile]
 ```
 
 ### Using --no-session
--- a/cli/clear-cache.go
+++ b/cli/clear-cache.go
@@ -1,0 +1,51 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/99designs/aws-vault/v6/vault"
+	"github.com/99designs/keyring"
+	"github.com/alecthomas/kingpin"
+)
+
+type ClearCacheCommandInput struct {
+	ProfileName string
+}
+
+func ConfigureClearCacheCommand(app *kingpin.Application, a *AwsVault) {
+	input := ClearCacheCommandInput{}
+
+	cmd := app.Command("clear-cache", "Clear cached sessions")
+
+	cmd.Arg("profile", "Name of the profile").
+		HintAction(a.MustGetProfileNames).
+		StringVar(&input.ProfileName)
+
+	cmd.Action(func(c *kingpin.ParseContext) (err error) {
+		keyring, err := a.Keyring()
+		if err != nil {
+			return err
+		}
+		err = ClearCacheCommand(input, keyring)
+		app.FatalIfError(err, "clear-cache")
+		return nil
+	})
+}
+
+func ClearCacheCommand(input ClearCacheCommandInput, keyring keyring.Keyring) error {
+	ck := &vault.CredentialKeyring{Keyring: keyring}
+	sk := &vault.SessionKeyring{Keyring: ck.Keyring}
+	var n int
+	var err error
+	if input.ProfileName == "" {
+		n, err = sk.RemoveAll()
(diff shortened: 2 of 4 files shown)


mtibben:
Cool, I think that looks good @olfway, and will work with the OIDC token caching (when it's implemented)

Anyone with any further opinions on the ergonomics of this sub-command?
/cc @lox @pda @jstewmon @FernandoMiguel @lwc

rossmckelvie:
How does this differ from `aws-vault remove -s <profile>` to clear the cached session? The UX problem described in #412 I would resolve by asking to confirm that they want to delete the stored credentials.

USAGE.md
```diff
@@ -275,9 +275,11 @@ $ aws-vault login work
 
 ### Removing stored sessions
 
-If you want to remove sessions managed by `aws-vault` before they expire, you can do this with the `--sessions-only` flag.
+If you want to remove sessions managed by `aws-vault` before they expire, you can do this with `aws-vault clear-cache` command.
+
+You can also specify a profile to remove sessions for this profile only.
 ```shell
-aws-vault remove <profile> --sessions-only
+aws-vault clear-cache [<profile>]
```
pda, line 282: How about square to make it clear that it's an optional argument? See https://en.wikipedia.org/wiki/Command-line_interface#Command_description_syntax for the closest thing I can find to a standard.

```suggestion
aws-vault clear-cache [profile]
```

mtibben:
> How does this differ from aws-vault remove -s <profile> to clear the cached session? The UX problem described in #412 I would resolve by asking to confirm that they want to delete the stored credentials.

I agree it would be nice if we could incorporate this better in the `aws-vault remove` command

> Or perhaps in v6, change the flags around so you have to pass an additional flag to delete the credentials, but by default will purge the cache.

I think we should be careful about flipping semantics for destructive commands :)

> Should it be "clear-cache" or "clear-sessions" though?

We're about to implement OIDC caching, and I suspect if SAML gets implemented there may be other tokens that also need to be cached, which may not be strictly "sessions". Also there is sometimes confusion between GetSessionToken "sessions" and AssumeRole "sessions". It would be good to clarify terminology

mtibben:
I've added support for caching OIDC tokens in #628.

We now have 3 different types of credentials being stored in the Keyring. Master credentials, STS (session) credentials, and OIDC tokens.

So we need to consider how to deal with this in the `list`, `remove` and now this new `clear-cache` command

We don't currently account for this well in the list command, the oicd token gets listed under credentials. 

```
$ aws-vault list
Profile                  Credentials              Sessions
=======                  ===========              ========
default                  -                        -
99designs                99designs                sts.GetSessionToken:7h14m23s
contests                 -                        -
-                        oidc:http://example.org/oidc   -
```

The OIDC token in the information hierarchy here is different to a STS session credentials because it could potentially be used by multiple profiles. Any suggestions on how to display this better?

So I'm thinking
- `aws-vault rm`: nothing, displays usage
- `aws-vault rm -s`: remove all session credentials and oidc tokens for all profiles
- `aws-vault rm PROFILE`: interactively prompt the user to confirm each master and session credential to remove
- `aws-vault rm PROFILE -s`: remove all session credentials and oidc tokens associated with the given profile
- and maybe `aws-vault rm PROFILE -y`: remove all master/session/oidc credentials for the given profile

Any thoughts @rossmckelvie @olfway @pda

rossmckelvie:
I like where you're going with putting the OIDC url in the credentials column. Following your example it would look something like this for dual-SSO organizations:
```
Profile              Credentials                               Sessions
=======              ===========                               ========
test-sso-role1       oidc:https://test.awsapps.com/start       sso.GetRoleCredentialsInput:7h59m52s
test-sso-role2       oidc:https://test.awsapps.com/start       sso.GetRoleCredentialsInput:6h21m44s
test-sso-alt1        oidc:https://new2.awsapps.com/start       sso.GetRoleCredentialsInput:1h11m02s
```

mtibben:
I went backwards and forwards on how this should work a few times. I arrived back at the original proposal for a `clear` command in #412, and also took @pda's suggestion to deprecate the `--sessions-only` flag.

New PR is at #644

PR #593, opened 2020-05-22
Add credential helper provider

Fall1ngStar:
Implements credential process as a new credential provider.

Based on the [AWS CLI documentation](https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#sourcing-credentials-from-external-processes)

aws-vault can be used as a source for credential_process but could not use one to retrieve credentials.

### DIFF
--- a/vault/credentialprocessprovider.go
+++ b/vault/credentialprocessprovider.go
@@ -2,9 +2,11 @@ package vault
 
 import (
 	"encoding/json"
+	"fmt"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"os"
 	"os/exec"
-	"strings"
+	"runtime"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


vault/credentialprocessprovider.go
```diff
@@ -0,0 +1,53 @@
...
+	p.SetExpiration(cred.Expiration, p.ExpiryWindow)
+	return credentials.Value{
+		AccessKeyID:     cred.AccessKeyId,
+		SecretAccessKey: cred.SecretAccessKey,
+		SessionToken:    cred.SessionToken,
+	}, nil
+}
+
+func (p *CredentialProcessProvider) callCredentialProcess() (CredentialProcessResponse, error) {
+	params := strings.Split(p.CredentialProcess, " ")
```
mtibben, line 41: You'll want to either use [shellwords](https://github.com/Wing924/shellwords) here or pass it to an actual shell as [AssumeRoleWithWebIdentityProvider does](https://github.com/99designs/aws-vault/pull/587/files#diff-cf36306ae347a936c254fd164fd396ceR93-R97)

vault/vault.go
```diff
@@ -227,6 +237,8 @@ func (t *tempCredsCreator) provider(config *Config) (credentials.Provider, error
 		return NewSSORoleCredentialsProvider(t.keyring.Keyring, config)
 	} else if config.HasRole() && (config.HasWebIdentityTokenFile() || config.HasWebIdentityTokenProcess()) {
 		return NewAssumeRoleWithWebIdentityProvider(t.keyring.Keyring, config)
+	} else if config.HasCredentialProcess() {
+		return NewCredentialProcessProvider(config)
```
mtibben, line 241: This is the source credentials right? So should probably be `sourceCredProvider = NewCredentialProcessProvider(config)`

vault/credentialprocessprovider.go
Fall1ngStar: Updated based on AssumeRoleWithWebIdentityProvider as suggested

vault/vault.go
Fall1ngStar: What are "master credentials" ? Are they used directly, without creating a temporary session like the `--no-session` parameter ?

mtibben:
Will this cause issues when using aws-vault as a credential_process (i.e `credential_process = aws-vault ... `)? Seems like this would make it very easy to get into an infinite loop.

Can you explain the use-case here @Fall1ngStar ?

Fall1ngStar:
I have a small tool that generates me temporary credentials for specific AWS account. However, it is very slow for reasons out of my control. aws-vault helps me to keep the generated credentials in memory, and to use them to assume other roles easily. 

It may be possible to create an infinite loop, but I don't think one would be able to create a loop involuntarily: In order to create a loop, one would have to set the credential process to an aws-vault command for the same profile, which doesn't make much sense as this would not yield any credentials in the first place. If you see another case where a loop would happend, please tell me, and I will be happy to provide a solution.

mtibben:
Thanks @Fall1ngStar, I don't think we'll support this for the time being as there is not a clear use-case for the general user and instead could lead to problems for users already using aws-vault in credential process

mtibben:
Fixed in #1087

Issue #331, opened 2019-01-22
Credentials were refreshed, but the refreshed credentials are still expired


teeberg:
I added a profile like this to my `~/.aws/config`:
```
[profile teeberg]
credential_process = aws-vault exec teeberg --json
```
and it's asking me for access to they mac keychain, but I keep getting this error:
> Credentials were refreshed, but the refreshed credentials are still expired.

with all subsequent requests failing (naturally).

Looking at the command's output, it does mention `"Expiration":"2019-01-22T11:48:09Z"`, which is in the past. I don't have this problem when running my own helper script as mentioned in the issue description in https://github.com/99designs/aws-vault/issues/299. Is this an issue with the `--json` flag?

cc @frezbo

frezbo:
@teeberg the ask of the keychain is a different issue. Which command are you trying?
This is my output:

Output of aws-vaullt exec with `-j` flag

```
aws-vault exec -j product
{"Version":1,"AccessKeyId":"<redacted>","SecretAccessKey":"<redacted>","SessionToken":"<redacted>","Expiration":"2019-01-23T01:28:51Z"}
```

s3 ls output

```
~ (aws:none)(kc:none)$ aws s3 ls --profile product
2017-09-12 17:02:40 <redacted>-awsmacietrail-dataevent
```

AWS config file

```
[profile product]
region=us-east-1
credential_process=aws-vault exec -j product
```

teeberg:
Sorry, didn't mean to say that the keychain would be an issue, just meant to mention it as a data point to help debugging that it seems in fact to be looking at previous session data.

With a config like
```
[profile teeberg]
credential_process = aws-vault exec teeberg --json
```
I only get this:
```
> AWS_PROFILE=teeberg aws sts get-caller-identity

Credentials were refreshed, but the refreshed credentials are still expired.
```
but this profile works:
```
[profile teeberg]
credential_process = aws-vault exec teeberg -- aws_vault_credential_process
```
(with `aws_vault_credential_process` from #299)

dpoggi:
This appears to be due to https://github.com/99designs/aws-vault/blob/v4.5.0/cli/exec.go#L143

This formats a local time from `time.Now()` as a UTC timestamp, causing a seemingly "early" expiration. The fix would be to operate on the result of `time.Now().UTC()` instead.

teeberg:
[...]

PR #600, opened 2020-06-04
Use RFC3339 consistently for expiration times

ajkerrigan:
Hi aws-vault maintainers, thanks for this lovely tool and the excellent recent additions in the v6 beta! I've got a minor update here to use RFC3339 for all `time.Format()` calls related to credential expiration.

`exec` was using RFC3339 in some places but `"2006-01-02T15:04:05Z"` in others. The latter format renders:

```
2020-06-04T13:14:44-04:00
```

as:

```
2020-06-04T13:09:44Z
```

and leads to `Credentials were refreshed, but the refreshed credentials are still expired` errors as noted in #331. In the v6 beta I've seen those errors with both the credential helper and `--ecs-server` functionality.

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -226,7 +226,7 @@ func execCredentialHelper(input ExecCommandInput, config *vault.Config, creds *c
 		credentialData.SessionToken = val.SessionToken
 	}
 	if credsExpiresAt, err := creds.ExpiresAt(); err == nil {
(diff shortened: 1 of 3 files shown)


mtibben:
Huh, nice catch. I'm not sure how that happened, I think the assumption was that the time would be outputted as UTC with that particular format

Thank you @ajkerrigan

PR #610, opened 2020-07-04
Add possibility to check for localised error messages, add localised …

### LINKED ISSUE #607: Windows 10 server not working
Hi there,

I'm using version 5.4.4 on Windows 10 installed with chocolatey. In general I can use aws-vault without problems, but when I try to start the server I get an error. The following log is when running the command in an elevated Powershell 7 prompt (tried other prompts, Powershell 1 or cmd, no difference). Any ideas?

```
C:\Windows\System32> aws-vault exec dev -s --debug
2020/07/02 15:22:17 aws-vault v5.4.4
2020/07/02 15:22:17 [keyring] Considering backends: [wincred pass file]
2020/07/02 15:22:17 Loading config file C:\Users\user\.aws\config
2020/07/02 15:22:17 Parsing config file C:\Users\user\.aws\config
2020/07/02 15:22:17 Using region "eu-west-1" from AWS_REGION
2020/07/02 15:22:17 profile dev: using stored credentials
2020/07/02 15:22:17 profile dev: using GetSessionToken (with MFA)
2020/07/02 15:22:17 Looking for sessions for dev
2020/07/02 15:22:17 Looking up all keys in keyring
2020/07/02 15:22:17 Session "session,ZGV2,YXJuOmy,1593698620" expires in 41m22.4149802s
2020/07/02 15:22:17 Re-using cached credentials ****************XXXX generated from GetSessionToken, expires in 41m22.4149802s
2020/07/02 15:22:17 Starting `aws-vault server` in the background
aws-vault: error: exec: Failed to start credential server: The credential proxy server isn't running. Run aws-vault server as Administrator in the background and then try this command again
```

```Powershell
aws-vault.exe server --debug
2020/07/02 21:25:54 aws-vault v5.4.4
2020/07/02 21:25:54 [keyring] Considering backends: [wincred pass file]
2020/07/02 21:25:54 Loading config file C:\Users\user\.aws\config
2020/07/02 21:25:54 Parsing config file C:\Users\user\.aws\config
```

EDIT:

I debugged the code and the problem seems to be 

```Go
if err == nil || strings.Contains(string(out), "The object already exists") {
		return []byte{}, nil
}
```

This will work on systems where the language is set to be English, but fail on all others with a confusing error message. Maybe #596 is the same problem.

I'm happy to commit a PR with my localised netsh error message. There doesn't seem to be a better way than parsing the error messages... Additionally, I would then test if the localised language can be matched and otherwise provide a more concise error message. Let me know if that sounds good to you :)

Agraphie:
This fixes #607.
Added the possibility to internationalise/localise error messages from the `netsh` command via cmd. 

I also looked into the possibility of using Powershell instead of netsh, but didn't see any benefit as I did not find proper error codes either and the errors messages were also localised.

### DIFF
--- a/server/ec2alias_windows.go
+++ b/server/ec2alias_windows.go
@@ -8,16 +8,39 @@ import (
 	"strings"
 )
 
+var alreadyRegisteredLocalised = []string{
+	"The object already exists",
+	"Das Objekt ist bereits vorhanden",
+}
+
+var runAsAdministratorLocalised = []string{
+	"Run as administrator",
+	// truncate before 'Umlaut' to avoid encoding problems coming from Windows cmd
+	"Als Administrator ausf",
+}
+
 func installEc2EndpointNetworkAlias() ([]byte, error) {
 	out, err := exec.Command("netsh", "interface", "ipv4", "add", "address", "Loopback Pseudo-Interface 1", "169.254.169.254", "255.255.0.0").CombinedOutput()
 
-	if err == nil || strings.Contains(string(out), "The object already exists") {
+	outMsg := string(out)
+
+	if err == nil || msgFound(alreadyRegisteredLocalised, outMsg) {
 		return []byte{}, nil
 	}
 
-	if strings.Contains(string(out), "Run as administrator") {
+	if msgFound(runAsAdministratorLocalised, outMsg) {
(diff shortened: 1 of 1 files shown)


Agraphie:
Thanks for merging this!

PR #608, opened 2020-07-02
Refresh cached credentials if already expired when calling Retrieve

jimbrowne:
I'm fairly certain this is a bug, but could be specific to one of my use cases.  In this case I have a credential `42lines-vault` that is restricted by MFA.  I have a profile I assume `qa-assume` from the first credential.  The `qa-assume` credential is used by the metadata server.

When the ec2 metadata server refreshes the expired `qa-assume` credentials, there is no check to see if the cached `42lines-vault` credentials are expired;  `cachedsessionprovider.go` only checks for existence.

Logs which include call tracing I added.  (I have set the expiration time on the assume credential to 15 minutes to reduce the testing window, but this same behavior happens if the expiration is one hour.)  Note the expiration time of the cached credential:
`Re-using cached credentials ****************F2EM from sts.GetSessionToken, expires in -8m44.905996s`

Current upstream:
```
bash-5.0$ curl -s http://169.254.169.254:80/latest/meta-data/iam/security-credentials/local-credentials
2020/06/29 23:19:23 server.go:2012 ServeHTTP() -> ec2.go:124 func1()
2020/06/29 23:19:23 Credentials.IsExpired() = true
2020/06/29 23:19:23 credentials.go:263 singleRetrieve() -> assumeroleprovider.go:28 Retrieve()
2020/06/29 23:19:23 assumeroleprovider.go:29 Retrieve() -> assumeroleprovider.go:55 assumeRole()
2020/06/29 23:19:23 assumeroleprovider.go:59 assumeRole() -> assumeroleprovider.go:43 roleSessionName()
2020/06/29 23:19:23 credentials.go:263 singleRetrieve() -> cachedsessionprovider.go:25 Retrieve()
2020/06/29 23:19:23 [keyring] Querying keychain for service="aws-vault", keychain="aws-vault.keychain"
2020/06/29 23:19:23 [keyring] Found 2 results
2020/06/29 23:19:23 [keyring] Querying keychain for service="aws-vault", account="sts.GetSessionToken,NDJsaW5lcy12YXVsdA,YXJuOmF3czppYW06OjY4OTM1NTQ5NjQ4MzptZmEvamJyb3duZQ,1593497446", keychain="aws-vault.keychain"
2020/06/29 23:19:30 [keyring] Found item "aws-vault session for 42lines-vault (expires 2020-06-30T06:10:46Z)"
2020/06/29 23:19:30 Re-using cached credentials ****************F2EM from sts.GetSessionToken, expires in -8m44.905996s
[...]

PR #612, opened 2020-07-10
Add --min-duration option

jimbrowne:
AWS Java SDK (at least v1 https://github.com/aws/aws-sdk-java/issues/1893; likely v2 as well) can not handle credentials with an expiration less than 15 minutes in the future.  Add a `--min-duration` option to set an expiration floor on the credentials used by the ec2/ecs server to force expiration.  Default to 15 minutes plus a slop factor of 5 seconds.

An alternate approach would be to promote the value used for `ExpiryWindow` to an option *and* track that value so it can be used when calculating the expiration of cached server credentials.  That is not currently happening (a `--duration` of 20m immediately shows credentials with an expiration of 15m.)

Log of this change in action:
```
2020/07/10 10:29:56 Re-using cached credentials ****************5HET from sts.GetSessionToken, expires in 7h54m25.591359s
2020/07/10 10:29:57 Generated credentials ****************3WM3 using AssumeRole, expires in 20m59.941839s
2020/07/10 10:29:57 Setting subprocess env: AWS_DEFAULT_REGION=us-east-1, AWS_REGION=us-east-1
2020/07/10 10:29:57 Starting child process: /usr/local/bin/bash 
2020/07/10 10:29:57 Starting EC2 Instance Metadata server on 127.0.0.1:9099
bash-5.0$ curl -s http://169.254.169.254:80/latest/meta-data/iam/security-credentials/local-credentials
2020/07/10 10:29:59 Credentials.IsExpired() = false
2020/07/10 10:29:59 Serving credentials via http ****************3WM3, expiration of 2020-07-10T17:45:57Z (15m57.565722s)
2020/07/10 10:29:59 http: 127.0.0.1:61877: 200 GET /latest/meta-data/iam/security-credentials/local-credentials (508.552µs)
{CENSORED}
bash-5.0$ curl -s http://169.254.169.254:80/latest/meta-data/iam/security-credentials/local-credentials
2020/07/10 10:30:57 Credentials.IsExpired() = false
2020/07/10 10:30:57 Forcing expiration of credentials due to minValidDuration (14m59.132351s)
2020/07/10 10:30:58 Generated credentials ****************KLUN using AssumeRole, expires in 20m59.727339s
2020/07/10 10:30:58 Serving credentials via http ****************KLUN, expiration of 2020-07-10T17:46:58Z (15m59.727264s)
2020/07/10 10:30:58 http: 127.0.0.1:61877: 200 GET /latest/meta-data/iam/security-credentials/local-credentials (405.20597ms)
{CENSORED}
```

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -28,6 +28,7 @@ type ExecCommandInput struct {
 	CredentialHelper bool
 	Config           vault.Config
 	SessionDuration  time.Duration
+	MinValidDuration time.Duration
 	NoSession        bool
 }
 
@@ -68,6 +69,10 @@ func ConfigureExecCommand(app *kingpin.Application, a *AwsVault) {
 	cmd.Flag("server", "Run a server in the background for credentials").
 		Short('s').
 		BoolVar(&input.StartEc2Server)
+
+	cmd.Flag("min-duration", "Trigger a refresh of background server credentials if time to expiration is below this threshold.  Defaults to 15m5s").
+		Default("15m5s").
+		DurationVar(&input.MinValidDuration)
 
 	cmd.Flag("ec2-server", "Run a EC2 metadata server in the background for credentials").
 		Hidden().
@@ -185,7 +190,7 @@ func updateEnvForAwsVault(env environ, profileName string, region string) enviro
 }
 
 func execEc2Server(input ExecCommandInput, config *vault.Config, creds *credentials.Credentials) error {
-	if err := server.StartEc2CredentialsServer(creds, config.Region); err != nil {
+	if err := server.StartEc2CredentialsServer(creds, config.Region, input.MinValidDuration); err != nil {
 		return fmt.Errorf("Failed to start credential server: %w", err)
 	}
 
@@ -196,7 +201,7 @@ func execEc2Server(input ExecCommandInput, config *vault.Config, creds *credenti
 }
 
 func execEcsServer(input ExecCommandInput, config *vault.Config, creds *credentials.Credentials) error {
-	uri, token, err := server.StartEcsCredentialServer(creds)
+	uri, token, err := server.StartEcsCredentialServer(creds, input.MinValidDuration)
 	if err != nil {
 		return fmt.Errorf("Failed to start credential server: %w", err)
 	}
--- a/server/ec2.go
+++ b/server/ec2.go
@@ -46,7 +46,7 @@ func isServerRunning(bind string) bool {
 }
 
 // StartEc2CredentialsServer starts a EC2 Instance Metadata server and endpoint proxy
-func StartEc2CredentialsServer(creds *credentials.Credentials, region string) error {
+func StartEc2CredentialsServer(creds *credentials.Credentials, region string, minValidDuration time.Duration) error {
 	if !isServerRunning(ec2MetadataEndpointAddr) {
 		if err := StartEc2EndpointProxyServerProcess(); err != nil {
 			return err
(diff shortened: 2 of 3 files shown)


mtibben:
Hmmm doesn't `--duration` do this already?

mtibben:
Right, so currently the ExpiryWindow is hard-coded at 5 minutes - see `defaultExpirationWindow` - and it's referenced in 8 places in the code. It sounds like what you're really wanting to do is to make that configurable. Perhaps doing this via an env var would make sense, else this arg would need to be added to `exec`, `login` and `server` ?

jimbrowne:
>  It sounds like what you're really wanting to do is to make [defaultExpirationWindow] configurable

`defaultExpirationWindow` could be made configurable via option or environment variable, but the true expiration time has to be tracked separately as well.  I can re-factor this change to use that approach and be configured via option or environment variable.  My bias is for an option, but I'm happy to defer to your preference.  Given that both AWS Java SDK versions have this 15 minute restriction I'd also suggest that the default value be changed from five to fifteen minutes.

mtibben:
> can not handle credentials with an expiration less than 15 minutes in the future
> Given that both AWS Java SDK versions have this 15 minute restriction
Hey @jimbrowne, reading aws/aws-sdk-java#1893 I might be missing something but it doesn't seem like there is any restriction. The issue there is that the java SDKs minimum expiry window was 15 minutes, and other SDKs are 5 minutes. In this case, aws-vault also uses the common 5 minute default. From my understanding it doesn't mean the credentials stop working, it's just that they're refreshed if they're within that window

jimbrowne:
> The issue there is that the java SDKs minimum expiry window was 15 minutes, and other SDKs are 5 minutes.

(I dug through the Java SDKv2 and v2 uses a 15 _second_ threshold, so not affected by this issue.)

AWS Java SDK v1 will re-fetch the credentials on every request if the credentials returned expire in the next 15 minutes.  It's been over a month, so I don't have the logs at hand and can't recall if this led to overwhelming the `aws-vault` proxy or if it was causing problems on the Java side due to too many threads blocked on the mutex around fetching credentials.  The Java application is at times making multiple calls per second to AWS.

Looking at the `GetSessionToken` call (referenced in the IAM FAQ) perhaps the correct answer is failing to enforce a 15 minute minimum is a bug rather than needing this option.  I don't believe it is possible for the EC2 Metadata instance service to return credentials with less than 15 minutes of lifetime.

Since some may desire the more liberal five minute expiration an option to tune the minimum duration enforced seems appropriate.  As this looks to only affect AWS Java SDK v1 I can maintain a fork until we are fully on SDK v2; though this can affect anyone else using `aws-vault` and v1.

mtibben:
> I don't believe it is possible for the EC2 Metadata instance service to return credentials with less than 15 minutes of lifetime
Are you referring to the real ec2 metadata service or the aws-vault one?

In any case, it seems that if you could configure `defaultExpirationWindow` this would solve your issue

mtibben:
Can you try the latest v6 beta @jimbrowne? I think this should now be resolved with the `AWS_MIN_TTL` env var and the fix in #608

mtibben:
I'm still in 2 minds about a `--min-duration` flag. 

The point of `--duration` was that it was a simple flag to indicate "i need x time, give me a session". We're only using the duration for _new_ sessions, but this is generally fine because it's just the MFA session token that is cached, not the role session which is created each time with the given duration.

So I _think_ `AWS_MIN_TTL` should help with your use-case now, and that is the reason I'm not sure `--min-duration` is required

jimbrowne:
> Are you referring to the real ec2 metadata service or the aws-vault one?

PR #613, opened 2020-07-13
Add a prompt that is native to Windows

BradBarnich:
This solved a pain point for me when using aws-vault on windows as a credential helper, since the console prompt doesn't work in that scenario. 

Forgive me, this is the first Go code that I've ever written. Hope it can be worked into something mergeable

Thanks

### DIFF
--- a/prompt/credui_windows.go
+++ b/prompt/credui_windows.go
@@ -21,7 +21,7 @@ type creduiInfoA struct {
 	hbmBanner      uintptr
 }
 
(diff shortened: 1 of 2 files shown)


mtibben:
Thanks @BradBarnich! I made a few tweaks to remove the use of globals and renamed to `credui` to be more specific - i.e. how "native" is it? Does this work with every type of windows - windows xp, windows IoT?

BradBarnich:
Thank you!

It is an older API, so I presume it is fairly compatible. But I don’t know.

PR #622, opened 2020-07-30
Add --force-new-session option to exec command

jimbrowne:
Addresses https://github.com/99designs/aws-vault/issues/122#issuecomment-661893667.  To prevent periods of expired credentials from the metadata server and to enable prompting a user for an MFA refresh at convenient times this additional option is needed.

Built by refactoring Ber Zoidberg's pull request https://github.com/99designs/aws-vault/pull/474 to work with current master.  I did not include the modification to the rotate command.

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -29,6 +29,7 @@ type ExecCommandInput struct {
 	Config           vault.Config
 	SessionDuration  time.Duration
 	NoSession        bool
+	ForceNewSession  bool
 }
 
 // AwsCredentialHelperData is metadata for AWS CLI credential process
@@ -56,6 +57,10 @@ func ConfigureExecCommand(app *kingpin.Application, a *AwsVault) {
 
 	cmd.Flag("region", "The AWS region").
 		StringVar(&input.Config.Region)
+
+	cmd.Flag("force-new-session", "Force a new session to be created and update cache").
+		Short('f').
+		BoolVar(&input.ForceNewSession)
 
 	cmd.Flag("mfa-token", "The MFA token to use").
 		Short('t').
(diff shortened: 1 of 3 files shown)


mtibben:
Hey @jimbrowne, can you explain what you are trying to achieve with a use-case and logs? If the metadata server serves up expired credentials, that is a bug, so I'm unsure what the option is achieving.

jimbrowne:
I'll have to set up an environment to re-look at the expired credentials case, but the human use case for this is easy to explain.  From my referenced comment in #122 :

> I now have a babysitter process that watches time remaining in an MFA session. I do not want any time period of invalid credentials, so the babysitter process sleeps and attempts to refresh the MFA credentials when there is less than a few minutes left

Compare current:

Time 00:00: User MFA/Yubikeys in terminal to create a session, say one hour
Time 01:00: Credentials expire, MFA/Yubikey is prompted in a terminal window the user isn't looking at, meanwhile their locally running,  AWS using applications are failing on all AWS requests, especially confusing less technical users.

With the ability to force a new session by ignoring whatever older credential may be in cache:

Time 00:00: User MFA/Yubikeys in terminal to create a session, say one hour
Time 00:40: Babysitting process (external to `aws-vault`) notices there is 20 minutes of validity remaining, prompts user via `osascript` to touch their Yubikey as a part of creating a new session.  Can monitor if user did touch the Yubikey and keep trying with notifications.
[ Locally running, AWS using applications never have a time period in which their AWS requests fail due to expired credentials.]

I never want `aws-vault` to serve expired credentials.  To do that I have to prompt the user in-advance of the expiration time, and I need the ability to force a replacement of any cached credential.  The former is handled by the babysitting process, the latter requires the functionality of this pull request.

For more context, the locally running applications aren't necessarily driven by human actions.  They are reading messages off of SQS queues and performing other actions, so it's not a case of the human performing an action and the action failing due to credentials and there's a tight feedback loop (e.g. from running the `aws` CLI).  The applications just start failing in odd ways that are irritating to developers and confusing to less technical users.

mtibben:
Can you try the latest v6 beta @jimbrowne? I think this should now be resolved with the addition of `clear`, the `AWS_MIN_TTL` env var and the fix in #608

jimbrowne:
I will run a test case to see if using `clear` followed immediately by an `exec` will work.  Of the top of my head this will create a race condition, but probably one with which we can live.  I believe `--force-new-session` is cleaner, but my hope is to be able to use a non-forked version of `aws-vault` so I'll see if this works well enough.

PR #628, opened 2020-08-17
Add OIDC token caching

### LINKED ISSUE #571: Restore OIDC token caching

mtibben:
Adds caching for OIDC tokens. As these tokens aren't attached to a profile, these will show up as "credentials" under `aws-vault list` for now.

@itsdalmo @rickardl could you please test this?

Fixes #571

### DIFF
--- a/vault/oidctokenkeyring.go
+++ b/vault/oidctokenkeyring.go
@@ -1,0 +1,71 @@
+package vault
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/99designs/keyring"
+	"github.com/aws/aws-sdk-go/service/ssooidc"
+)
+
+type OIDCTokenKeyring struct {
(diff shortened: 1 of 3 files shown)


itsdalmo:
I have built and tested the latest commit (5f20470) on `master` now, and everything seems to work for the initial setup etc (have to wait some hours to see what happens to expired sessions and OIDC tokens). The one thing I'm a bit worried about is the lack of caching for the OIDC Client credentials, which according to the AWS API documentation should be persisted (and afaik the expiration tended to be months into the future):

> Registers a client with AWS SSO. This allows clients to initiate device authorization. The output should be persisted for reuse through many authentication requests.



Source: https://docs.aws.amazon.com/singlesignon/latest/OIDCAPIReference/API_RegisterClient.html

What are your thoughts?

PR #636, opened 2020-08-19
show remaining oidc time

flyinprogrammer:
```
./aws-vault list
Profile                  Credentials                           Sessions                         
=======                  ===========                           ========
-                        oidc:https://magicurl/start/          oidc:5h53m17s
```

Maybe this helps?

### DIFF
--- a/cli/list.go
+++ b/cli/list.go
@@ -136,7 +136,17 @@ func ListCommand(input ListCommandInput, awsConfigFile *vault.ConfigFile, keyrin
 	for _, credentialName := range credentialsNames {
 		_, ok := awsConfigFile.ProfileSection(credentialName)
 		if !ok {
-			fmt.Fprintf(w, "-\t%s\t-\t\n", credentialName)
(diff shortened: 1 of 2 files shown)


mtibben:
Hey thanks for the effort here @flyinprogrammer.

Calling `Get` on the keyring will cause a prompt for most backends, something we want to avoid on the `list` command. The underlying issue is that the [keyring](https://github.com/99designs/keyring) library doesn't have proper support for retrieving metadata... we've hacked around this for STS sessions by storing the expiry in the key, but I'm not keen to duplicate those hacks in OIDCTokenKeyring. Instead I think we should fix keyring to support unstructured metadata so that we can use it in instances like this

Also, we have 3 types of objects stored in the keyring now - master "credentials", "sessions"  and "oidc tokens". I'm still not sure the best visual way to display that is, there has been some discussion here https://github.com/99designs/aws-vault/pull/591#issuecomment-674711707. Here you've split the oidc under both the credentials and sessions header which might be confusing as there is only one object

PR #640, opened 2020-08-25
Add pass-otp as an otp prompt

Tacklebox:
This pr adds support for using pass-otp as an otp prompt.
I am not sure if this would be better implemented as generic otp
prompt that can call any command to generate an otp code. That would
also be able to take the place of the ykman prompt. Either way,
given that pass is a supported backend, I think it makes sense to
support pass as an otp prompt as well.

### DIFF
--- a/prompt/passotp.go
+++ b/prompt/passotp.go
@@ -1,0 +1,34 @@
+package prompt
+
+import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


mtibben:
Nice! Thanks @Tacklebox

PR #656, opened 2020-09-11
Store SSO expiration times as UTC

ajkerrigan:
Maintain consistency with session credential behavior by storing
token and credential expiration times as UTC.


mtibben:
I've taken a slightly different approach in #657

PR #660, opened 2020-09-18
Handle additional username formats

cirego:
Not all AWS User ARNs have the username as the second field. Some may
have it as the 3rd or 4th field. However, it seems very likely that most accounts
have them as the last field in the ARN.

This patch changes the behavior to grab the last field in the username and I hope
that it is behavior preserving for the original use case as well.

This patch is required for the organization that I work in and it would be great if we
can get this merged upstream!

Thanks for all the hard work that went into aws-vault. It's a great tool and I use
it everyday!


mtibben:
Thanks @cirego

PR #662, opened 2020-09-19
Support sts_regional_endpoints configuration

### LINKED ISSUE #543: Cannot use assume role with "opt-in" regions (e.g., ap-east-1)
- [ X] I am using the latest release of AWS Vault
- [ ] I have provided my `.aws/config` (redacted if necessary)

I have not been able to get aws-vault to work with ap-east-1 which uses regional sts for assuming roles.

mtibben:
Support the `sts_regional_endpoints` in the aws config file or via the `AWS_STS_REGIONAL_ENDPOINTS` env var.

Reference: https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#aws-sts

This may help with #543

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -118,6 +118,7 @@ To configure the default flag values of `aws-vault` and its subcommands:
 To override the AWS config file (used in the `exec`, `login` and `rotate` subcommands):
 * `AWS_REGION`: The AWS region
 * `AWS_DEFAULT_REGION`: The AWS region, applied only if `AWS_REGION` isn't set
+* `AWS_STS_REGIONAL_ENDPOINTS`: STS endpoint resolution logic, must be "regional" or "legacy"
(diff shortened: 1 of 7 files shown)


ronkorving:
@mtibben How do I enable this as a user from my config file? This should make AWS Vault run a bit faster for us on the other side of the planet (of us-east-1) I guess. But it's not entirely clear for me from the doc update in this PR on how to do this in my config file.

I've tried

```ini
sts_regional_endpoints = regional
```

But when I execute `aws vault exec myprofile -- env |grep AWS` I do not see `AWS_STS_REGIONAL_ENDPOINTS` getting set. Does it only apply to the inner workings of AWS Vault and not set an env var for whatever I end up executing next? If it only applies to internals, is there a way for me to confirm I've configured it right?

mtibben:
Yep try with --debug

ronkorving:
Works! Awesome! Thank you :)

Issue #670, opened 2020-10-09
aws-vault fails with `specified keyring backend not available` when using `kwallet` without specifying the `--backend` flag


Mr-istov:
Hello, until a few days ago while using `aws-vault` **v5.4.0** I didn't had any problems running `aws-vault` with `kwallet` as a backend without specifying the `--backend` flag, but after upgrading to **v6.1.0** it started throwing me the following error:
```
aws-vault: error: Specified keyring backend not available, try --help
```

This is the output with the `--debug` flag:
```
2020/10/09 17:29:23 [keyring] Considering backends: [secret-service]
2020/10/09 17:29:23 [keyring] Failed backend secret-service: The name org.freedesktop.secrets was not provided by any .service files
aws-vault: error: Specified keyring backend not available, try --help
```

If I run `aws-vault` with specifying the `--backend=kwallet` flag it seems that it works, but I wonder why I didn't had this problem before, was this changed with the upgrade somehow?

- [ ] I am using the latest release of AWS Vault
- [ ] I have provided my `.aws/config` (redacted if necessary)
- [x] I have provided the debug output using `aws-vault --debug` (redacted if necessary)

gdamjan:
the regression has been introduced with this change: https://github.com/99designs/aws-vault/commit/7572b93daed03c44a3b984913bced6ebdb2dca80#diff-83de63909ead282862b13050fe347d68

removing the `Default(backendsAvailable[0]).` line fixes the issue - by not specifying a backend, and allowing keyring to find it automatically.

Mr-istov:
@mtibben Was there a reason as to why was that line added, and if it is not required, could it be removed?

mtibben:
AWS Vault should be able to tell you what backend it's going to use before it does - the "fallback" behaviour is good when  when determining what default to display, but is not desirable when actually executing. If keyring is not returning available backends correctly, the problem is likely there.

gdamjan:
In both cases aws-vault reports available backends and **their order**…
(v6.1.0):
```
  --backend=secret-service   Secret backend to use [secret-service kwallet pass file] ($AWS_VAULT_BACKEND)
```

patched with #672:
```
  --backend=BACKEND          Secret backend to use [secret-service kwallet pass file] ($AWS_VAULT_BACKEND)
```

the problem is that the `Default(backendsAvailable[0]).` line in aws-vault tells keyring to fix the backend to just that one, which actually limits what keyring is capable of.

[...]

PR #672, opened 2020-10-10
Remove default value from backend flag

Mr-istov:
Addresses https://github.com/99designs/aws-vault/issues/670#issuecomment-706337487. To allow the keyring to find the `backend` automatically instead of making the `secret-service` default and throwing an error when not using `secret-service` as a `backend`.
Revert changes made by https://github.com/99designs/aws-vault/commit/7572b93daed03c44a3b984913bced6ebdb2dca80#diff-83de63909ead282862b13050fe347d68 in which the change was introduced.

### DIFF
--- a/cli/global.go
+++ b/cli/global.go
@@ -85,7 +85,6 @@ func ConfigureGlobals(app *kingpin.Application) *AwsVault {
 		BoolVar(&a.Debug)
 
 	app.Flag("backend", fmt.Sprintf("Secret backend to use %v", backendsAvailable)).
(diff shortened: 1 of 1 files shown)


mtibben:
See [this comment](https://github.com/99designs/aws-vault/issues/670#issuecomment-706866983), the problem is more likely to be in keyring

Mr-istov:
Closing this PR, see https://github.com/99designs/aws-vault/issues/670#issuecomment-707846310 for more information as to why.

PR #675, opened 2020-10-15
Add pinentry prompt driver

jtdoepke:
This PR adds an OTP prompt driver for GnuPG's [pinentry](https://gnupg.org/related_software/pinentry/index.html) programs. I figure if you're using the pass backend, pinentry is probably available.


Liath:
Can this get resurrected?

PR #685, opened 2020-11-04
STS AssumeRole Session Tags implementation

### LINKED ISSUE #684: Feature: add support for STS Session Tags when assuming roles
Our company is using aws-vault in our devops workflows and we would like to implement an ABAC authorisation strategy against some resources.

It would be great if aws-vault supported IAM Session Tags [when `AssumeRole()` is used](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html#id_session-tags_adding-assume-role).

Ideally, the session and transitive tags should be set up separately for each role in the chain  using `~/.aws/config` profile parameters as well as environment variables.

stanvit:
This PR is trying to resolve #684 and implements the way to set Session Tags and Transitive Session Tags when `AssumeRole()` provider is used.

- Two new profile configuration parameters are introduced: `session_tags` and `transitive_session_tags`
- Two new environment variables are introduced: `AWS_SESSION_TAGS` and `AWS_TRANSITIVE_TAGS`

It is possible to configure each profile in the chain individually, as well as the target profile with environment variables.

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -328,22 +328,14 @@ func (cl *ConfigLoader) populateFromConfigFile(config *Config, profileName strin
 	if config.STSRegionalEndpoints == "" {
 		config.STSRegionalEndpoints = psection.STSRegionalEndpoints
 	}
-	if sessionTags := psection.SessionTags; sessionTags != "" && len(config.SessionTags) == 0 {
-		if config.SessionTags == nil {
-			config.SessionTags = make(map[string]string)
-		}
-		for _, tag := range strings.Split(sessionTags, ",") {
-			kvPair := strings.SplitN(tag, "=", 2)
-			if len(kvPair) != 2 {
-				return errors.New("Failed to parse session_tags profile setting, must be <key1>=<value1>,[<key2>=<value2>[,...]]")
-			}
-			config.SessionTags[strings.TrimSpace(kvPair[0])] = strings.TrimSpace(kvPair[1])
+	if sessionTags := psection.SessionTags; sessionTags != "" && config.SessionTags == nil {
+		err := config.SetSessionTags(sessionTags)
+		if err != nil {
+			return fmt.Errorf("Failed to parse session_tags profile setting: %s", err)
 		}
 	}
-	if transitiveSessionTags := psection.TransitiveSessionTags; transitiveSessionTags != "" && len(config.TransitiveSessionTags) == 0 {
-		for _, tag := range strings.Split(transitiveSessionTags, ",") {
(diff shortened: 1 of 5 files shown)


vault/config.go
```diff
@@ -326,6 +328,23 @@ func (cl *ConfigLoader) populateFromConfigFile(config *Config, profileName strin
...
+			if len(kvPair) != 2 {
+				return errors.New("Failed to parse session_tags profile setting, must be <key1>=<value1>,[<key2>=<value2>[,...]]")
+			}
+			config.SessionTags[strings.TrimSpace(kvPair[0])] = strings.TrimSpace(kvPair[1])
+		}
+	}
+	if transitiveSessionTags := psection.TransitiveSessionTags; transitiveSessionTags != "" && len(config.TransitiveSessionTags) == 0 {
+		for _, tag := range strings.Split(transitiveSessionTags, ",") {
+			config.TransitiveSessionTags = append(config.TransitiveSessionTags, strings.TrimSpace(tag))
+		}
```
mtibben, line 346: could you split this logic out into it's own function?

vault/config.go
```diff
@@ -417,6 +436,27 @@ func (cl *ConfigLoader) populateFromEnv(profile *Config) {
...
+			}
+			log.Printf("Using session_tags %v from AWS_SESSION_TAGS", profile.SessionTags)
+		}
+
+		if transitiveSessionTags := os.Getenv("AWS_TRANSITIVE_TAGS"); transitiveSessionTags != "" && len(profile.TransitiveSessionTags) == 0 {
+			for _, tag := range strings.Split(transitiveSessionTags, ",") {
+				profile.TransitiveSessionTags = append(profile.TransitiveSessionTags, strings.TrimSpace(tag))
+			}
+			log.Printf("Using transitive_session_tags %v from AWS_TRANSITIVE_TAGS", profile.TransitiveSessionTags)
+		}
```
mtibben, line 459: probably shares some logic with this?

vault/config.go
stanvit: please check the latest change, I created a couple of setters for `vault.Config` that parse string into the corresponding map and slice.

I'm not 100% confident that it's the best approach - it's "economical", but kind of moves parsing away from `vault.ProfileSection` to `vault.Config`. Please tell me your thoughts on that.

christophetd:
@stanvit Thanks a lot for the PR, this is definitely needed, looking forward to seeing this merged :tada:

dmccaffery:
@mtibben : anything required to get this PR merged -- I am happy to chip in with any updates.

PR #687, opened 2020-11-04
USAGE.md update to include GPG_TTY insight for pass backend

prelegalwonder:
Update Usage Documentation to add hint to issue sometimes found with aws-vault interacting with pass without certain TTY options set for GPG.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -7,6 +7,7 @@
   * [Environment variables](#environment-variables)
 * [Backends](#backends)
   * [Keychain](#keychain)
+  * [Pass](#pass)
 * [Managing credentials](#managing-credentials)
   * [Using multiple profiles](#using-multiple-profiles)
   * [Listing profiles and credentials](#listing-profiles-and-credentials)
@@ -149,7 +150,7 @@ If you're looking to configure the amount of time between having to enter your K
 
 ![keychain-image](https://imgur.com/ARkr5Ba.png)
 
-<br>
+### Pass
 
 **Note:** If using `pass` backend and you're experiencing the below error, you may need to `export GPG_TTY=$(tty)`
 ```
--- a/USAGE.md
+++ b/USAGE.md
@@ -149,6 +149,14 @@ If you're looking to configure the amount of time between having to enter your K
 
 ![keychain-image](https://imgur.com/ARkr5Ba.png)
 
+<br>
+
+**Note:** If using `pass` backend and you're experiencing the below error, you may need to `export GPG_TTY=$(tty)`
+```
+Error when retrieving credentials from custom-process: gpg: decryption failed: No secret key
+gpg: decryption failed: No secret key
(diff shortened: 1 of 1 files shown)


mtibben:
You've put this under the "Keychain" heading @prelegalwonder ?

prelegalwonder:
@mtibben I can move it, I wasn't quite sure where to place it. It seemed somewhat "keychain" related in general since it has to do with the GPG keychain. What would you prefer?

mtibben:
Add a heading for pass. Keychain is the macOS backend

prelegalwonder:
> Add a heading for pass. Keychain is the macOS backend

Ah sorry, missed that. Updated to give Pass its own section and put the tip under there. Is that what you had in mind?

USAGE.md
```diff
@@ -149,6 +150,14 @@ If you're looking to configure the amount of time between having to enter your K
...
 ![keychain-image](https://imgur.com/ARkr5Ba.png)
 
+### Pass
+
+**Note:** If using `pass` backend and you're experiencing the below error, you may need to `export GPG_TTY=$(tty)`
+```
+Error when retrieving credentials from custom-process: gpg: decryption failed: No secret key
+gpg: decryption failed: No secret key
+aws-vault: error: exec: Failed to get credentials for <your-profile>: exit status 2
+```
```
mtibben, line 160: > If using `pass` backend and you're experiencing the below error, you may need to

Can you explain the error, why it occurs, and why the suggested fix works?

prelegalwonder: Honestly, I'd be guessing at best and it may be inaccurate. It seems that gpg-agent requires TTY info since the password prompt is an interactive curses like interface that fills up the whole screen after a refresh instead of just a `password:` prompt at a newline. 

When interacting with pass directly I didn't have the issue and I'm assuming its because my terminal preserves that tty info. But my suspicion is that the system / shell call from aws-vault to pass doesn't preserve or set that info, so the act of setting GPG_TTY in the env var allows it to fall back to that to get the tty device (and this info about the session and its cols/rows). 

Do you want me to put my guesstimate of why its needed in the markdown?

mtibben: I'd prefer not to put guesses into the USAGE, but if you explain what is going on then at least it can be corrected if not accurate

mtibben: > the system / shell call from aws-vault to pass doesn't preserve or set that info, so the act of setting GPG_TTY in the env var allows it to fall back to that to get the tty device

Wait.. this sounds like a bug in aws-vault then. Can you confirm this and fix in aws-vault directly?

prelegalwonder: Sorry, I don't have any experience with Golang and don't have time to dive into it at this time. 

I understand if you don't want to accept the change based on the context. I just thought I'd contribute to try and save others the same lost time I had running into this issue.

PR #693, opened 2020-11-19
Update arch linux install instructions

LuRsT:
The package has been included in the community repository, so there's no need to install it from AUR


mtibben:
Nice, thanks @LuRsT

PR #704, opened 2020-12-18
Added support to pass aliases as commands in non windows environments.

tpreecesh:
Hi guys, we have  a need to use aliases as commands and I have changed the cli/exec.go execSyscall function to allow this.

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -337,16 +337,29 @@ func supportsExecSyscall() bool {
 func execSyscall(command string, args []string, env []string) error {
 	log.Printf("Exec command %s %s", command, strings.Join(args, " "))
 
-	argv0, err := osexec.LookPath(command)
-	if err != nil {
+	newArgs := make([]string, 0)
+	cmd, err := osexec.LookPath(command)
+	if len(cmd) < 1 || err != nil {
+		if runtime.GOOS != "windows" {
+			sh := os.Getenv("SHELL")
+			if strings.HasSuffix(sh, "/bash") || strings.HasSuffix(sh, "/fish") ||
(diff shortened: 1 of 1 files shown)


mtibben:
Do you mean like a bash alias? Can you can pass the full shell command to `exec` to achieve this?

Always executing via the shell feels like it may create a ton of edge-cases. Perhaps an aws-vault wrapper script could achieve this also.

georgekaz:
Yes it means shell aliases, for example, if you have `alias tfplan=terraform plan`, you can successfully execute `aws-vault exec my-profile terraform plan` but cannot execute `aws-vault exec my-profile tfplan`, at least that's what I experienced with zsh.
`aws-vault: error: exec: Couldn't find the executable 'tfplan': exec: "tfplan": executable file not found in $PATH`

mtibben:
Right but I imagine you can `aws-vault exec my-profile -- bash -c 'tfplan'`?

AWS Vault can't inject environment variables directly into your existing shell, so you can also start an interactive shell and then execute... i.e

```shell
$ aws-vault exec my-profile # starts a subshell with AWS env vars
$ tfplan
```

georgekaz:
Yes, `aws-vault exec my-profile -- bash -c 'tfplan'` does work so that's an alternative, and I didn't realise you could start a subshell using `aws-vault exec my-profile` so we can use that info, thanks. 

As it turns out, because this process will be containerised, it's more suitable to for us to use https://github.com/benkehoe/aws-sso-credential-process as described here: https://www.cloudgardener.dev/how-to-use-aws-developer-tools-with-aws-sso/ because it doesn't rely on a local keystore.

Thanks

PR #729, opened 2021-02-22
aws helper scripts fixed to work on more platforms

chrismazanec:
- removed bashisms from AWS helper scripts so they work under Linux / dash too
- forced (`-f`) `ykman` to overwrite MFA code on YubiKey if it exists already. this is useful when MFA config in AWS console has been deleted and setting up a new MFA using this script. `ykman` would otherwise fail silently and left AWS MFA in an inconsistent state (created but unactivated).
- fixed cleanup function to actually run on EXIT (`-z "$OUTFILE"` ?)
- removed `-u` (`nounset` ) to not fail with `unbound variable` if the script fails before setting `OUTFILE`

### DIFF
--- a/contrib/scripts/aws-iam-create-yubikey-mfa.sh
+++ b/contrib/scripts/aws-iam-create-yubikey-mfa.sh
@@ -1,28 +1,23 @@
 #!/bin/sh
 # Adds a Yubikey TOTP device to IAM
 
-set -eu
+set -e
 
 if [ -n "${AWS_SESSION_TOKEN:-}" ]; then
   echo "aws-vault must be run without a STS session, please run it with the --no-session flag" >&2
   exit 1
 fi
 
-cleanup()
-{
-  if [ -z "$OUTFILE" ]; then
-    rm "$OUTFILE"
-  fi
-}
+cleanup() { rm -f "$OUTFILE"; }
 trap cleanup EXIT
 
 waittime() {
     # wait until next code can be generated
     # if SECONDS are :00 or :30, generate right away
     SECONDS=$(date +%S)
-    if (( ${SECONDS#0} >= 1 && ${SECONDS#0} <= 29 )); then
+    if [ ${SECONDS#0} -gt 0 ] && [ ${SECONDS#0} -lt 30 ]; then
         WAIT_SECONDS=$(( 30 - ${SECONDS#0} ))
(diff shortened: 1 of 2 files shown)


nikolay (approved):

contrib/scripts/aws-iam-create-yubikey-mfa.sh
```diff
@@ -1,28 +1,23 @@
 #!/bin/sh
 # Adds a Yubikey TOTP device to IAM
 
-set -eu
+set -e
```
mtibben, line 4: I'd prefer the safer `set -euo pipefail`

contrib/scripts/aws-iam-create-yubikey-mfa.sh
```diff
@@ -1,28 +1,23 @@
...
 fi
 
-cleanup()
-{
-  if [ -z "$OUTFILE" ]; then
-    rm "$OUTFILE"
-  fi
-}
+cleanup() { rm -f "$OUTFILE"; }
 trap cleanup EXIT
```
mtibben, line 12: you could do the trap _after_ setting `OUTFILE`

chrismazanec: I was thinking about that, but in that case, temp file would exist longer time leaving more time for a potential attacker to find it and later read it. That's the same reason `mktemp` and also `rm` is straight before and after the command generating TOTP secret. `rm` would otherwise done on `trap`. what do you think?

contrib/scripts/aws-iam-create-yubikey-mfa.sh
chrismazanec: `set -eo pipefail` makes sense, for the `-u` please read my other comment and tell me what you think

contrib/scripts/aws-iam-create-yubikey-mfa.sh
nikolay: You can then use `declare`.

chrismazanec: `declare` did not seem to help. I have put `-u` `(nounset)` back along with default for `OUTFILE`, which seems to work ok. `-o pipefail` is a bashism and doesn't work in `dash`, so I would rather not put that in.

chrismazanec:
can this be merged, please?

Issue #741, opened 2021-03-07
Replace deprecated ykman commands with newer unaliased versions


tomelliff:
Yubikey-manager v4.0.0 [deprecated the use of `ykman oath code`, replacing it with `ykman oath accounts code`](https://github.com/Yubico/yubikey-manager/commit/f1bbfc3a13ee0f6cef2656c365fbdc579da2d56f). It has [aliased the old command for now](https://github.com/Yubico/yubikey-manager/blob/068bb0d8717f86607dd66deb8a50cec9e30be298/ykman/cli/aliases.py#L102) but I suspect it will be removed in v5.

AWS-Vault could either detect the version of `ykman` and use the appropriate command or, and I think this is my preference for simplicity, just update to use the new command and release that change as a breaking change for AWS-Vault, requiring v4.0.0+ Yubikey-manager to be installed.

rantoniuk:
@mtibben any chance for getting this released 🥺 🙏🏻

nvdbleek:
The fix is in v6.4.0-beta1 is there an ETA for the official release of v6.4.0?

PR #743, opened 2021-03-07
Fix ykman deprecated command

tomelliff:
Fixes https://github.com/99designs/aws-vault/issues/741

Alternate, non breaking version of https://github.com/99designs/aws-vault/pull/742.

As mentioned in the issue I think the breaking change is probably fine because it doesn't need merging immediately but this version would allow it to be merged as is and support Yubikey manager before v4 and remove the deprecated warning in v4+.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
 	github.com/aws/aws-sdk-go v1.34.13
+	github.com/blang/semver v3.5.1+incompatible
 	github.com/danieljoos/wincred v1.1.0 // indirect
 	github.com/google/go-cmp v0.5.2
 	github.com/keybase/go-keychain v0.0.0-20200502122510-cda31fe0c86d // indirect
--- a/go.sum
+++ b/go.sum
@@ -14,6 +14,8 @@ github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d h1:UQZhZ2O0vMHr2c
 github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d/go.mod h1:rBZYJk541a8SKzHPHnH3zbiI+7dagKZ0cgpgrD7Fyho=
 github.com/aws/aws-sdk-go v1.34.13 h1:wwNWSUh4FGJxXVOVVNj2lWI8wTe5hK8sGWlK7ziEcgg=
 github.com/aws/aws-sdk-go v1.34.13/go.mod h1:5zCpMtNQVjRREroY7sYe8lOMRSxkhG6MZveU8YkpAk0=
+github.com/blang/semver v3.5.1+incompatible h1:cQNTCjp13qL8KC3Nbxr/y2Bqb63oX6wdnnjpJbkM4JQ=
+github.com/blang/semver v3.5.1+incompatible/go.mod h1:kRBLl5iJ+tD4TcOOxsy/0fnwebNt5EWlYSAyrTnjyyk=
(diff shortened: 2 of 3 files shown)


liamdawson:
Annoyingly, looks like we can't rely on a status-code based check for `ykman` v3.1.2:

```shell
$ ykman oath accounts code --single test
Usage: ykman oath [OPTIONS] COMMAND [ARGS]...
Try 'ykman oath -h' for help.

Error: No such command 'accounts'.

$ echo $?
2
```

```
$ ykman oath code --single arn:aws:iam
Touch your YubiKey...
Usage: ykman oath code [OPTIONS] [QUERY]
Try 'ykman oath code -h' for help.

Error: Touch credential timed out!

$ echo $?
2
```

So it seems like a solution that handles multiple `ykman` versions would either have to run the first command (and parse it for `No such command 'accounts'`, or do what you're doing here and parse the version.

ismailyenigul:
I think,  the code  should run `ykman -v` to check ykman version and run the commands based on  the detected version of ykman.

tomelliff:
@ismailyenigul I'm not sure I understand what you're commenting on? The code in this PR gets the output of `ykman --version` and checks if it is v4 or higher.

steeef:
Looks like there's conflicts again. Planning on resolving them here?

tomelliff:
> Looks like there's conflicts again. Planning on resolving them here?

Happy to fix conflicts as long as there's acceptance from a maintainer that they want this change, particularly over the alternate version (#742) . I don't tend to resolve conflicts on PRs before I've had any feedback from a maintainer though as it can just be wasted work and doubly so here where I have 2 competing proposals depending on how backwards compatible with `ykman` versions this project wants to be versus how much extra code they want to support that.

mtibben:
@tomelliff the reasons your PRs have not yet had feedback

1. You have left this PR in "Draft" status and not asked for feedback. Draft PR status indicates that the PR is NOT ready for feedback.
2. You have two PRs doing the same thing without explanation
3. The PR descriptions are very brief and don't explain what they are doing

tomelliff:
Apologies @mtibben. The proposal was to either have a breaking change and require `ykman` > v4 which is what https://github.com/99designs/aws-vault/pull/742 does or to have a non breaking change as per this pull request. I wasn't sure which you'd prefer so raised both and then you could decide which you'd rather merge depending on your policy around breaking changes.

I left this one in draft status and the other not as draft status to indicate my preference and that they should be considered conflicting.

Or another option would be to merge this one now and then allow for the other pull request to be merged (after being rebased on this one) at a later major version change of AWS-Vault.

What would you like me to do next? I can close one if you are sure which way you'd like to go in respect to the breaking changes and unmark this as draft if you'd like to go down this route.

mtibben:
I'd suggest default to the newer command line but allow users to opt into v3 behaviour with an env var

tomelliff:
> I'd suggest default to the newer command line but allow users to opt into v3 behaviour with an env var

Happy to make that change if that's what you'd prefer but that feels like you don't get the niceness of #742 where it just does away with supporting v3 ykman completely but is slightly more UX awkward than the change here with the same side effect of having code supporting v3 (although less of it and no extra dependency on `semver` which is maybe a good enough argument).

mtibben:
My rationale

This is a breaking change - if we follow the go versioning philosophy it could be argued a new prompt driver `ykman4` is appropriate.

But given we'd expect most to upgrade ykman via pip, I think changing the ykman default in line with that is fair enough, because most will be on the latest version. But it still is a breaking change and we still want a way to opt out. I'd prefer not to do version detection ourselves because that in itself may break in future - who knows? And as for the UX issue - well `YKMAN_OATH_CREDENTIAL_NAME` already needs to be set so setting `AWS_VAULT_YKMAN_VERSION=3` or similar doesn't seem unreasonable.

christophgysin:
@tomelliff Would you have time to update this PR accordingly?

PR #749, opened 2021-03-16
Allow specifying a browser to login with

dmattia:
Instead of logging in through the default browser, this adds an optional `--browser` flag that, if specified, will update the browser to login with.

As an example of one use case, I use this to login to multiple different IAM Roles at one time.


mtibben:
Hey @dmattia the `--stdout` flag is for this use case. Can you run something like `mybrowser "$(aws-vault login foo --stdout)"` to accomplish this?

dmattia:
That works

Issue #755, opened 2021-03-19
Feature: Make usage of Firefox containers


T00mm:
Currently I am using a custom script that lacks 99% of the functionality that aws-vault provides but what it does is, it uses the profile name in combination with [Firefox containers](https://addons.mozilla.org/en-GB/firefox/addon/multi-account-containers/?utm_content=addons-manager-reviews-link&utm_medium=firefox-browser&utm_source=firefox-browser) and [Open url in container](https://addons.mozilla.org/en-GB/firefox/addon/open-url-in-container/?utm_content=addons-manager-reviews-link&utm_medium=firefox-browser&utm_source=firefox-browser) to seperate each login from eachother.  Making it easy to access multiple consoles in the same browser and with colored tabs.

Kessir:
The `-s` flag on `aws-vault login` will "Print login URL to stdout instead of opening in default browser"

You can then open this URL however you want. 
In this example, I open the AWS console in a specific Microsoft Edge profile:

```open -na "Microsoft Edge" --args --profile-directory="Profile 2" $(aws-vault login -s staging)```

mtibben:
Use `-s` for custom behaviour like this

PR #756, opened 2021-03-19
Added support for firefox multi-containers

T00mm:
#755 I have implemented the Firefox Container option that works with

https://addons.mozilla.org/en-GB/firefox/addon/multi-account-containers/
https://addons.mozilla.org/en-GB/firefox/addon/open-url-in-container

I completely understand if you got the feeling that this should not merge as this is too dependent on multiple addons and one specific browser.

I have created a small preview of what it does

https://user-images.githubusercontent.com/3200970/111841619-88048380-88fe-11eb-8a1c-a536deb24f1c.mp4

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -18,12 +18,13 @@ import (
 )
 
 type LoginCommandInput struct {
-	ProfileName     string
-	UseStdout       bool
-	Path            string
-	Config          vault.Config
-	SessionDuration time.Duration
-	NoSession       bool
+	ProfileName         string
+	UseStdout           bool
+	Path                string
+	Config              vault.Config
+	SessionDuration     time.Duration
+	NoSession           bool
+	UseFirefoxContainer bool
 }
 
 func ConfigureLoginCommand(app *kingpin.Application, a *AwsVault) {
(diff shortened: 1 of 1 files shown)


mtibben:
Yeah this is getting too specific - use the `--stdout` flag and write a shell alias to achieve the same

T00mm:
With only stdout this can't work as stdout doesn't provide a profile name. Is there a plugin system somewhere on the roadmap as I agree with that this is too specific use-case.

mtibben:
something like this bash function would work on macos. Not too sure how to powershell on windows
```bash
awslogin() {
  firefox "$(printf ext+container:name=%s\&url=%s $1 $(aws-vault login --stdout $1))"
}
```

awilkins:
This extension wants the `&` in the extension-url for itself, so you need to urlencode the login URL for this to work.

I've used `jq` here because every AWS power user needs `jq`, right?

```sh
aws-login() {
    firefox "$(printf 'ext+container:name=%s&url=%s' $1 $(aws-vault login --stdout $1 | jq -sRr @uri))"
}
```

PR #769, opened 2021-04-21
Update AWS SDK to v2

mtibben:
The new AWS SDK has been GA since January, and seems to have a few improvements. The major benefit this will bring will be use of `context` in the EC2 / ECS server.


FernandoMiguel:
this is really good news

PR #773, opened 2021-04-26
Use LookupEnv rather than Getenv for AWS_VAULT_FILE_PASSPHRASE

heyvito:
Given one can simply press <kbd>Return</kbd> when prompted for a password in the terminal, I think we could also allow `AWS_VAULT_FILE_PASSPHRASE` to be empty.
This changes `fileKeyringPassphrasePrompt` in order to allow empty passphrases by using [`LookupEnv`](https://golang.org/pkg/os/#LookupEnv) instead of [`Getenv`](https://golang.org/pkg/os/#Getenv)


mtibben:
Thanks @heyvito

PR #778, opened 2021-05-14
Add option to follow XDG base directory spec

### LINKED ISSUE #707: Using $XDG_CONFIG_HOME For File Backend Keys
In case of using encrypted file backend (`AWS_VAULT_BACKEND=file`), aws-vault stores keys in `$HOME/.awsvault`.

To prevent filling user home directory with hidden files, can the application follow [XDG Base Directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and store the keys in `$XDG_CONFIG_HOME/awsvault` (`$HOME/.config/awsvault`)?

Or is there a way to adjust the default encrypted file backend location?

michaelbeaumont:
Closes #707 
I'm not sure how this should affect the `pass` directories.

### DIFF
--- a/cli/global.go
+++ b/cli/global.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path"
 
 	"github.com/99designs/aws-vault/v6/prompt"
 	"github.com/99designs/aws-vault/v6/vault"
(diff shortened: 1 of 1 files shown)


patrickli:
I personally think this directory should reside in `XDG_DATA_HOME` instead of `XDG_CONFIG_HOME`. Mainly because many people (including me) put `XDG_CONFIG_HOME` under version control.

* You don't want to include secrets in version control.
* Assumed roles will expire and key files related to assumed roles will be added/deleted/changed quite often.

Of course you can choose to ignore the folder but I think it is more appropriate to keep it out of `~/.config`.

Issue #623, opened 2020-08-02
Export STS creds to credentials file


RyanJarv:
Opening this mostly to see if there is any interest this and if I submitted a PR for it if it would be considered here.

This would be another form of running aws-vault, like aws-vault exec or, --server, but would instead either merge or update temporary credentials in ~/.aws/credentials (taking care to avoid updating non-sts creds). I suspect there would also be functionality to control when key's are updated, i.e. only if they expire in x hours. One difference from other modes might be it may make sense to control which profiles are exported via ~/.aws/config so that aws-vault can operate on all of them.

For me this solves my use case for originally wanting https://github.com/99designs/aws-vault/pull/400 merged, but should be somewhat simpler and doesn't require running in the background. The problem is that many tools don't support all of the options in ~/.aws/config that are needed with aws-vault, this end's up costing a lot of time in figuring out what's going on. Exporting to the credentials file ensure's compatibility with those tools and is useful for sharing multiple profiles to a single docker container in a way that most setup's already support. Having the temporary key's in the credentials file rather than the environment also gives you more access control's compared to the environment or tcp socket. One trade off might be the batch export mode if used would expose all key's it's enabled for at any given time rather then the specific one that was needed, so that would have to be up to the user whether or not to export specific or any enabled creds (which I'm imagining would be most useful in a cronjob).

Really this could be done in a shell script but have been finding myself needing to do this for more projects lately and imagine other people are running into the same thing and it would be nice if this could be done in aws-vault directly.

Let me know if you have any thoughts or suggestions here as well.

Thanks!

FernandoMiguel:
I entirely disagree with this proposal.
aws-vault was built exactly to avoid having (long lived) creds in clear text saved on disk.

any tools interacting with AWS should support AWS SDKs, which means env var creds and metadata (--server/--ecs) would work out of the box

ryan-gerstenkorn-sp:
@FernandoMiguel
> aws-vault was built exactly to avoid having (long lived) creds in clear text saved on disk.
I agree with this for long lived credentials, this would specifically be for temporary sts credentials.

> any tools interacting with AWS should support AWS SDKs, which means env var creds and metadata (--server/--ecs) would work out of the box
This get's more complicated when you need to support multiple profiles without assuming roles. There's also many cases where custom credential resolving doesn't behave exactly like the default lookup, or the config file doesn't load by default. I can start making a list of these cases if it's helpful here.

mtibben:
[...]

PR #785, opened 2021-06-15
Add export-creds.sh script

RyanJarv:
Adds a script to ./contrib that will export short or long term credentials for a given profile to ~/.aws/credentials.

This was discussed in #623 a while back.

Edit: So far I've only tested this on MacOS, will update if I get a chance to test this elsewhere.


contrib/scripts/export-creds.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env sh
```
irgeek, line 1: You're specifying a POSIX shell here, but several of the features you're using below aren't available in the POSIX shell. I ran [shellcheck](https://github.com/koalaman/shellcheck) over this script and I'll add the POSIX errors it reports to the lines that triggered them.

contrib/scripts/export-creds.sh
```diff
@@ -0,0 +1,106 @@
...
+
+set -euo pipefail
```
irgeek, line 3: SC3040: In POSIX sh, set option pipefail is undefined.

contrib/scripts/export-creds.sh
```diff
@@ -0,0 +1,106 @@
...
+
+umask 077
+
+err_report() {
+    echo "[ERROR] Command on line $1 exited with exit code $2"
+    exit $2
+}
+trap 'err_report $LINENO $?' ERR
```
irgeek, line 11: SC3047: In POSIX sh, trapping ERR is undefined.

contrib/scripts/export-creds.sh
```diff
@@ -0,0 +1,106 @@
...
+	passed in through Stdin (TODO: patch to awscli to fix this). This however does not apply after
+	the secret is exported to the credentials file.
+
+	This script also checks the permissions on ~/.aws/credentials to ensure it is not readable by
+	all users on the system and exits with an error if it is.
+'
+}
+
+NO_SESSION=''
+POSITIONAL=()
```
irgeek, line 57: SC3030: In POSIX sh, arrays are undefined.

contrib/scripts/export-creds.sh
```diff
@@ -0,0 +1,106 @@
...
+        NO_SESSION='--no-session'
+        shift
+        ;;
+        -h|--help)
+        help_long
+        exit 1
+        shift
+        ;;
+        *)    # unknown option
+        POSITIONAL+=("$1") # save it in an array for later
```
irgeek, line 70: SC3024: In POSIX sh, += is undefined.

irgeek (commented):
There are some non-POSIX features being used here, but you're using `sh` in the shebang. This is likely to break in mysterious ways for users on systems where `sh` isn't a symlink to `bash`. The easiest fix is probably just to use `bash` in the shebang.

contrib/scripts/export-creds.sh
RyanJarv: wonder if there's an alternative here, can't stand when shell scripts silently fail.

RyanJarv:
Awesome, thanks for checking this. Suppose I should have known better than to trust simply running bash's sh would enforce compatibility.

irgeek:
The AWS CLI you're invoking to scribble in the credentials file [supports](https://docs.aws.amazon.com/sdkref/latest/guide/file-location.html#file-location-config) the `AWS_SHARED_CREDENTIALS_FILE` environment variable to set a different location for the shared credentials file. Your read check isn't taking that into account, so you might be checking readability for the wrong file.

Also, the `ls -l` there throws an error on `stderr` if the file doesn't exist. If it doesn't exist, the AWS CLI takes care of setting permission correctly when creating the file, so you can probably just ignore the error an pipe `stderr` to `/dev/null` to clear that up.

irgeek:
> Suppose I should have known better than to trust simply running bash's sh would enforce compatibility.

In my experience, the `sh` mode of `bash` seems to have been created to enforce _incompatibly_.

contrib/scripts/export-creds.sh
```diff
@@ -0,0 +1,106 @@
...
+    exit 101
+fi
+
+
+aws-vault exec $NO_SESSION "${PROFILE}" -- sh -c "
+    set -eu
+    aws configure --profile "${PROFILE}-creds" set region \"\$AWS_REGION\"
+    aws configure --profile "${PROFILE}-creds" set aws_access_key_id \"\$AWS_ACCESS_KEY_ID\"
+    aws configure --profile "${PROFILE}-creds" set aws_secret_access_key \"\$AWS_SECRET_ACCESS_KEY\"
+    aws configure --profile "${PROFILE}-creds" set aws_session_token \"\${AWS_SESSION_TOKEN:-}\"
```
mtibben, line 103: This script may be unnecessarily verbose. Why not something simple like the following. 

```bash
#!/bin/sh
# Configure aws-cli using the AWS env vars present
set -eux
aws configure --profile "${1:-$AWS_VAULT}" set region "$AWS_REGION"
aws configure --profile "${1:-$AWS_VAULT}" set aws_access_key_id "$AWS_ACCESS_KEY_ID"
aws configure --profile "${1:-$AWS_VAULT}" set aws_secret_access_key "$AWS_SECRET_ACCESS_KEY"
aws configure --profile "${1:-$AWS_VAULT}" set aws_session_token "${AWS_SESSION_TOKEN:-}"
```

Call it with `aws-vault exec myprofile -- aws-configure-set-via-env-vars.sh myprofile-creds`

What am I missing?

RyanJarv:
> The AWS CLI you're invoking to scribble in the credentials file [supports](https://docs.aws.amazon.com/sdkref/latest/guide/file-location.html#file-location-config) the `AWS_SHARED_CREDENTIALS_FILE` environment variable to set a different location for the shared credentials file. Your read check isn't taking that into account, so you might be checking readability for the wrong file.

For some reason I wasn't able to get this working with the `aws configure` subcommand, seemed to just ignore the env variable. I probably need to take a look at it again though, kind of doubt it wouldn't be supported here.

irgeek:
> For some reason I wasn't able to get this working with the `aws configure` subcommand, seemed to just ignore the env variable. I probably need to take a look at it again though, kind of doubt it wouldn't be supported here.

It definitely works with AWS CLI -- I tested it myself. If you run it like this:

```bash
AWS_SHARED_CREDENTIALS_FILE=~/.aws/credentials2 export-creds.sh [profile]
```

The lines written by AWS CLI show up in `~/.aws/credentials2`. I was going to ask if maybe you were using an older version of AWS CLI, but it appears `AWS_SHARED_CREDENTIALS_FILE` has been supported since 2015, so you'd need to be running a _very_ old version.

contrib/scripts/export-creds.sh
irgeek: This does feel like an elegant solution. My only change would be to drop the `x` from `set -eux` so it doesn't echo credentials to the console. I imagine a script like this will used in a continuous-delivery / GitOps automated deployment scope where output is usually logged.

mtibben:
Fixed in 2518fb1

PR #798, opened 2021-08-23
Add support for session tags when creating a federated token

jondavidschober:
Using the session_tags functionality as a base, I added a new feature to allow adding session tags to the federation tokens that `aws-vault login` uses

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -9,8 +9,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/service/sts/types"
 	"github.com/mitchellh/go-homedir"
 	ini "gopkg.in/ini.v1"
 )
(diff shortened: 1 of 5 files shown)


vault/config.go
```diff
@@ -147,6 +147,7 @@ type ProfileSection struct {
 	STSRegionalEndpoints    string `ini:"sts_regional_endpoints,omitempty"`
 	SessionTags             string `ini:"session_tags,omitempty"`
 	TransitiveSessionTags   string `ini:"transitive_session_tags,omitempty"`
+	FederationSessionTags   string `ini:"federation_session_tags,omitempty"`
```
mtibben, line 150: Is there a case for. re-using `session_tags` instead of introducing `federation_session_tags`?

jondavid-schober-sp: We could re-use session_tags. I think the reason I went this route was because when initially investigated this feature, I made an assumption that the session tags were used for the federation token, but it wasnt, and it became unclear where they were used. I opted to be more explicit about where it's being used (context of federation). I would be ok with reusing SessionTags

PR #807, opened 2021-09-16
Add support for source identity on sessions

### LINKED ISSUE #771: Support for Source Identity on sessions?
AWS recently added a new feature to IAM sessions called a "source identity" - it sounds like a "stronger" version of using a session name to identify who assumed a role:

> When you perform actions in AWS, the information about your session can be logged to AWS CloudTrail for your account administrator to monitor. Administrators can configure roles to require identities to pass a custom string that identifies the person or application that is performing actions in AWS. This identity information is stored as the source identity in AWS CloudTrail. When the administrator reviews activity in CloudTrail, they can view the source identity information to determine who or what performed actions with assumed role sessions.
> 
> After a source identity is set, it is present in requests for any AWS action taken during the role session. The value that is set persists when a role is used to assume another role through the AWS CLI or AWS API, known as role chaining. The value that is set cannot be changed during the role session. Administrators can configure granular permissions based on the presence or value of the source identity to further control AWS actions that are taken with shared roles. You can decide whether the source identity attribute can be used, whether it is required, and what value can be used.
> 
> The way that you use source identity differs from role session name and session tags in an important way. The source identity value can't be changed after it is set, and it persists for any additional actions that are taken with the role session. Here's how you can use session tags and role session name:
> 
> *    Session tags – You can pass session tags when you assume a role or federate a user. Session tags are present when a role is assumed. You can define policies that use tag condition keys to grant permissions to your principals based on their tags. Then you can use CloudTrail to view the requests made to assume roles or federate users. To learn more about session tags, see Passing session tags in AWS STS.
> 
> *    Role session name – You can use the sts:RoleSessionName condition key in a role trust policy to require that your users provide a specific session name when they assume a role. Role session name can be used to differentiate role sessions when a role is used by different principals. To learn more about role session name, see sts:RoleSessionName.
> 
> We recommend that you use source identity when you want to control the identity that assumes a role. Source identity is also useful for mining CloudTrail logs to determine who used the role to perform actions. 

References:

* https://aws.amazon.com/about-aws/whats-new/2021/04/aws-identity-and-access-management-now-makes-it-easier-to-relate-a-users-iam-role-activity-to-their-corporate-identity/
* https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_control-access_monitor.html

vroad:
Added support for specifying source identity when assuming roles. Can be specified with `source_identity` in config file or `AWS_SOURCE_IDENTITY` environment variable.
Fixes #771

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -118,6 +118,20 @@ session_tags = key1=value1,key2=value2,key3=value3
 transitive_session_tags = key1,key2
 ```
 
+#### `source_identity`
+
+It is possible to set [source identity](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_control-access_monitor.html) when `AssumeRole` is used. Custom config variable `source_identity` allows you to set the value.
+
+```ini
+[profile root]
+region=eu-west-1
+
+[profile order-dev]
+source_profile = root
+role_arn=arn:aws:iam::123456789:role/developers
+source_identity=your_user_name
+```
+
 
 ### Environment variables
 
@@ -153,6 +167,8 @@ To override or set session tagging (used in `exec`):
 * `AWS_SESSION_TAGS`: Comma separated key-value list of tags passed with the `AssumeRole` call, overrides `session_tags` profile config variable
 * `AWS_TRANSITIVE_TAGS`: Comma separated list of transitive tags passed with the `AssumeRole` call, overrides `transitive_session_tags` profile config variable
 
+To override or set the source identity (used in `exec` and `login`):
+* `AWS_SOURCE_IDENTITY`: Specifies the source identity for assumed role sessions
 
 ## Backends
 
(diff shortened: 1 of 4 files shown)


vroad:
@mtibben I tried adding documentation, but not sure what to write. Currently It only says that it allows you set value. Could you make this better?

stekern:
Any updates to this? This would be a nice improvement.

gswallow:
This looks amazing but I'm not entirely sure how to use it.  Is it up to the person who retrieves an STS token through Vault to invent their own source identity, or can I write something to the aws/roles/deploy folder like this?

```
vault write aws/roles/deploy role_arns=arn:aws:iam::012345678901:role/VaultDeploymentServiceRole credential_type=assumed_role source_identity="hey you"
```

And then restrict who's allowed to assume my role by using sts:SourceIdentity conditions in my trust policy?

PR #815, opened 2021-10-11
Enable dependabot to get security updates 

neilnaveen:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically

Having knowledge about vulnerabilities of the dependencies helps the project owners decide on their dependencies security posture to make decisions.

If the project decides to get updates only on security updates and not on any version updates then setting these options would not open any PR 's open-pull-requests-limit: 0

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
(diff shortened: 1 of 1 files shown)


amirdafnyman1:
Please approve this PR. There are pending security updates and enabling dependabot will allow these to be created automatically as PRs.

PR #819, opened 2021-10-16
Add `--stdout` option to `aws-vault exec`

### LINKED ISSUE #678: Feature: AWS SSO "--no-browser" option
For various reasons such as not wanting to use the  system default browser, wanting to open the SSO link in a specific Firefox Containers container, etc, it would be useful to be able to optionally only print the `VerificationUriComplete` instead of only showing it if `open.Run()` fails.  My suggestion would be to add a new `--no-browser` or similar option that could be used to trigger this behaviour (with the default being the current widely useful "auto-open" method).

Once the link is opened once, it cannot be re-used, so currently there is no way to grab the link before it has been used the first time (unless the default URL handling browser fails to open or is missing)

https://github.com/99designs/aws-vault/blob/8a679cf21c5f74b7d9e967aa6ada2f9b543ba8cd/vault/ssorolecredentialsprovider.go#L129-L132

lsowen:
By default `aws-vault` opens the the AWS SSO link in the user's
default browser, but specifying `--no-browser` will only print
the AWS SSO link to the terminal, allowing the user to open
the link in a non-default way (eg firefox profile, on a different
machine, etc)

Fixes #678

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -32,6 +32,7 @@ type ExecCommandInput struct {
 	Config           vault.Config
 	SessionDuration  time.Duration
 	NoSession        bool
+	UseStdout        bool
 }
 
 func (input ExecCommandInput) validate() error {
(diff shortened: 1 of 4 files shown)


m4yankpomelo:
> hi @m4yankpomelo it looks like Github is still waiting on some action from a maintainer for "1 workflow"?

Accidentally approved, apologies :)

nazarewk:
@lsowen could you merge it?

lsowen:
hi @mtibben do you have any availability to review this PR?

mtibben:
@lsowen I think it would be best to make this consistent with `exec --stdout` and use the same flag?

lsowen:
@mtibben updated the PR from `--no-browser` to `--stdout`

PR #832, opened 2021-12-29
Make MFA prompt input secret

Volatus:
This PR simply makes the terminal prompt for the MFA code secret for better security.

Signed-off-by: Ismayil Mirzali <ismayilmirzeli@gmail.com>


analytically:
@mtibben

mtibben:
@Volatus MFA prompts are time-sensitive, so for usability it is generally considered a reasonable trade-off to keep the input visible

PR #833, opened 2022-01-02
docs: update 404ing Homebrew on Linux link

3stacks:
It looks like brew.sh no longer has a specific section for Linux formulae


3stacks:
Oops, this was already done here: https://github.com/99designs/aws-vault/pull/831

PR #858, opened 2022-02-06
Add manifest file for requiring elevated rights when running

Agraphie:
This PR should solve the issue with admin privileges on Windows. Currently, an elevated prompt needs to be opened first before executing aws-vault as aws-vault will not prompt for elevated privileges. Also there is some custom error format parsing via returned error messages to recognise if aws-vault was run from a non-elevated prompt. Potential solutions are:

- manifest file + syso binary file. This should be picked up immediately. Can be extended with icons and a signing key key. See https://docs.microsoft.com/de-de/windows/win32/sbscs/application-manifests, https://github.com/akavel/rsrc (for generating) or https://github.com/WireGuard/wireguard-windows/blob/master/manifest.xml (for areal world example)
- running elevation via powershell. This works, but is a rather hacky way to ask for elevation
- custom code e.g. like https://stackoverflow.com/a/59147866 to ask for permissions from within the app

I think the manifest + syso file would be the cleanest and most common solution. Once the syso file is present, `go build` should pick up the file. Another improvement would be to add a signing key and a logo/icon. Would also be better to not checkin the .syso file and build it via the build process.

I'm happy for any feedback.


mtibben:
@Agraphie On Windows 11 this causes the binary to not run at all

Agraphie:
Thanks for the feedback. I think I created this before I had Windows 11 (so on Windows 10), so let me test this and get back to you. Maybe they changed something for the manifest file.

Agraphie:
I couldn't find a convenient way to solve this and unfortunately I don't use aws-vault anymore. Thanks for checking though!

PR #864, opened 2022-02-11
[login] allow sourcing STS credentials from environment (closes #861)

### LINKED ISSUE #861: aws-vault login with existing STS temporary credentials
Hello!

I'd like to use aws-vault with existing STS temporary credentials. I don't want aws-vault to call `sts:AssumeRole` for me; I want to be able to pass it my access key ID, secret access key and session token and have it generating the AWS Console sign-in link.

Is this currently supported? My understanding is that it's currently [not supported](https://github.com/99designs/aws-vault/blob/master/cli/login.go#L99-L103) 

Thank you

christophetd:
Sample usage:

```
export AWS_ACCESS_KEY_ID=ASIA25...
export AWS_SECRET_ACCESS_KEY=ASyJq4...
export AWS_SESSION_TOKEN=IQoJb3JpZ2luX2Vj...
aws-vault login --from-env
``` 

Sample error:

``` 
export AWS_ACCESS_KEY_ID=ASIA25...
export AWS_SECRET_ACCESS_KEY=ASyJq4...
aws-vault login --from-env

aws-vault: error: login: failed to retrieve a session token. Cannot generate a login URL without it
exit status 1
```

Main use-case: Allow generating sign-in links when temporary credentials have been obtained from another way than aws-vault. There is third-party tooling doing it such as https://github.com/NetSPI/aws_consoler, but I feel like it's worth it to have it built in aws-vault.

Notes:
- While we _could_ support `login --from-env` with only IAM credentials, it would a bit more complex (retrieve from environment, then use the `GetFederationToken` provider) for something that IMHO doesn't add value. If you have IAM credentials, they are long-lived by definition and they should be managed through aws-vault anyway


mtibben:
Hmm I wonder if `--from-env` is even required. If the profile is missing, source from the environment?

That would also allow workflows such as

```shell
$ aws-vault exec foo
$ aws s3 ls
$ aws-vault login
```

Where aws-vault could use the environment to login from within an exec subshell.

Any downsides to that approach?

christophetd:
Good point, and it's a workflow I've needed multiple times. The only downside I see is that if you run `aws-vault login` without creds in your environment and forget the profile name, which error message to display? We're unsure if the user expected to have creds in the environment, or if they just forgot the profile name

christophetd:
Done in [bfbbfda](https://github.com/99designs/aws-vault/pull/864/commits/bfbbfda528def23b2231ea3fd98f3dace780bd2a), let me know how that looks!

christophetd:
@mtibben I've been using this feature locally for a few days, works great for me. Let me know if you require any changes!

christophetd:
@mtibben Eager to get this merged as I'm using it every day and I'm sure it will be useful to other folks. :) Let me know any way I can help!

vault/environmentvariablescredentialsprovider.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"log"
+	"os"
+)
+
+// EnvironmentVariablesCredentialsProvider retrieves AWS credentials available in the OS environment variables
+type EnvironmentVariablesCredentialsProvider struct {
```
mtibben, line 12: There might be some overlap with the AWS SDK here? Should we consider using [`config.NewEnvConfig`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config#NewEnvConfig) and/or [`credentials.StaticCredentialsProvider`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials#StaticCredentialsProvider)

christophetd: Yes, it's a good point. Using `config.NewEnvConfig` allows to be aligned to the logic of the SDK (and hence CLI), and simplifies the code by a great deal. Implemented in [d07990f](https://github.com/99designs/aws-vault/pull/864/commits/d07990fc4d1bba9094483b5ed50af9da0eb3179a) and rebased on latest `master`

christophetd:
Thanks for the review, when's the next release planned for?

mtibben:
Out now! I made one modification to the error message - hope this is OK? 67d1aea780b367767d6667413e6400f40f30eecc

christophetd:
No strong opinion. Having this behavior in usage docs would have made it easier to discover, possibly?

PR #740, opened 2021-03-04
Allow --no-daemonize for the ec2 metadata server so launchd/nohup can daemonize

nitrocode:
For now, I've merged the changes in a [public fork](https://github.com/nitrocode/aws-vault-awesome)

Closes https://github.com/99designs/aws-vault/issues/735 (discussion from @mtibben here)

Thanks to @myoung34 for most of the help in adding the `--no-daemonize` switch.

This allows the `--server` to be nohupped.

```bash
$ make aws-vault-darwin-amd64
$ nohup ./aws-vault-darwin-amd64 \
  exec \
  sso_engineer \
  --server \
  --no-daemonize \
  --debug >/var/log/aws_vault/server.log 2>/var/log/aws_vault/server.errors.log &
$ echo $SHLVL
1
```

Inspiration: https://github.com/99designs/aws-vault/issues/548#issuecomment-614569533

Benefits
* Allows us to `nohup` the app
* If the app is nohupped, a new subshell is no longer dropped so if we kill the aws-vault server, it won't kill our terminal
* Enables the server to be put into a launchd plist. See this [gist](https://gist.github.com/nitrocode/cd864db74a29ea52c7b36977573d01cb). 

Questions:
1. Should I included the proxy and server plist additions to the USAGE.md (PR https://github.com/99designs/aws-vault/pull/733) in this PR?

PR #733, opened 2021-02-27
USAGE.md - Add launchctl plist

nitrocode:
See [custom fork](https://github.com/nitrocode/aws-vault-awesome/) for these changes.

Closes https://github.com/99designs/aws-vault/issues/718

See [Markdown](https://github.com/nitrocode/aws-vault/blob/patch-1/USAGE.md#launchd-plist) version

Issue #361, opened 2019-04-08
Promote proxy server to first class feature


Nuru:
There are a few open issues regarding the proxy server already, and I am opening this new one with new information/issues/use cases and will try to consolidate some of the existing issues here. 

To start with, let me define "proxy server". As [@lox explained](https://github.com/99designs/aws-vault/pull/174#issuecomment-340921440) in #174 , `aws-vault` runs 2 servers: a credentials server at 127.0.0.1:9099, which is the smart server that handles the credentials and manages refreshing them, etc., and a proxy server at 169.254.169.254:80, which simply provides the endpoint that all the AWS tools use but mainly is pretty dumb and just forwards requests to the credentials server. 

This issue is confined to issues with the proxy server.

Because the proxy server binds to privileged port 80, it needs to run (or at least be started) as root, generally either via `sudo` or by running in a `--privileged` Docker container. Because of this, and because by itself it is not a security risk, the proxy server is designed to be started once and left running, in contrast to the credentials server which should only be running while credentials are needed and is thus started and stopped on every `aws-vault exec` command. 

The command to start the proxy server, `aws-vault server` was hidden by #208 because it was confusing people. While I understand the confusing nature of it (although the command is named "server" it does not start the credentials server, which is what might be expected, as in #222 and #224), I think hiding it was a mistake, and I am opening this issue in part to make sure that the command remains available and supported, because I need it.

The reason I need it is that the current automatic starting of the proxy server does not work very well. `checkServerRunning` performs [the simplest check](https://github.com/99designs/aws-vault/blob/d3ffc27e67e886a83ccf5c723a0d2c9ec9db4f64/server/server.go#L80-L83), namely trying to open a TCP connection to 169.254.169.254:80,  to see if the proxy server is already running and if that check succeeds, `aws-vault exec --server` does not start a proxy server because it thinks one is already running.

I believe the fragility of this check, the inability to restart the proxy server described in #360, and the proxy server being hidden from the UI are contributing to the confusion about, if not completely causing, #313, #335, and #338. Promoting the proxy server to a documented, first-class role would help people understand those issues, would open the way for my feature request #345, would have caught #360 earlier, and would allow the removal of the hidden `sudo` command as [lox seems to prefer](https://github.com/99designs/aws-vault/pull/174#issuecomment-340401194) and I definitely prefer. 

I have run into 2 situations where `aws-vault exec --server` fails without providing a good explanation in its output, both due to `checkServerRunning` returning true when it should return false. 
[...]

Issue #345, opened 2019-02-28
Allow multiple `exec`s to share credential server


Nuru:
Similar to #222, I have a situation where I may want to run `aws-vault exec role --server sh` from multiple shells on the same host. This is currently not supported, but I would like it to be.

Currently, if you use the `--server` flag but the server is already running, it fails:
```
aws-vault: error: Failed to start credential server: listen tcp 127.0.0.1:9099: bind: address already in use
```
 I can sort of work around that by checking if the credential server is running and, if it is, setting up the new shell to use it, but that is not a real solution for 2 reasons:
1. I cannot be sure the credentials being served are for the same role I am trying to assume.
2. The credentials server will be stopped when the process that started it finished. 

What I would like is for the second `exec --server` to verify that the existing server is serving credentials for the same role and, if so, use that server. 

Allowing a stand-alone credential server as in #224 would solve the problem of keeping the credential server alive, but would still have the problem of `aws-vault exec role --server` failing even when the role was already active. So I see this as an important addition to #224 that would also be useful without it.

sellers:
Since the "server" simulates the effect of an EC2 Metadata server, it would be hard to have this work in the contents of AWS commands and APIs.   They expect the credential provider to be in the standard provider chain.    (e.g. it opens up tcp4       0      0  169.254.169.254.80     *.*                    LISTEN      linked to :9090).  

I see this being akin to AWS ec2 metadata service supporting multiple profile roles on an EC2 instance.  This should be deferred until that functionality existing with AWS.

Nuru:
@sellers I think you misunderstand my proposal. I am only proposing that a second attempt to start a credentials server _for the same credentials_ reuse the existing server.

Nuru:
@sellers This might also help fix a current problem with`aws-vault`. Currently, when running Docker on Linux on an AWS instance, by default, the Alpine 3.9 containers I run have connectivity to the AWS meta-data server. When I then run `aws-vault --server` it fails. Apparently `aws-vault` sees the AWS meta-data server and thinks it is its own, so it does not start one, but the eventual command fails because it does not have the right credentials. Generally, "assume role" fails because the instance does not have a role. The solution to this is to simply run `aws-vault server` first, but that is undocumented now due to #208.

lox:
I never fathomed that someone would want to run `aws-vault` on aws, would love to hear what you are using it for @Nuru

Nuru:
[...]

PR #876, opened 2022-03-03
Add standalone ECS role server

mtibben:
Adds a new command `aws-vault ecs-role-server` to start a standalone ECS role server.

The ECS server will respond to requests on `/role-arn/YOUR_ROLE_ARN`, making it usable with the `AWS_CONTAINER_CREDENTIALS_FULL_URI` or `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` environment 
variables. These environment variables are used by the AWS SDKs as part of the [default credential provider chain](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default).

The major use-case for this are applications where using `aws-vault exec` workflow is awkward. In particular, using aws-vault on the host with docker images running in a VM. While it's _possible_ to start a container with `aws-vault exec --ecs-server docker run ...`, management of the aws-vault server process does become awkward, and may be better delegated to launchd, brew services, or some other local init script. Using the default ECS IP address `169.254.170.2` means that apps in a docker container no longer need AWS keys at all - instead they just need to specify the role they want to assume, the same way they would in an actual ECS environment.

Here's a diagram with the envisioned architecture for accessing aws-vault from a docker container.

Screen Shot 2022-03-03 at 12 16 15 pm

This use-case is similar to the goal of [amazon-ecs-local-container-endpoints](https://github.com/awslabs/amazon-ecs-local-container-endpoints/blob/mainline/docs/features.md#vend-credentials-to-containers), however the difference here is that the long-lived AWS credentials are getting sourced from your keychain via aws-vault.

The idea of a standalone server has been previously raised in #222, #345, #361, #733 and #740, so keen to also get feedback from folks who've previously requested something similar. I know this doesn't hit all the use-cases contained in those issues yet, but feedback on general direction and architecture of a standalone server would be good.

### DIFF
--- a/cli/ecs-role-server.go
+++ b/cli/ecs-role-server.go
@@ -1,0 +1,82 @@
+package cli
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/99designs/aws-vault/v6/server"
+	"github.com/99designs/aws-vault/v6/vault"
+	"github.com/99designs/keyring"
+	"github.com/alecthomas/kingpin"
+)
+
+type EcsServerCommandInput struct {
+	ProfileName         string
+	Config              vault.Config
+	AuthToken           string
+	Port                int
+	RoleSessionDuration time.Duration
+}
+
+func ConfigureEcsRoleServerCommand(app *kingpin.Application, a *AwsVault) {
+	input := EcsServerCommandInput{}
+
+	cmd := app.Command("ecs-role-server", "Starts a standalone ECS credential server")
+
+	cmd.Arg("profile", "Name of the source profile").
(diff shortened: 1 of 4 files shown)


mtibben:
After playing with this for a few days, I've determined there is no added benefit over using `aws-vault exec --ecs-server`. The execution model of `aws-vault exec` is simpler to reason about and is more secure as the server comes down once the sub-process exits. I'll open another PR to add the role route to `exec` without the standalone server.

mtibben:
See #878

PR #988, opened 2022-08-03
Add gopass prompt

patrickpichler:
The `gopass` prompt works basically the same as the `pass` prompt, but
it is calling the `gopass` specific command to retreive one otp token.

### DIFF
--- a/prompt/gopassotp.go
+++ b/prompt/gopassotp.go
@@ -1,0 +1,35 @@
+package prompt
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strings"
+
+	exec "golang.org/x/sys/execabs"
+)
+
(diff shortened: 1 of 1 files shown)


patrickpichler:
I hope this is fine :) 

Gopass works a bit different to pass, hence its own prompt provider :)

mtibben:
@patrickpichler what are your thoughts on this https://github.com/99designs/aws-vault/pull/1006#issuecomment-1233508808

patrickpichler:
I would favor a script prompt! 

I had a quick look at how one could implement it, but given that I wanted a quick way 
of it working, I decided to go with the less work intensive way of just providing another 
prompt :slightly_smiling_face:

mtibben:
`mfa_process` implemented in #1123. Feedback welcome!

mtibben:
Closed in favour of #1123

PR #968, opened 2022-07-03
Add 1password prompt

sudosubin:
Add `1password` prompt (which is for mfa otp).

- Set prompt
  - Use with environment: `AWS_VAULT_PROMPT=1password`
  - Use with flag: `--prompt=1password`
- Add 1password item with tag `AWS Vault` (or customize with environment variable: `AWS_VAULT_OP_TAG_NAME`)
  image
- Update 1password item url with aws mfa serial
  image
- Also need one-time password section
  - https://support.1password.com/one-time-passwords/

PR #1006, opened 2022-08-13
Add 1password prompt driver

gordonmleigh:
Add support for `--prompt=1password` via the 1password CLI v2.

This is partly inspired by #968 but is cross-platform. I've also added docs. I don't have access to windows so can't test it there.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -33,6 +33,7 @@
     - [Prerequisites](#prerequisites)
     - [Setup](#setup)
     - [Usage](#usage-1)
+  - [Using 1password](#using-1password)
   - [Shell completion](#shell-completion)
   - [Desktop apps](#desktop-apps)
   - [Docker](#docker)
@@ -528,6 +529,33 @@ Further config:
  - `AWS_VAULT_PROMPT=ykman`: to avoid specifying `--prompt` each time
  - `YKMAN_OATH_CREDENTIAL_NAME`: to use an alternative ykman credential
  - `AWS_VAULT_YKMAN_VERSION`: to set the major version of the ykman cli being used. Defaults to "4"
+
+## Using 1password
+
+1password an be used with AWS Vault to generate MFA codes.
+
+### Prerequisites
+ 1. [1password CLI v2](https://1password.com/downloads/command-line/)
+
+You can check the version of the CLI tool you have by running `op --version`.
+
+### Setup
+ 1. Log into the AWS web console with your IAM user credentials, and navigate to  _My Security Credentials_
+ 2. Under _Multi-factor authentication (MFA)_, click `Manage MFA device` and add a Virtual MFA device
+ 3. Save the MFA token to 1password using the QR code
+ 4. Add the tag "aws-vault" to the 1password item
+ 5. Add the MFA ARN to the URLs for the 1password item
+
+### Usage
+Using the `1password` prompt driver, aws-vault will execute `op` to find a matching entry and then generate a OTP.
+
+```shell
+aws-vault exec --prompt 1password ${AWS_VAULT_PROFILE_USING_MFA} -- aws sts get-caller-identity
+```
+
+Futher config:
+ - `AWS_VAULT_PROMPT=1password`: to avoid specifying `--prompt` each time
+ - `AWS_VAULT_OP_TAG_NAME`: to find items using a tag different from the default `aws-vault`
 
 ## Shell completion
 
--- a/prompt/onepassword.go
+++ b/prompt/onepassword.go
@@ -1,0 +1,86 @@
+package prompt
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+
+	"strings"
+
+	exec "golang.org/x/sys/execabs"
+)
+
+// onePasswordUrl represents the type of the `urls`
+type onePasswordUrl struct {
+	Href string `json:"href"`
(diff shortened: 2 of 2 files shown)


mazzy89:
that's a nice long waited-for feature but @gordonmleigh, how did you solve the 1password authentication? your guide does not mention anything about the 1pass cli login.

gordonmleigh:
@mazzy89 the 1password CLI deals with authentication. It supports [Biometric Unlock](https://developer.1password.com/docs/cli/get-started#sign-in) just like the browser extensions.

mazzy89:
Yes I know it deals with authentication.

Ok let me reprahase the question. 
Have you tried to run this in MacOS without the biometric support?

gordonmleigh:
I turned off biometric support in the 1password preferences and as expected it needs a manual login instead. Here is what the 1password docs say on the matter:

```
AUTOMATIC SIGN-IN WITH BIOMETRIC UNLOCK

If biometric unlock is enabled, 1Password CLI will automatically
try to sign in to an account in this order:

 1. The account specified with the '--account' flag.
 2. The account specified in the 'OP_ACCOUNT' environment variable.
 3. The account that was most recently signed in to with 'op signin' 
    in this terminal window.
 4. The account that was most recently signed in to with 'op signin'
    in any other terminal window.

If you haven't signed in to any account yet, you'll be prompted
to select an account.

USE SESSION TOKENS

If biometric unlock is disabled, 1Password CLI uses a session token
to authenticate commands with 1Password.com. Sessions expire after 30 
minutes of inactivity. If you sign in using 'eval $(op signin)', the session
token will automatically be stored in the environment.

If you want to manage the session token manually, it can be obtained by
running:

        op signin --raw

The session token can be passed to other commands through the '--session' flag.
```

I don't think it would be particularly convenient without biometric support, but I don't see that as a reason not to implement it.

mazzy89:
Exactly without biometric support 1p CLI is a nightmare in term of UX.

I have a couple of Macbook with biometric support but for other cases like desktop without the official Apple keyboard then the UX sucks. 

There is a community discussion about this topic where a couple of 1p engineers are involved to improve the flow.

Don't get me wrong. Personally I still vote to have this feature merged but we should warn about the UX in case of absence of biometric support.

hndrewaall:
Chiming in cuz I want this feature and disagree with the above comment about the value of 2FA. 2FA in practice primarily guards against password stuffing/replay attacks. This is still true even if a user is essentially only authenticating with their fingerprint. To capture the seed, you'd need to own the end machine, and at that point the user is screwed anyway.

gordonmleigh:
@mtibeen I had also intended to also raise a PR for such a prompt, though I also agree with @hndrewaall. I actually started work on this already but I didn't raise the PR yet because this repo doesn't look very active.

We would need to add another command line argument to pass the command line to use. I suggest that this should be available for all drivers, so we would have to add an argument to the `prompt.Func` type. For example for this 1password prompt it would be useful to specify which account to use, in case you have several (work, client, home, etc).

I propose that it would be useful to be able to specify the prompt and option in the ini file too, e.g. `aws_vault_prompt` and `aws_vault_prompt_options`. Alternatively, maybe it would be better to specify the options for each prompt, so you could still choose which prompt to use on the fly, but this would complicate ini file and command line parsing.

Let me know your thoughts and I will raise the PR.

edentsai:
maybe something like `credential_process`  in `~/.aws/config`?

- [AWS Docs: Sourcing credentials with an external process](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html)
- [aws-vault: Using `credential_process`](https://github.com/99designs/aws-vault/blob/master/USAGE.md#using-credential_process)

for example, To configure `mfa_process` with a custom command for fetch MFA token:

```
# file: $HOME/.aws/config

[profile demo]
mfa_serial = arn:aws:iam::123456789012:mfa/demo
mfa_process = op item get "aws-demo" --otp --cache
```

when `aws-vault exec demo` require a MFA token,
it will automatically run `mfa_process` command to fetch a MFA token.

mtibben:
> Chiming in cuz I want this feature and disagree with the above comment about the value of 2FA. 2FA in practice primarily guards against password stuffing/replay attacks. This is still true even if a user is essentially only authenticating with their fingerprint. 

MFA is not "primarily" about preventing password stuffing, although that is a clear benefit. It's about diversifying your security surface area into multiple factors.

> To capture the seed, you'd need to own the end machine, and at that point the user is screwed anyway.

No, wrong, the user is not screwed because they are using MFA. This is the whole point of MFA, to avoid security problems like this. If the end machine is owned, you may capture one factor, but you still need a second to compromise AWS.


The original point remains - if you store your password and TOTP together, you've converged 2 factors.

hndrewaall:
In what scenario do you own the machine and be able to log input and capture session keys? You may not be able to reliably reproduce codes, but if you can MITM all traffic the MFA token on your AWS account seems sort of pointless. You can effectively own their AWS account along with anything else they ever access.

The reality is that password stuffing absolutely is the main threat users face, and is 99% of the time a reason there is any benefit to MFA. But, this is the sort of thing one would want data on to answer definitively, and even if one disagrees, hopefully one agrees that it should be up to the user to decide what threat model is appropriate for them.

mtibben:
`mfa_process` implemented in #1123. Feedback welcome!

PR #1013, opened 2022-08-26
add support for using AWS_SESSION_TOKEN

goldenbender:
Currently, when using temporary credentials, you cannot use ```aws-vault exec``` as it doesn't capture the temporary token normally stored under ```AWS_SESSION_TOKEN```, this fixes that by allowing for an (optional) capture of the token both by supplying it & by taking it from your env


cli/add.go
```diff
@@ -27,7 +27,7 @@ func ConfigureAddCommand(app *kingpin.Application, a *AwsVault) {
 		Required().
 		StringVar(&input.ProfileName)
 
-	cmd.Flag("env", "Read the credentials from the environment (AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY)").
+	cmd.Flag("env", "Read the credentials from the environment (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY & AWS_SESSION_TOKEN)").
```
tenitski, line 30:
```suggestion
	cmd.Flag("env", "Read the credentials from the environment (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN)").
```
Using same wording as here https://github.com/99designs/aws-vault/blob/master/USAGE.md

goldenbender: Done, thanks for the idea!

mtibben:
`aws-vault add --env` is designed to capture long-lived credentials and put them into the keychain. What would be the purpose of capturing temporary credentials?

PR #1040, opened 2022-09-16
cli/global: configure TouchID if --biometrics is set

milesbxf:
This adds support for unlocking aws-vault using Touch ID, from https://github.com/99designs/keyring/pull/115

### DIFF
--- a/cli/global.go
+++ b/cli/global.go
@@ -31,6 +31,7 @@ type AwsVault struct {
 
 	keyringImpl   keyring.Keyring
 	awsConfigFile *vault.ConfigFile
+	UseBiometrics bool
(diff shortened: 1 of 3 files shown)


mtibben:
github.com/lox/go-touchid is a prototype. Recent macos keychain APIs have biometrics as a standard option, however there are limitations that make it difficult to integrate

milesbxf:
> github.com/lox/go-touchid is a prototype. Recent macos keychain APIs have biometrics as a standard option, however there are limitations that make it difficult to integrate

Thanks @mtibben . Is there anywhere I can read more about the limitations?

Issue #854, opened 2022-02-01
`aws-vault` suddenly started to report "UnauthorizedException: Session token not found or invalid" every morning


foxylion:
Since a few days I receive every morning a message from `aws-vault` that looks like the following:
```
aws-vault: error: exec: Failed to get credentials for xyz-sso: operation error SSO: GetRoleCredentials,
https response error StatusCode: 401, RequestID: , UnauthorizedException: Session token not found or invalid
```

I frequently use the command `aws-vault exec xyz-sso --json` to generate some new temporary credentials.
Previously this worked flawlessly and opened the SSO page, if my session had expired, but now I only receive the above mentioned error.

To fix it I need to run `aws-vault clear` manually.

---

My `aws-vault` configuration:
```
AWS_VAULT_BACKEND=file
AWS_VAULT_FILE_PASSPHRASE=*some-secret-string*
```

I am using the latest release of AWS Vault
```
aws-vault --debug
2022/02/01 09:12:01 aws-vault 6.4.0
```
I have provided my `.aws/config` (redacted if necessary)
```
[profile xyz-sso]
sso_start_url = https://xyz.awsapps.com/start
sso_region = eu-central-1
sso_account_id = 1234567890
sso_role_name = AdministratorAccess
region = eu-central-1
```
I have provided the debug output using `aws-vault --debug` (redacted if necessary)
```
➜ aws-vault exec xyz-sso --json --debug
2022/02/01 09:13:05 aws-vault 6.4.0
2022/02/01 09:13:05 Loading config file /home/xyz/.aws/config
2022/02/01 09:13:05 Parsing config file /home/xyz/.aws/config
2022/02/01 09:13:05 [keyring] Considering backends: [file]
2022/02/01 09:13:05 Profile 'default' missing in config file
2022/02/01 09:13:05 [keyring] Expanded file dir to /home/xyz/.awsvault/keys/
2022/02/01 09:13:05 [keyring] Expanded file dir to /home/xyz/.awsvault/keys/
2022/02/01 09:13:05 [keyring] Expanded file dir to /home/xyz/.awsvault/keys/
2022/02/01 09:13:05 [keyring] Expanded file dir to /home/xyz/.awsvault/keys/
2022/02/01 09:13:05 [keyring] Expanded file dir to /home/xyz/.awsvault/keys/
2022/02/01 09:13:05 [keyring] Expanded file dir to /home/xyz/.awsvault/keys/
[...]

Issue #722, opened 2021-02-11
aws-vault: error: login: Failed to get credentials for REDACTED: : Session token not found or invalid         status code: 401, request id: 


yuklia:
OS: MacOS Big Sur, Version 11.1 
- [+] I am using the latest release of AWS Vault
v6.2.0
- [+] I have provided my `.aws/config` (redacted if necessary)
```
[profile REDACTED]
sso_start_url = https://REDACTED
sso_region = eu-west-1
sso_account_id = REDACTED
sso_role_name = AWSAdministratorAccess
region = eu-central-1
output = json
[profile REDACTED]
sso_start_url = https://REDACTED
sso_region = eu-west-1
sso_account_id = REDACTED
sso_role_name = AdministratorAccess
region = eu-central-1
output = json
```
- [+] I have provided the debug output using `aws-vault --debug` (redacted if necessary)

```
$ aws-vault --debug login REDACTED
2021/02/11 17:18:55 aws-vault v6.2.0
2021/02/11 17:18:55 [keyring] Considering backends: [keychain]
2021/02/11 17:18:55 Loading config file /Users/yulia/.aws/config
2021/02/11 17:18:55 Parsing config file /Users/yulia/.aws/config
2021/02/11 17:18:55 Profile 'default' missing in config file
2021/02/11 17:18:55 [keyring] Querying keychain for service="aws-vault", keychain="aws-vault.keychain"
2021/02/11 17:18:55 [keyring] Found 2 results
2021/02/11 17:18:55 [keyring] Querying keychain for service="aws-vault", keychain="aws-vault.keychain"
2021/02/11 17:18:55 [keyring] Found 2 results
2021/02/11 17:18:55 [keyring] Querying keychain for service="aws-vault", keychain="aws-vault.keychain"
2021/02/11 17:18:55 [keyring] Found 2 results
2021/02/11 17:18:55 [keyring] Querying keychain for service="aws-vault", account="sso.GetRoleCredentials,R2xvYnVzLVN0YWdpbmc,aHR0cHM6Ly9kLTkzNjcwM2Q2NjIuYXdzYXBwcy5jb20vc3RhcnQv,-62135596800", keychain="aws-vault.keychain"
2021/02/11 17:18:55 [keyring] No results found
2021/02/11 17:18:55 [keyring] Querying keychain for service="aws-vault", account="oidc:https://REDACTED", keychain="aws-vault.keychain"
[...]

PR #1052, opened 2022-10-13
Recreate OIDC token when 401 error is returned

cezarsa:
Occasionally AWS returns 401 unauthorized errors on `GetRoleCredentials` with OIDC tokens that shouldn't have expired yet according to the expiration date returned by AWS. The error message in these situations looks like this:
```
aws-vault: error: exec: Failed to get credentials for <redacted>: operation error SSO: GetRoleCredentials, https response error StatusCode: 401, RequestID: 10a9bfff-90f9-48f4-8ab7-<redacted>, UnauthorizedException: Session token not found or invalid
```
When this happened the only solution was to run `aws-vault clear`.

This error has occurred in the past when AWS returned the wrong expiration date for some time (https://github.com/99designs/aws-vault/issues/854 and https://github.com/99designs/aws-vault/issues/722) and it still happens occasionally to some users, I guess that's due to AWS invalidating tokens on some situations for security reasons.

This PR adds the ability for aws-vault to detect when a 401 error is returned, clear the cached OIDC token and retry the operation once with a new OIDC token.

### DIFF
--- a/vault/ssorolecredentialsprovider.go
+++ b/vault/ssorolecredentialsprovider.go
@@ -5,11 +5,13 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net/http"
 	"os"
 	"time"
 
 	"github.com/99designs/keyring"
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/service/sso"
 	ssotypes "github.com/aws/aws-sdk-go-v2/service/sso/types"
 	"github.com/aws/aws-sdk-go-v2/service/ssooidc"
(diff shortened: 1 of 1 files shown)


vault/ssorolecredentialsprovider.go
```diff
@@ -66,6 +69,23 @@ func (p *SSORoleCredentialsProvider) getRoleCredentials(ctx context.Context) (*s
 		RoleName:    aws.String(p.RoleName),
 	})
 	if err != nil {
+		if cached && p.OIDCTokenCache != nil {
```
mtibben, line 72: there's a possibility that a token is invalidated between the time that it's created, so I wonder if checking `cached` is necessary at all

cezarsa: Checking `cached` ensures `getRoleCredentials` won't be stuck in an infinite recursion loop if some AWS bug was to cause `CreateToken` to return invalid or expired tokens. In these rare situations, I thought it would be safer to return an error immediately instead of possibly retrying forever.

mtibben:
Thanks @cezarsa

PR #1088, opened 2022-12-09
Handle [sso-session] section

dima-vm:
New AWS CLI changed config format for SSO values. Instead of specifying `sso_start_url` and `sso_region`,
it now makes `sso_session` profile param, that points to a separate `[sso-session]` section.

This PR adds support for the `sso_session` param and parsing `[sso-session]` section.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
 /aws-vault
 /aws-vault-*
 /SHA256SUMS
-
-/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /aws-vault
 /aws-vault-*
 /SHA256SUMS
+
+/.idea
--- a/vault/config.go
+++ b/vault/config.go
@@ -165,7 +165,7 @@ func (s ProfileSection) IsEmpty() bool {
 
 // ProfileSections returns all the profile sections in the config
 func (c *ConfigFile) ProfileSections() []ProfileSection {
-	var result []ProfileSection
+	result := []ProfileSection{}
 
 	if c.iniFile == nil {
 		return result
--- a/vault/config.go
+++ b/vault/config.go
@@ -225,11 +225,7 @@ func (c *ConfigFile) SSOSessionSection(name string) (SSOSessionSection, bool) {
 	if c.iniFile == nil {
 		return ssoSession, false
 	}
-	// default profile name has a slightly different section format
 	sectionName := "sso-session " + name
-	if name == defaultSectionName {
-		sectionName = defaultSectionName
-	}
 	section, err := c.iniFile.GetSection(sectionName)
 	if err != nil {
 		return ssoSession, false
--- a/vault/config.go
+++ b/vault/config.go
@@ -137,6 +137,7 @@ type ProfileSection struct {
 	SourceProfile           string `ini:"source_profile,omitempty"`
 	ParentProfile           string `ini:"parent_profile,omitempty"` // deprecated
 	IncludeProfile          string `ini:"include_profile,omitempty"`
+	SSOSession              string `ini:"sso_session,omitempty"`
 	SSOStartURL             string `ini:"sso_start_url,omitempty"`
 	SSORegion               string `ini:"sso_region,omitempty"`
 	SSOAccountID            string `ini:"sso_account_id,omitempty"`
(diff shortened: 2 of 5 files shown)


vault/config.go
```diff
@@ -149,33 +150,43 @@ type ProfileSection struct {
...
+
 func (s ProfileSection) IsEmpty() bool {
 	s.Name = ""
 	return s == ProfileSection{}
 }
 
 // ProfileSections returns all the profile sections in the config
 func (c *ConfigFile) ProfileSections() []ProfileSection {
-	result := []ProfileSection{}
+	var result []ProfileSection
```
dima-vm, line 168: Fixed IDE warning, no semantic change.

vault/vault.go
```diff
@@ -218,7 +218,7 @@ func (t *tempCredsCreator) provider(config *Config) (aws.CredentialsProvider, er
 		if err != nil {
 			return nil, err
 		}
-	} else if config.HasSSOStartURL() {
+	} else if config.HasSSOStartURL() || config.HasSSOSession() {
 		return NewSSORoleCredentialsProvider(t.keyring.Keyring, config)
```
dima-vm, line 222: In proper case, both of them are true. However, `[sso-session]` may lack `sso_start_url`. In that case I feel like it should still go the SSO way.

vault/config.go
```diff
@@ -149,33 +150,43 @@ type ProfileSection struct {
...
+			}
 
-		// ignore the default profile if it's empty
-		if section == defaultSectionName && profile.IsEmpty() {
+			result = append(result, profile)
+		} else if strings.HasPrefix(section, "sso-session ") {
+			// Not a profile
+			continue
+		} else {
+			log.Printf("Unrecognised ini file section: %s", section)
```
mtibben, line 187: this change makes this section difficult to comprehend

vault/config.go
mtibben: why?

.gitignore
```diff
@@ -1,3 +1,5 @@
 /aws-vault
 /aws-vault-*
 /SHA256SUMS
+
+/.idea
```
mtibben, line 5: put this in your global gitignore

vault/vault.go
```diff
@@ -218,7 +218,7 @@ func (t *tempCredsCreator) provider(config *Config) (aws.CredentialsProvider, er
 		if err != nil {
 			return nil, err
 		}
-	} else if config.HasSSOStartURL() {
+	} else if config.HasSSOStartURL() || config.HasSSOSession() {
```
mtibben, line 221: this logic should be moved to the config

vault/config.go
```diff
@@ -205,6 +216,26 @@ func (c *ConfigFile) ProfileSection(name string) (ProfileSection, bool) {
...
+	sectionName := "sso-session " + name
+	section, err := c.iniFile.GetSection(sectionName)
+	if err != nil {
+		return ssoSession, false
+	}
+	if err = section.MapTo(&ssoSession); err != nil {
+		panic(err)
+	}
+	return ssoSession, true
+}
```
mtibben, line 237: why? This doesn't seem to be called from anywhere

vault/config.go
```diff
@@ -307,6 +338,19 @@ func (cl *ConfigLoader) populateFromConfigFile(config *Config, profileName strin
...
+		ssoSection, ok := cl.File.SSOSessionSection(psection.SSOSession)
+		if !ok {
+			// ignore missing profiles
+			log.Printf("[sso-session] '%s' missing in config file", psection.SSOSession)
+		}
+		config.SSOStartURL = ssoSection.SSOStartURL
+		config.SSORegion = ssoSection.SSORegion
+		config.SSORegistrationScopes = ssoSection.SSORegistrationScopes
+	}
 	if config.SSOStartURL == "" {
```
mtibben, line 354: seems SSOStartURL present in a profile should override SSOSession?

vault/config.go
```diff
@@ -532,12 +576,18 @@ type Config struct {
...
+	// SSOStartURL specifies the URL for the AWS IAM Identity Center user portal, legacy option.
 	SSOStartURL string
 
-	// SSORegion specifies the region for the AWS IAM Identity Center user portal.
+	// SSORegion specifies the region for the AWS IAM Identity Center user portal, legacy option.
 	SSORegion string
 
+	// SSORegistrationScopes specifies registration scopes for the AWS IAM Identity Center user portal.
+	SSORegistrationScopes string
+
```
mtibben, line 590: doesn't seem this is ever used by aws-vault?

vault/config.go
mtibben: Or should it be an error when this occurs?

Issue #1100, opened 2022-12-22
[Docs] MFA Instructions do not reflect new multiple MFA device capabilities AWS provides


LouisTsiattalou:
AWS has recently added a feature to allow multiple MFA methods for individual IAM users. Now you can have multiple MFA methods associated with the user, the `mfa_serial` entry no longer follows the `mfa_serial=arn:aws:iam::<acc_id>:mfa/<your.aws.username>` format anymore. This hasn't had any negative effects on the usage of aws-vault, but some of the documentation is now out of date.

kforsthoevel:
I would be interested on how to make aws-vault work with FIDO2 (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_mfa_enable_fido.html#enable-fido-mfa-for-own-iam-user). Does anyone have a clue?

mtibben:
Is this fixed by https://github.com/99designs/aws-vault/pull/1101 ?

kforsthoevel:
IMHO, #1101 is about `TOTP` and not `FIDO2`. So unfortunately is not fixed.

sftim:
I think #1101 fixes it; for example, by deleting the text
> you can usually derive it pretty easily using the format `arn:aws:iam::[account-id]:mfa/[your-iam-username]`

kforsthoevel:
Could you provide an example of a working setup?

sftim:
@kforsthoevel this issue is about updating the docs for `aws-vault` now that users can have more than one MFA device (which might be a physical device, or something virtual).

Because this issue is about multiple MFA devices, it's not the right place to ask for advice about using FIDO2 devices with aws-vault.

carlspring:
Hi,

The docs really need to be updates with concrete examples of how to configure things from scratch for a user without having the specify the AWS access / secret keys and using MFA instead. I've done this previously with a 6.x version a long time ago, but I can't seem to get it to work at all right now. It keeps asking for the keys instead.

PR #1101, opened 2022-12-22
Update Documentation to reflect new Multi-MFA per IAM User functionality on AWS.

LouisTsiattalou:
See https://github.com/99designs/aws-vault/issues/1100

I've gotten AWS Vault to work using a Yubikey and a virtual TOTP device, using AWS' new feature for multiple MFA devices. I've updated the documentation to be consistent with this new capability, and have explained how to get both a Yubikey and a Virtual TOTP device working with the same IAM User.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -522,7 +522,7 @@ You can verify these prerequisites by running `ykman info` and checking `OATH` i
     ykman oath accounts add -t arn:aws:iam::${ACCOUNT_ID}:mfa/${MFA_DEVICE_NAME}
     ```
     replacing `${ACCOUNT_ID}` with your AWS account ID and `${MFA_DEVICE_NAME}` with the name you gave to the MFA device. It will prompt you for a base32 text and you can input the key from step 3. Notice the above command uses `-t` which requires you to touch your YubiKey to generate authentication codes.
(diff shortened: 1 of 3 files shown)


USAGE.md
```diff
@@ -405,7 +405,7 @@ role_arn = arn:aws:iam::33333333333:role/role2
 mfa_serial = arn:aws:iam::111111111111:mfa/tom
 ```
 
-Be sure to specify the `mfa_serial` for the source profile (in the above example `tom`) so that aws-vault can match the common `mfa_serial`.
+For aws-vault <=v4, be sure to specify the `mfa_serial` for the source profile (in the above example `tom`) so that aws-vault can match the common `mfa_serial`.
```
mtibben, line 408: Hmm I don't think the version of aws-vault has any bearing on this. Specifying the mfa_serial in multiple places is required for aws-vault to determine if chaining is possible

LouisTsiattalou: Sure thing, I got that from extrapolating from https://github.com/99designs/aws-vault/pull/1101/files/a03c62bdcff2a324adb364001900fd4a9f06ae8b#diff-58a825c70bccfbe691d7ddbddb133275f4f657de6aeb4a502fa9fce279f8bf8dR414, but I can remove the changes to this line if useful.

LouisTsiattalou: Ok, reverted.

PR #1117, opened 2023-01-28
Add aws_vault_credential_process config option

ash99d:
Adding `aws_vault_credential_process` as a configuration option in aws config. It allows specifying a command that will be executed to generate credentials. This is useful for supporting ad-hoc scenarios such as using as custom implementation of IAM Identity Provider logic that requires calling identity vendor specific APIs. This allows for using the security advantages of aws-vault while using a custom credentials source such as IDP CLI.

### DIFF
--- a/vault/assumerolewithcredentialprocessprovider.go
+++ b/vault/assumerolewithcredentialprocessprovider.go
@@ -1,0 +1,80 @@
+package vault
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"runtime"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	ststypes "github.com/aws/aws-sdk-go-v2/service/sts/types"
+)
+
+// CredentialFromProcessProvider retrieves credentials by running aws_vault_credential_process
+type CredentialFromProcessProvider struct {
+	StsClient                 *sts.Client
+	RoleARN                   string
+	AWSVaultCredentialProcess string
+	Duration                  time.Duration
+}
+
+// Retrieve generates a new set of credentials using the credential provider
+func (p *CredentialFromProcessProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
+	creds, err := p.assumeRole(ctx)
+	if err != nil {
+		return aws.Credentials{}, err
+	}
+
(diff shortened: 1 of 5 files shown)


vault/assumerolewithcredentialprocessprovider.go
```diff
@@ -0,0 +1,80 @@
...
+	creds, err := p.assumeRole(ctx)
+	if err != nil {
+		return aws.Credentials{}, err
+	}
+
+	return aws.Credentials{
+		AccessKeyID:     aws.ToString(creds.AccessKeyId),
+		SecretAccessKey: aws.ToString(creds.SecretAccessKey),
+		SessionToken:    aws.ToString(creds.SessionToken),
+		CanExpire:       true,
```
mtibben, line 37: Can this be determined from the AccessKeyID? Or is this value not important?

vault/config.go
```diff
@@ -127,26 +127,27 @@ func (c *ConfigFile) parseFile() error {
...
+	SourceProfile             string `ini:"source_profile,omitempty"`
+	ParentProfile             string `ini:"parent_profile,omitempty"` // deprecated
+	IncludeProfile            string `ini:"include_profile,omitempty"`
+	SSOStartURL               string `ini:"sso_start_url,omitempty"`
+	SSORegion                 string `ini:"sso_region,omitempty"`
+	SSOAccountID              string `ini:"sso_account_id,omitempty"`
+	SSORoleName               string `ini:"sso_role_name,omitempty"`
+	WebIdentityTokenFile      string `ini:"web_identity_token_file,omitempty"`
+	WebIdentityTokenProcess   string `ini:"web_identity_token_process,omitempty"`
+	AWSVaultCredentialProcess string `ini:"aws_vault_credential_process,omitempty"`
```
mtibben, line 146: Shall we make this the standard `credential_process`?

vault/vault.go
```diff
@@ -166,6 +166,33 @@ func NewAssumeRoleWithWebIdentityProvider(k keyring.Keyring, config *Config) (aw
 	return p, nil
 }
 
+// NewAssumeRoleWithCredentialProcessProvider returns a provider that generates
+// credentials using AssumeRoleWithCredentialProcess
+func NewAssumeRoleWithCredentialProcessProvider(k keyring.Keyring, config *Config) (aws.CredentialsProvider, error) {
```
mtibben, line 171: `NewCredentialFromProcessProvider`

mtibben:
Fixed in #1087

PR #1118, opened 2023-01-29
Enable running server in foreground

jchorl:
This is an attempt to make the `exec --server` functionality more approachable, in line with https://github.com/99designs/aws-vault/issues/361

There were a few peculiarities I noticed:
- `exec --server` quietly runs `sudo` under the hood
- It's not clear if its running or not
- Exiting is difficult, and it seemingly doesn't clean up reliably

I added a `--server-foreground` option, which blocks in the foreground. One issue is that `sudoers` doesn't pass through env vars by default. The CLI UI is verbose, and we can probably do better.
```
$ sudo AWS_VAULT_BACKEND=file AWS_VAULT_PROMPT=ykman ./aws-vault --debug exec dev --server --server-foreground
...
2023/01/29 05:03:12 EC2 Instance Metadata endpoint proxy server running on 169.254.169.254:80
```

ctrl-c here properly exits and cleans up. This aligns well with a bunch of my workflows, where I want to kick up aws creds in one terminal, do some other tasks, and then shut down the server.

Some of the benefits of this implementation:
- The proxy gets properly torn down when the process exits
- You must explicitly call with `sudo`
- It is obvious when the server is running - it's running in the foreground and blocking
- The webservers get gracefully turned down per go guidance
- It eliminates multiple processes trying to figure out if another has set up the proxy
- It eliminates subshell calls. Go has good context management and goroutine management. We should leverage those.
- In line with eliminating subshell calls, we get to delete some platform-dependent subshelling (yay!)

I have done preliminary testing locally, and it works. I haven't rigorously tested the various flows. I would be happy to invest some time in testing, or ideally write some automated tests (I see there aren't many today).

Anyway, I wanted to align on direction, interface and implementation before writing tests. Let me know if the approach looks good and I'd be happy to take feedback/polish things up.


mtibben:
1. What's the use-case? What are the listed benefits in service of?
2. Various architectures were previously discussed, and the [2-part design](https://github.com/99designs/aws-vault/pull/174#issuecomment-340922427) was determined to best for the goals of aws-vault. Have you considered the discussion there?

jchorl:
> 1. What's the use-case? What are the listed benefits in service of?

It's a good question. For better or worse, a common workflow is to fire up a cred server in a terminal and iterate on things in separate terminals. For example, you might pull some files down from s3, run some processing on them (some of which may not support env vars so you need a server), push outputs back up to S3. It allows you to hold multiple tmux panes using the same server, multiple applications, etc. I understand it is an explicit anti-goal to have people leave these things running all day in the background. However the ergonomics of continually prefixing with `aws-vault ...` gets toilsome and devs just start long-running shells with creds instead (or servers). More on this below.

>     2. Various architectures were previously discussed, and the [2-part design](https://github.com/99designs/aws-vault/pull/174#issuecomment-340922427) was determined to best for the goals of aws-vault. Have you considered the discussion there?

I actually hadn't seen that thread. There's some good points. General thoughts:

> The more I think about it, I think the two part server design is better. It prevents people from running long lived credential servers, which basically negate the security value of aws-vault.

Does it? If you properly specify your command to `aws-vault`, sure. But if you just run `aws-vault exec profile --server` you get a weird middleground where you're dropped into a new shell, with a magic server up in the background. It's tough to see what's going on, and tough to shut it down after.

IMO, more effective is to limit the lifetime on assume-role tokens to force the user to re-auth periodically. I think the tradeoffs of issuing semi-long-lived tokens are understood (we're talking <12h anyway here), and it's understood that once tokens are issued there's some risk to them anyway. Should we make it harder to steal them? Certainly. But it's a usability/dev-prod tradeoff.

> I generally dislike the fact that we call sudo ourselves anyway. We could drop that and leave it up to the user to use sudo as makes sense.

+1 to this. Better to force the user to use `sudo`.

Other points of interest:
- This doesn't put other features in the background. It moves _everything_ to the foreground. It makes it obvious to the user that the server is running. It blocks. If anything fails, it tears it all down, gracefully.
- It provides graceful shutdown that _actually cleans up_. No lingering iptables rules (unless you kill -9 it). As I understand it, the current situation is that you can either:
    1. `curl` `/stop`, which will trigger [this handler](https://github.com/99designs/aws-vault/blob/2348df3aed190d8c1a186482112252b1e4db83b5/server/ec2proxy.go#L53)
    2. _manually_ `SIGTERM` `aws-vault proxy` proc which triggers [this handler](https://github.com/99designs/aws-vault/blob/2348df3aed190d8c1a186482112252b1e4db83b5/cli/proxy.go#L32)

    Sure, this isn't a huge deal. It's definitely odd that this leaves lingering configs and there's an `exit 0` buried in a server handler.

Anyway, happy to go either way here - just thought I'd share some self-perceived usability improvements based on workflows I've seen as food for thought.

mtibben:
> I understand it is an explicit anti-goal to have people leave these things running all day in the background

Exactly. It's an explicit goal of aws-vault to remove footguns and _prevent_ developers doing this. Switching to the --ecs-server as the default --server behaviour will progress us towards this goal

> IMO, more effective is to limit the lifetime on assume-role tokens to force the user to re-auth periodically

That's the opposite goal of `--server`. Server is there so you _don't_ have to reauth.

> It provides graceful shutdown that actually cleans up

What is not being cleaned up? Is there a bug?

mtibben:
btw I believe `--ecs-server` has the exact behaviour you're looking for

server/ecsserver.go
```diff
@@ -105,8 +106,20 @@ func (e *EcsServer) AuthToken() string {
 	return e.authToken
 }
 
-func (e *EcsServer) Serve() error {
-	return e.server.Serve(e.listener)
+func (e *EcsServer) Serve(ctx context.Context) error {
+	group, ctx := errgroup.WithContext(ctx)
+	group.Go(func() error {
+		return startProxy(ctx)
+	})
```
mtibben, line 113: ecs server doesn't use the proxy

mtibben:
Hey @jchorl, does `--server` in the latest v7 beta work more in line with your expectations?

mtibben:
So I've been looking into this a little more. The separate proxy server started via sudo is there because if `sudo` was to be used to run the main aws-vault process, the user accessing the secure keystore / keychain would be the root user instead of the actual user. Therefore the process running under sudo is scoped only to bind to the required ip/port, then proxy to the user's process. Once the proxy server is running and bound to 169.254.169.254:80, it doesn't require restarting and re-sudo-ing no matter how many times the main aws-vault process is stopped or started

However it's certainly not ideal that the proxy process sticks around indefinitely. I wonder if it should shut itself down after a period of inactivity.

jchorl:
> Hey @jchorl, does `--server` in the latest v7 beta work more in line with your expectations?

Yeah!! I checked out your additions - pretty awesome :). I think they go a long way to encourage least-surprising behaviour and eliminate confusion.

> Therefore the process running under sudo is scoped only to bind to the required ip/port, then proxy to the user's process.

Yeah, this is a really fair point.

> However it's certainly not ideal that the proxy process sticks around indefinitely. I wonder if it should shut itself down after a period of inactivity.

Hmm. I worry a time bomb might actually be more surprising.

At this point I think this PR is fine to close out. I also did some digging though to see if there is a better way.

I believe you can use [capabilities](https://superuser.com/a/892391) (at least on linux) to add `CAP_NET_ADMIN` to the `ip addr ...` calls, `CAP_NET_BIND_SERVICE` to the proxy net listen to grab port 80, and not pass either of these to the child process.

Mostly for my own edification, I whipped up an example:

```golang
package main

import (
	"fmt"
	"log"
	"net/http"
	"os/exec"
	"syscall"
)

// https://github.com/torvalds/linux/blob/f3a2439f20d918930cc4ae8f76fe1c1afd26958f/include/uapi/linux/capability.h#L203
// inspiration from https://github.com/golang/go/blob/af6bb3f9b2ebfe9a53c3de49ec880ca647dadccf/src/syscall/exec_linux_test.go#L609
const CAP_NET_ADMIN = 12

const (
	ec2MetadataEndpointAddr = "169.254.169.254:80"
)

func installEc2EndpointNetworkAlias() ([]byte, error) {
	cmd := exec.Command("ip", "addr", "add", "169.254.169.254/24", "dev", "lo", "label", "lo:0")
	cmd.SysProcAttr = &syscall.SysProcAttr{
		AmbientCaps: []uintptr{CAP_NET_ADMIN},
	}

	return cmd.CombinedOutput()
}

func removeEc2EndpointNetworkAlias() ([]byte, error) {
	cmd := exec.Command("ip", "addr", "del", "169.254.169.254/24", "dev", "lo", "label", "lo:0")
	cmd.SysProcAttr = &syscall.SysProcAttr{
		AmbientCaps: []uintptr{CAP_NET_ADMIN},
	}

	return cmd.CombinedOutput()
}

func main() {
	out, err := installEc2EndpointNetworkAlias()
	if err != nil {
		log.Fatalf("installing network alias: %s", out)
	}

	defer removeEc2EndpointNetworkAlias()

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintf(w, "Hi")
	})
	log.Fatal(http.ListenAndServe(ec2MetadataEndpointAddr, nil))
}
```

```shell
$ go build main.go
$ ./main
2023/02/27 01:33:23 listen tcp 169.254.169.254:80: bind: cannot assign requested address
$ sudo setcap CAP_NET_ADMIN,CAP_NET_BIND_SERVICE=+eip ./main
$ ./main
# ^ that worked
```

I suspect this would be somewhat nightmarish to support in a commonly used library. Although it does look like capabilities are very well supported and have been for a long time. I'm not sure if you could add the caps to the binary prior to distributing, or if every machine needs to apply them at installation.

This looks very promising to avoid sudo in general though, and avoid elevating subprocess permissions, while getting the perms you need to do what the user intended. I wonder if you could even avoid `CAP_NET_ADMIN` for a finer-grained capability.

Anyway, this is a bit out of my depths but was a really neat learning experience and hopefully provides some food for thought.

PR #1123, opened 2023-02-03
Add `mfa_process` config

mtibben:
Adds`mfa_process` to the aws config that works similarly to `credential_process`.

This will allow us to remove the "pass" prompt while allowing flexibility for those who wish to use a password manager to generate TOTP tokens.

Note that use of `mfa_process` is generally not considered best practice - instead it is recommended that you use a dedicated device for MFA.


mtibben:
Environment variable seems reasonable, I'd prefer to avoid templating. Feel free to open a PR

PR #1139, opened 2023-02-17
update docs for `credential_process` section

ArjunDandagi:
aws-vault latest version:
- there is no subcommand called export 
- updated the docs to reflect the proper commands to use in `credential_process`


mtibben:
Usage doc is for master branch. You can view the usage doc for git tags too

PR #1146, opened 2023-02-20
Add region to `aws-vault export` output

### LINKED ISSUE #1145: Include aws region when exporting env vars
#1135 introduces the `aws-vault export` command which can print credentials as environment variables.

To support profiles in default aws regions, could `AWS_DEFAULT_REGION` and `AWS_REGION` also be exported perhaps? 🙏

mtibben:
Fixes #1145

### DIFF
--- a/cli/export.go
+++ b/cli/export.go
@@ -113,9 +113,9 @@ func ExportCommand(input ExportCommandInput, f *vault.ConfigFile, keyring keyrin
 	} else if input.Format == FormatTypeExportINI {
 		return printINI(credsProvider, input.ProfileName, config.Region)
 	} else if input.Format == FormatTypeExportEnv {
-		return printEnv(input, credsProvider, "export ")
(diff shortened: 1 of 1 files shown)


cli/export.go
```diff
@@ -188,5 +208,9 @@ func printEnv(input ExportCommandInput, credsProvider aws.CredentialsProvider, p
 	if creds.CanExpire {
 		fmt.Printf("%sAWS_CREDENTIAL_EXPIRATION=%s\n", prefix, iso8601.Format(creds.Expires))
 	}
+	if region != "" {
+		fmt.Printf("%sAWS_REGION=%s\n", prefix, region)
```
tekumara, line 212: Could we also set AWS_DEFAULT_REGION too in order to support [boto3 python clients](https://docs.aws.amazon.com/sdkref/latest/guide/feature-region.html)?

tekumara (commented):
thank you for adding this!

PR #1150, opened 2023-02-21
Allow login with master creds in environment

mtibben:
Use a federation token when master creds are sourced from the environment

Fixes #886


christophetd:
I tested this branch while being authenticated as an IAM role (after a successful `aws-vault exec`), and it doesn't seem to be working as I would expect:

```
$ aws-vault exec foo
$ aws sts get-caller-identity
{
    "UserId": "XXX:christophetd",
    "Account": "012345678912",
    "Arn": "arn:aws:sts::012345678912:assumed-role/admin/christophetd"
}
$ git checkout login-with-master-creds
$ go run . login
aws-vault: error: login: operation error IAM: GetUser, https response error StatusCode: 400, RequestID: 78c10110-6778-44cc-ac8a-e41ca01b3613, api error ValidationError: Must specify userName when calling with non-User credentials
```

USAGE.md
```diff
@@ -331,11 +331,12 @@ You can use the `aws-vault login` command to open a browser window and login to
 $ aws-vault login work
 ```
 
-If you have temporary STS credentials already available in your environment, you can have aws-vault use these credentials to sign you in.
-This is useful when you had to use something else than aws-vault to retrieve temporary credentials:
+If you have credentials already available in your environment, aws-vault will use these credentials to sign you in to the AWS console.
```
christophetd, line 334: Any way we can also support additional credentials source? (e.g. `--server` or `--ecs-server`)

mtibben:
> I tested this branch while being authenticated as an IAM role (after a successful aws-vault exec), and it doesn't seem to be working as I would expect:

@christophetd OK I think I fixed that now

USAGE.md
mtibben: not at this time

christophetd:
Works well with env variables now! (it doesn't with `--server` or `--ecs-server` FWIW). Great stuff!

PR #1168, opened 2023-02-28
Support `login` from profile with `credential_process`

### LINKED ISSUE #1134: [Feature Request] Support `aws-vault login` for `credential_process` returning temporary credentials
Similar to #1087, I would like to use aws-vault to start console sessions from cached credentials created by an external process.

christophgysin:
Fixes #1134

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -106,7 +106,7 @@ func LoginCommand(input LoginCommandInput, f *vault.ConfigFile, keyring keyring.
 	} else {
 		// Use a profile from the AWS config file
 		ckr := &vault.CredentialKeyring{Keyring: keyring}
-		if config.HasRole() || config.HasSSOStartURL() {
+		if config.HasRole() || config.HasSSOStartURL() || config.HasCredentialProcess() {
 			// If AssumeRole or sso.GetRoleCredentials isn't used, GetFederationToken has to be used for IAM credentials
 			credsProvider, err = vault.NewTempCredentialsProvider(config, ckr)
 		} else {


vault/vault.go
```diff
@@ -269,14 +269,8 @@ func NewTempCredentialsProvider(config *Config, keyring *CredentialKeyring) (aws
...
-func NewFederationTokenCredentialsProvider(ctx context.Context, profileName string, k *CredentialKeyring, config *Config) (aws.CredentialsProvider, error) {
-	credentialsName, err := FindMasterCredentialsNameFor(profileName, k, config)
-	if err != nil {
-		return nil, err
-	}
-	masterCreds := NewMasterCredentialsProvider(k, credentialsName)
-
-	return NewFederationTokenProvider(ctx, masterCreds, config)
+func NewFederationTokenCredentialsProvider(ctx context.Context, profileName string, keyring *CredentialKeyring, config *Config) (aws.CredentialsProvider, error) {
+	return NewTempCredentialsProvider(config, keyring)
```
mtibben, line 273: A bit confused by this - what's the point of the `NewFederationTokenCredentialsProvider` function at all? You're not creating a federated token at all anymore

mtibben:
I've taken this further in #1170 - are you able to try that branch?

vault/vault.go
christophgysin1: Neither did the previous code. This is simply returning a credentials provider that shall be used to create the federation token.

The previous code went up the chain to the master credentials. This PR changes this to use temporary credentials as defined in the config.

Of course, we can also completely eliminate `NewFederationTokenCredentialsProvider` if you don't think there might be a need for different behaviour for creating federation tokens.

christophgysin1: Either way, found the right place to change to behaviour for configurations that use `credential_process`. Please see that latest commit.

PR #1170, opened 2023-03-01
Only use federation token if no session token exists

### LINKED ISSUE #1134: [Feature Request] Support `aws-vault login` for `credential_process` returning temporary credentials

mtibben:
Only create a federation token if no session token exists.

Fixes #1134


mtibben:
Hmm this doesn't work for profiles without an assume_role ..... It seems like the console only works with assume role or federation tokens

PR #1178, opened 2023-03-06
Web identity profiles with role ARNs are valid

### LINKED ISSUE #1177: Web identity + role fails to validate in the 7.0.0 release
- [X] I am using the latest release of AWS Vault
- [X] I have provided my `.aws/config` (redacted if necessary)

After upgrading to 7.0.0, we're seeing errors:

```
aws-vault: error: exec: Error getting temporary credentials: profile 'XXXX' has more than one source of credentials
```

This appears to be related to some new validation code(https://github.com/99designs/aws-vault/blob/ec5e53c91b9990c39c0af69de45f21e436abaa23/vault/config.go#L684-L709), which counts both a `role` and a `web_identity_token_process` as two independent sources. However, the web identity provider _requires_ both of these to be set: https://github.com/99designs/aws-vault/pull/587#issue-616928047

I'm not sure what the motivation with the new validation check was, but I'm guessing we need to only increment the count when a role exists when the web identity file/process is not set.

lstoll-sfdc:
Profiles that specify web identity process or files require the role ARN to be set as well. Since the 7.0.0 release, this has been failing validation because it is now considered to have two sources of credentials. This fixes the validation to correctly consider this as a profile with one source credential.

Fixes #1177

### DIFF
--- a/vault/config.go
+++ b/vault/config.go
@@ -698,7 +698,12 @@ func (c *ProfileConfig) Validate() error {
 	}
 	if c.HasSourceProfile() {
 		n++
-	} else if c.HasRole() {
+	}
+	if c.HasRole() &&
+		// these cases require the role to be set in addition, so it's part of
+		// their credential.
+		!c.HasSourceProfile() &&
+		!c.HasWebIdentity() {
 		n++
 	}
 	if n > 1 {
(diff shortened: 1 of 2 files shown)


mtibben:
Thanks @lstoll-sfdc

Issue #1054, opened 2022-10-14
cannnot start AWS-StartSSHSession with MFA


azusa:
- [x] I am using the latest release of AWS Vault

```
6.6.0-Homebrew
```

```
ProductName:	macOS
ProductVersion:	12.6
BuildVersion:	21G115
```

- [x] I have provided my `.aws/config` (redacted if necessary)

```
[profile ssm-user2]

region = ap-northeast-1
output = json
mfa_serial=arn:aws:iam::491673950671:mfa/ssm-user
```

- [x] I have provided the debug output using `aws-vault --debug` (redacted if necessary)

```
[]:$ ssh ssm                                                                [~]
2022/10/15 00:14:45 aws-vault 6.6.0-Homebrew
2022/10/15 00:14:45 Loading config file /Users/azusa/.aws/config
2022/10/15 00:14:45 Parsing config file /Users/azusa/.aws/config
2022/10/15 00:14:45 [keyring] Considering backends: [keychain]
2022/10/15 00:14:45 Profile 'default' missing in config file
2022/10/15 00:14:45 [keyring] Querying keychain for service="aws-vault", keychain="aws-vault.keychain"
2022/10/15 00:14:45 [keyring] Found 3 results
2022/10/15 00:14:45 profile ssm-user2: using stored credentials
2022/10/15 00:14:45 profile ssm-user2: using GetSessionToken (with MFA)
2022/10/15 00:14:45 [keyring] Querying keychain for service="aws-vault", keychain="aws-vault.keychain"
2022/10/15 00:14:45 [keyring] Found 3 results
2022/10/15 00:14:45 [keyring] Removing keychain item service="aws-vault", account="sts.GetSessionToken,(Removed)1665757089", keychain "aws-vault.keychain"
2022/10/15 00:14:48 [keyring] Querying keychain for service="aws-vault", keychain="aws-vault.keychain"
2022/10/15 00:14:48 [keyring] Found 2 results
2022/10/15 00:14:48 [keyring] Querying keychain for service="aws-vault", account="sts.GetSessionToken,(Removed),-62135596800", keychain="aws-vault.keychain"
2022/10/15 00:14:48 [keyring] No results found
Enter MFA code for arn:aws:iam::491673950671:mfa/ssm-user: 2022/10/15 00:14:48 Looking up keyring for 'ssm-user2'
[...]

PR #1184, opened 2023-03-09
Fix hang on invisible tty when using 2FA where prompt=terminal in some edge cases

onnos:
This implements the fix suggested by @ngyuki in https://github.com/99designs/aws-vault/issues/1054. When using `prompt=terminal` (which we use to support both Linux and Mac using the same workflow and config, without `osascript` involvement), we used to have a workaround in `.aws/config` that would redirect the terminal prompt to stdout in most cases by adding ` 2> $(tty)` to the end of the `aws-vault` `credential_process` call, like this:

```
credential_process=sh -c 'aws-vault --prompt terminal export base --duration 12h --format=json 2> $(tty)'
```

With this patch we can remove that redirect and the prompt always shows up, even when called through the SDK via `aws --profile somerole s3 ls`, for example. Also removes the creation of spurious files named `not a tty` when `prompt=terminal` is forced. 

This expands on https://github.com/99designs/aws-vault/pull/1149 a bit, where instead of just preferring a non-terminal prompt without an active tty it should show up in all cases anyway when specifically requested.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,7 @@ require (
 	github.com/dvsekhvalnov/jose2go v1.5.0 // indirect
 	github.com/godbus/dbus v0.0.0-20190726142602-4481cbc300e2 // indirect
 	github.com/gsterjov/go-libsecret v0.0.0-20161001094733-a6f4afe4910c // indirect
+	github.com/mattn/go-tty v0.0.4 // indirect
 	github.com/mtibben/percent v0.2.1 // indirect
 	github.com/xhit/go-str2duration v1.2.0 // indirect
 	golang.org/x/sys v0.5.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -52,8 +52,14 @@ github.com/jmespath/go-jmespath/internal/testify v1.5.1/go.mod h1:L3OGu8Wl2/fWfC
 github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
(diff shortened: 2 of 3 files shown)


mtibben:
Are there are any weird edge cases to this? Is it possible to have a terminal without a TTY these days? Does this work on Windows and other OSs?

onnos:
> Are there are any weird edge cases to this? 

I am not able to find any, in fact this only fixes an edge case. This particular path is only prompted when `prompt=terminal` is forced by greybeards like myself who do not want to leave the confines of the terminal to go through 2FA. In the normal flow the "workaround" of popup GUI prompts are used. While the GUI popups are good UI for many people, I'd argue that allowing a full-terminal flow when specifically requested is a nice thing to do. With v7, this was made possible by the changes in https://github.com/99designs/aws-vault/pull/1149, however one edge case still remains - specifically when `credential_process` is called in the background **and** `prompt-terminal` is specified, the `aws-vault` process has no access to any foreground ttys and still swallows the prompt.

Specifically:
with `~.aws/config` having the following sample configuration:
```
[profile base]
mfa_serial=arn:aws:iam::121212121212:mfa/username

[profile base-session]
credential_process=aws-vault7 --prompt terminal export base --duration 12h --format=json

[profile role]
role_arn=arn:aws:iam::242424242424:role/cross-account-role
source_profile=base-session
```

Running `aws s3 ls --profile role` will appear to hang. When you hit enter, you will get `Error when retrieving credentials from custom-process:` because the prompt was invisible. The version with this patch will show the prompt. Any other flow remains untouched.

> Is it possible to have a terminal without a TTY these days? 

I don't think so? I might be misunderstanding the question, but to my belief anything without a TTY cannot be considered a terminal. TTY's are essential for I/O and signalling. As [The TTY demystified](http://www.linusakesson.net/programming/tty/) says though, "the TTY subsystem — while quite functional from a user's point of view — is a twisty little mess of special cases."

> Does this work on Windows and other OSs?

I don't have a Windows machine to test on. That said, this works on Macos and Linux machines and I can't imagine it not working in any terminal emulator. Again though, this only targets one particular flow where `prompt=terminal` is forced - when that is not enabled the normal prompt flows will still trigger and work as they do currently.

I've rebased my branch and fixed the linter error that I saw show up in the Github Actions. Thanks for your quick review so far!

Issue #1198, opened 2023-03-20
[ecsserver] Make address configurable


nVlast:
- [x] I am using the latest release of AWS Vault
- [ ] I have provided my `.aws/config` (redacted if necessary)
- [ ] I have provided the debug output using `aws-vault --debug` (redacted if necessary)

I'm trying to configure the ECS server with Docker as shown [here](https://github.com/99designs/aws-vault/blob/master/USAGE.md#docker) on Linux (Debian 11). I didn't manage to make it work. As I tried to find out what's going on I tracked the issue down to [the ecs server listening on the `127.0.0.1` interface](https://github.com/99designs/aws-vault/blob/master/server/ecsserver.go#L70). When I changed that to the address of my Docker daemon I managed to get it working. I'm thinking that maybe the different way docker is set up in these platforms causes this but I wasn't able to find any related documentation.

Would you consider making the ecs server address configurable or does this pose a security risk? Maybe we could somehow query the os to get the Docker daemon's address instead?

I saw that #810 seems similar but it refers to `--server` and it seems people there managed to make it work on Linux at some point.

Also, if that feature is meant to work only on macOS that's ok.

frco9:
I have exactly the same issue, I was about to open one. I made it work on MacOS with docker Desktop but it is not working on Linux with docker-engine. 
It seems the issue is that the ecs-server  is only listening on the loopback (127.0.0.1) which is not available from docker on Linux. 
@mtibben, an easy fix would be instead of making it configurable to make the server listen on (0.0.0.0) which is what is recommended here : https://stackoverflow.com/a/73859760/5024020

I've tested this with a random python server listening locally on 0.0.0.0:8000 and it is reachable from the docker container. While if it is listening on 127.0.0.1:8000 it is not.

mtibben:
It seems to me the problem is that the proxy is [listening on `host.docker.internal`](https://github.com/99designs/aws-vault/blob/master/contrib/_aws-vault-proxy/main.go#L18
) but this may not be present on linux?

Have a read of https://stackoverflow.com/questions/24319662/from-inside-of-a-docker-container-how-do-i-connect-to-the-localhost-of-the-mach.

Does adding
```
extra_hosts:
    - "host.docker.internal:host-gateway"
```
to the docker-compose file work?

frco9:
Yes as I said, we need to add the `extra_hosts` entry in the docker compose file otherwise it won't work. 
[...]

PR #1199, opened 2023-03-22
fix: listen on all interfaces for ecs-server

frco9:
This MR fixes https://github.com/99designs/aws-vault/issues/1198

Currently eco-server only listens to the loopback address making it unreachable from docker containers on Linux.


mtibben:
What are the potential security side-effects here? _Should_ we be listening on all interfaces?

frco9:
That's a good question and it's a bit out of my domain of expertise. It would mean that if a user does not have a proper firewall, any user knowing the ip of the user, the port the ecs-server is working and the auth token, would be able to request sessions token from the user's ecs-server. 
I agree it's not perfect as it increase attack surface, although as the port and auth token is supposed to only be accessible from the aws-vaul subprocess, it is not that obvious to find them ? 

Otherwise, this could be made configurable as initially proposed in the issue ? Don't you have this access denied issue on Linux ?

mtibben:
No, I use macOS. Perhaps we add a `--listen-address` flag that defaults to `127.0.0.1`?

frco9:
@mtibben, I've made the changes, I'm not really used to develop in Golang. I've successfully compile and test it locally but please tell me if things can be refactored or if I've missed something.

frco9:
@mtibben could you take a look ?

mtibben:
Tried this out, however realised that the aws cli (not sure about the sdk) doesn't actually allow arbitrary hosts
```
Unsupported host '::'.  Can only retrieve metadata from these hosts: 169.254.170.2, localhost, 127.0.0.1
```

So the problem seems confined to the docker scenario in the issue. Can you see if the solution described at https://github.com/99designs/aws-vault/issues/1198#issuecomment-1484249196 works?

frco9:
Issue is just that `BaseURL` is not returning a valid https url that could be fetched it returns `http://[::]:{port}`. 

This seems linked to https://github.com/golang/go/issues/9334. To "fix" that and force to bind on ipv4 I have changed the network type from `tcp` to `tcp4` 

Now base url returns the proper value: 
```
AWS_CONTAINER_CREDENTIALS_FULL_URI=http://127.0.0.1:{port}
```

RemaniTinpo:
:wave: 

Hi,

I'm having the same issue as a Linux user.
Any chance that this PR would be merged?

Thanks a lot, @frco9 and @mtibben

awilkins:
Hey there!

As per my comment in the issue, on Linux, the safest value to pass to this argument would be the output of this

```sh
docker network inspect bridge | jq '.[0].IPAM.Config[0].Gateway'
```

In conjunction with adding this to your docker-compose.

```yaml
    extra_hosts:
     - host.docker.internal:host-gateway
```

`0.0.0.0` isn't _terrible_ because the ECS server requires an auth token for access, which is passed down to the child process of `aws-vault exec` via the environment. Because this isn't available to a hypothetical adversary elsewhere in the network, in theory you're safe. In practice, someone might find an exploit or DoS attack for the ECS server implementation, so it's safer to only listen to interfaces you're expecting traffic from.

So the full command (with this option added) would be

```sh
aws-vault \
    --ecs-server \
    --listen-address $(docker network inspect bridge | jq '.[0].IPAM.Config[0].Gateway') \
    my-profile -- \
    docker compose up --build aws-vault-proxy
```

PR #1200, opened 2023-03-26
Support for aws-vault exec --server on WSL

alex-budkar-amplitude:
Currently when running on WSL the only supported prompt driver is terminal Terminal prompt is not compatible with the server mode

To workaround we run credentials server on windows host and take advantage of windows cred storage and UX 
On WSL linux we run proxy command that talks to the credentials server running on host Because we don't need to run proxy on the windows host we made it optional If proxy is disabled we would not need privilege elevation

On windows:
```
aws-vault exec --ec2-server --wsl --no-proxy ${PROFILE_NAME}
```
On linux:
```
aws-vault proxy --wsl
```
After that all linux processes will be authenticate through windows host

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -27,6 +27,8 @@ type ExecCommandInput struct {
 	Args             []string
 	StartEc2Server   bool
 	StartEcsServer   bool
+	DontStartProxy   bool
+	IsWsl            bool
 	Lazy             bool
 	JSONDeprecated   bool
 	Config           vault.ProfileConfig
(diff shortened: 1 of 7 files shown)


mtibben:
> Currently when running on WSL the only supported prompt driver is terminal

Can't you install zenity or kdialog?

> To workaround the issue we run credentials server on windows host and take advantage of windows cred storage and UX

This seems similar to the [docker setup in the USAGE.md](https://github.com/99designs/aws-vault/blob/master/USAGE.md#docker). 

However the solution you've created in this PR is bespoke to wsl, a hyper-specific solution not really suitable to a general tool.

My suggestions 
 1. Use ECS server instead. It's more flexible for this use-case
 2. Set up proxying and networking outside of aws-vault, similar to the docker solution

And then this way you can add this use-case solution to contrib

PR #1214, opened 2023-07-14
Suggest using the Homebrew bottle rather than the cask

orf:
Is there a particular reason why we suggest using the casked version of `aws-vault` in the readme?

Homebrew bottles `aws-vault` and sets up completions for you upon install, which is quite handy.

If we'd like to keep `--cask`, shall we also note that there is a non-casked version as well?


FernandoMiguel:
I believe the reason to use cask is that is signed while the other is not

PR #1245, opened 2024-05-15
Update Yubikey scripts to support device name

russelldavies:
It seems the limitation of having to name virtual MFA devices the same as the IAM username has been removed.


mbevc1:
Maintained fork: https://github.com/ByteNess/aws-vault

Issue #1241, opened 2024-04-19
mfa_process does not seem to be working with Windows


kariya1111:
- [x]  I am using the latest release of AWS Vault: v7.2.0.
- [x]  I have provided my .aws/config (redacted if necessary)
- [ ]  I have provided the debug output using aws-vault --debug (redacted if necessary)

emvironment:
Windows 10, command prompt

.aws/config file:

```
[profile common]
region=ap-northeast-1
output=json
mfa_serial=arn:aws:iam::111111111111:mfa/xxxxxxxxx
mfa_process=op read "op://Private/xxxxxxxx?attribute=otp"
role_session_name=xxxxxxxxx`
```

executed command:
`aws-vault exec common --json --prompt=terminal
`

result:
`aws-vault: error: exec: Failed to get credentials for common: process provider: exec: "/bin/sh": file does not exist`

note:
If I delete `mfa_process=op read "op://Private/xxxxxxxx?attribute=otp"`from config, it worked.

I believe the issue lies in specifying only "/bin/sh" in the code below.
https://github.com/99designs/aws-vault/blob/e22aea12b03e8ce036e9af87dda9303806fa2a9e/vault/mfa.go#L50

I think it will work if it support multiple OS like the code below.
https://github.com/99designs/aws-vault/blob/e22aea12b03e8ce036e9af87dda9303806fa2a9e/cli/exec.go#L327-L331

mbevc1:
Solved in the fork: https://github.com/ByteNess/aws-vault/pull/50

PR #1255, opened 2024-09-30
Fix mfa_process compatibility issue on Windows

uubk:
This should address #1241 - windows doesn't have /bin/sh.
However, using cmd instead is also not quite straightforward since `exec.Command` has [some pitfalls](https://github.com/golang/go/issues/17149) on windows. This implements the suggested upstream workaround and allows us to launch mfa_process correctly on windows, even if the binary is in a path with spaces.

### DIFF
--- a/vault/mfa_unix.go
+++ b/vault/mfa_unix.go
@@ -1,0 +1,22 @@
+//go:build !windows
+
+package vault
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+func executeMFACommand(processCmd string) (string, error) {
+	cmd := exec.Command("/bin/sh", "-c", processCmd)
+	cmd.Stderr = os.Stderr
+
+	out, err := cmd.Output()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


Mishco:
I like this idea of a separate unix/windows Mfa process. Currently there is only a "solution" using WSL on Windows, which can be really slow.

Mishco (approved):

vault/mfa_unix.go
```diff
@@ -0,0 +1,22 @@
+//go:build !windows
```
mbevc1, line 1: This probably needs to be: `darwin || freebsd || openbsd || linux`

mbevc1:
Proposed feature looks good, would you mind opening a PR in the fork https://github.com/ByteNess/aws-vault ?

mbevc1:
Opened a PR in the fork and should be solved there: https://github.com/ByteNess/aws-vault/pull/50

FernandoMiguel:
moved to another fork

PR #1258, opened 2024-11-27
Implement OAuth2 PKCE in SSORoleCredentialsProvider

patrickod:
Inspired by [AWS CLI's recent change](https://aws.amazon.com/about-aws/whats-new/2024/11/aws-command-line-interface-pkce-single-sign-on/) this PR adds support for the OAuth2
"authorization code" grant flow accompanied by [Proof Key for Code
Exchange (PKCE) by OAuth Public Clients](https://datatracker.ietf.org/doc/html/rfc7636) and makes it the default.

In the "device code" flow the user is entirely responsible for matching
the one time password displayed by their terminal and browser. This is
error prone at best and a one click penalty for every login at worst.

In the "authorization code" flow we create an ephemeral HTTP server to
host an OAuth2 callback on 127.0.0.1. Users are redirected here after
confirming their login attempt and the code comparison step is automated
for them. As AWS enforces the use of 127.0.0.1 as a callback destination
host a remote attacker who successfully convinces a user to confirm
their malicious login prompt would not be able to access the resulting
credential as the user would be redirected to their own localhost and
not an attacker controlled destination.

If we determine that we are not able to open a browser (either from the
user specifying --stdout or observing that we are executing in an SSH
session) we revert to the previous "device code" flow and print the URLs
for copy/pasting into browsers.


vault/ssorolecredentialsprovider.go
```diff
@@ -201,3 +211,191 @@ func (p *SSORoleCredentialsProvider) newOIDCToken(ctx context.Context) (*ssooidc
...
+	n, err := crand.Read(codeVerifierBytes)
+	if err != nil || n != 32 {
+		return nil, fmt.Errorf("failed to generate PKCE verifier: %w", err)
+	}
+	codeVerifier := base64.RawURLEncoding.EncodeToString(codeVerifierBytes)
+
+	// generate the code challenge: base64(sha256(codeVerifier))
+	codeChallengeBytes := sha256.Sum256([]byte(codeVerifier))
+	codeChallenge := base64.RawURLEncoding.EncodeToString(codeChallengeBytes[:])
+	log.Printf("Generated PKCE code_challenge: %q", codeChallenge)
```
christophetd, line 227: I don't think this necessarily needs to be displayed by default

vault/ssorolecredentialsprovider.go
```diff
@@ -201,3 +211,191 @@ func (p *SSORoleCredentialsProvider) newOIDCToken(ctx context.Context) (*ssooidc
...
+		ClientType:   aws.String("public"),
+		GrantTypes:   []string{"authorization_code", "refresh_token"},
+		Scopes:       []string{"sso:account:access"},
+		IssuerUrl:    aws.String(p.StartURL),
+		RedirectUris: []string{"http://127.0.0.1/oauth/callback"},
+	})
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("Created new OIDC client (expires at: %s)", time.Unix(clientCreds.ClientSecretExpiresAt, 0))
```
christophetd, line 243: same remark on logging (and for following code)

vault/vault.go
```diff
@@ -141,13 +141,20 @@ func NewAssumeRoleWithWebIdentityProvider(k keyring.Keyring, config *ProfileConf
 func NewSSORoleCredentialsProvider(k keyring.Keyring, config *ProfileConfig, useSessionCache bool) (aws.CredentialsProvider, error) {
 	cfg := NewAwsConfig(config.SSORegion, config.STSRegionalEndpoints)
 
+	// If we're in an SSH session, we can't use the browser for SSO so we print
+	// the URLs to stdout instead.
+	useStdout := config.SSOUseStdout
+	if os.Getenv("SSH_CONNECTION") != "" {
```
christophetd: this seems like an addition that's independant from the PKCE flow? Overall I wouldn't necessarily expect this "magic" behavior from aws-vault, as a user I prefer a consistent behavior and explicitely passing `--stdout` if needs be

vault/ssorolecredentialsprovider.go
```diff
@@ -122,7 +129,19 @@ func (p *SSORoleCredentialsProvider) getOIDCToken(ctx context.Context) (token *s
...
-	token, err = p.newOIDCToken(ctx)
+
+	// if we must use stdout (either by user choice or because we have determined we are in an SSH session where we cannot open a browser)
+	// then we use the "device code" grant flow, and print URLs to stdout for the user to manually copy/paste into their browser.
+	//
+	// Otherwise we use the "authorization code" grant flow with Proof Key for
+	// Code Exchange (PKCE) and open the browser automatically. The latter flow
+	// is more user-friendly and secure because the step comparing the challenge
+	// code between the CLI and browser is automated entirely.
+	if p.UseStdout {
```
christophetd: I commonly use `--stdout` even when I'm on the same machine as where `aws-vault` is running, for various purposes (e.g. wanting to login in a non-default browser). I'm not 100% convinced that `UseStdout` should be the basis for chosing between the device code and PKCE flow, how about instead following the same behavior as the AWS CLI? It seems to have a flag `--use-device-code` flag https://github.com/aws/aws-cli/commit/130005af5ea6a75705ed528aaf06d533f449bef9#diff-e07a10a6eb1a677e905b0498651e672137b5dff19d357933bd7bc3e36f845a3bL42 which defaults to false

vault/ssorolecredentialsprovider.go
```diff
@@ -201,3 +211,191 @@ func (p *SSORoleCredentialsProvider) newOIDCToken(ctx context.Context) (*ssooidc
...
+	authorizeURL := url.URL{
+		Scheme:   "https",
+		Host:     host,
+		Path:     "/authorize",
+		RawQuery: args.Encode(),
+	}
+	log.Printf("Authorize URL: %s", authorizeURL.String())
+
+	// redirect user to the authorize URL
+	log.Println("Opening SSO authorization page in browser")
```
christophetd: (related to my first comment of this PR, I think this should rather honor the `--stdout` parameter to decide what to do)

vault/ssorolecredentialsprovider.go
```diff
@@ -201,3 +211,191 @@ func (p *SSORoleCredentialsProvider) newOIDCToken(ctx context.Context) (*ssooidc
...
+	fmt.Fprintf(os.Stderr, "Opening the SSO authorization page in your default browser (use Ctrl-C to abort)\n%s\n", authorizeURL.String())
+	if err := open.Run(authorizeURL.String()); err != nil {
+		log.Printf("Failed to open browser: %s", err)
+	}
+
+	// await the authorization code from the callback server once the user has completed the flow.
+	var code string
+	timeout := time.After(1 * time.Minute)
```
christophetd: FWIW the device code implementation [here](https://github.com/99designs/aws-vault/blob/949040e0413c23ac97ecdd1f397680643a294e76/vault/ssorolecredentialsprovider.go#L181) waits indefinitely (modulo a timeout on the AWS side), so perhaps it's worth increasing it?

vault/ssorolecredentialsprovider.go
```diff
@@ -201,3 +211,191 @@ func (p *SSORoleCredentialsProvider) newOIDCToken(ctx context.Context) (*ssooidc
...
+	// constant time string comparison of want vs got state
+	state := r.URL.Query().Get("state")
+	if subtle.ConstantTimeCompare([]byte(state), []byte(s.state)) != 1 {
+		http.Error(w, "Invalid state", http.StatusBadRequest)
+		return
+	}
+
+	// send the authorization code to the channel
+	code := r.URL.Query().Get("code")
+	s.code <- code
```
christophetd: perhaps it's worth returning early if there's an error rather than letting the consumer timeout? In which case s.code could be a channel of a struct that can hold either an `error` or a `string`, and the consumer would do something like

```go
code, err := <- server.code
if err != nil {
   // error
}
```

vault/ssorolecredentialsprovider.go
patrickod: I was confused at first also at the use of `log.Printf` in this project but these log lines are not displayed unless the user passes `--debug` as they're otherwise sent to `io.Discard`
https://github.com/99designs/aws-vault/blob/d4706c87383ca9ac8492c48e2d8a84dbd7e6cc07/cli/global.go#L166-L168

vault/vault.go
patrickod: Happy to back this out and have it be a separate discussion. My thinking was that given we now have login flows that cater to browser and non-browser setups to direct the user to the one that makes sense in context but I can understand how that would be confusing. A number of us at Tailscale use `aws-vault` with remote workspaces and run into the no-browser scenario frequently.

vault/ssorolecredentialsprovider.go
patrickod: Great shout. I hadn't considered the same device non-default browser path that users might wish to take. I was primarily thinking about remote workstations and in that context printing the URL out to be opened locally would not make sense as the eventual redirect destination would be inaccessible to the browser.

vault/ssorolecredentialsprovider.go
christophetd: ah, good point

vault/ssorolecredentialsprovider.go
```diff
@@ -201,3 +211,191 @@ func (p *SSORoleCredentialsProvider) newOIDCToken(ctx context.Context) (*ssooidc
...
+		"code_challenge":        {codeChallenge},
+		"scopes":                {"sso:account:access"},
+	}
+	// prefer the base endpoint from client options, otherwise use a default
+	var host string
+	if p.OIDCClient.Options().BaseEndpoint != nil && *p.OIDCClient.Options().BaseEndpoint != "" {
+		host = *p.OIDCClient.Options().BaseEndpoint
+	} else {
+		host = "oidc.us-east-1.amazonaws.com"
+	}
```
adamstirk-ct: Hardcoding the region to `us-east-1` won't work in all use cases. For example our SSO region is `eu-west-1`.

To get this to work I used `host = fmt.Sprintf("oidc.%s.amazonaws.com", p.OIDCClient.Options().Region)`

mbevc1:
Maintained fork: https://github.com/ByteNess/aws-vault

Issue #1259, opened 2024-12-05
feat: Signout the existing session


mbevc1:
As a an engineer when I login to AWS Console I want to signout of previous session to prevent manula action and need to run `login` again. So when running `aws-vault login <profile>` do an automatic signout before redirecting to the browser.

mbevc1:
I've merged this here: https://github.com/ByteNess/aws-vault

Also trying to keep it maintained there 😉

PR #1268, opened 2025-05-16
feat: Add auto logout before login

### LINKED ISSUE #721: Feature: Automatically logout before login
Using `aws-vault login` while already logged in opens a window that tells you to log out. Could there be a way to logout automatically?

mbevc1:
Fixes: #721 #1259

If we have an active session we'll logout first before trying to login. Previous behaviour can be restored using a `--no-autologout` flag.

Note: Logout only works on `us-east-1` endpoint, but doesn't' effect setting service region in the URL.

### DIFF
--- a/cli/login.go
+++ b/cli/login.go
@@ -28,6 +28,7 @@ type LoginCommandInput struct {
 	Config          vault.ProfileConfig
 	SessionDuration time.Duration
 	NoSession       bool
+	NoAutoLogout    bool
 }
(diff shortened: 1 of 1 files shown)


mbevc1:
Well, if anyone find this useful; I've forked the repo and merged it here: https://github.com/ByteNess/aws-vault

FernandoMiguel:
moved to another fork

PR #1276, opened 2026-07-22
switched Yubico Support site URL …

silpol:
…from original site to Web Archive copy, until Yubico gets broken link solved

Problem is that Google has cached copy, which is searchable with stems of URL in original USAGE.md (this file), but Yubico Support site has it removed. Or better say, they are careless and do CMS updates without having vanity links.

So until Yubico folks gets more careful, we better use web archive copy.


mbevc1:
@silpol this is an abandoned project, feel free to contribute to the maintained fork: https://github.com/ByteNess/aws-vault

https://github.com/99designs/aws-vault#aws-vault

